PR #5, opened 2026-05-12
fix(mcp): pin PYTHONUSERBASE so cli-local MCP can import mcp under per-agent HOME (v0.7.4)

hellohanchen:
## Summary

- Fixes `ModuleNotFoundError: No module named 'mcp'` in the cli-local MCP server on macOS / Linux. Root cause: the adapter rewrites `HOME` (and `USERPROFILE`) on the claude subprocess so each agent has its own `~/.claude`; the MCP subprocess claude spawns inherits that HOME and Python recomputes user-site relative to it, landing on an empty `<per-agent-home>/Library/Python/.../site-packages` instead of the real user-site where `mcp` lives.
- Diagnosed externally on a macOS install (`puffo_agent` was editable-installed → loaded via absolute-path `.pth` so it survived; `mcp` was in real user-site so it broke).
- Fix: snapshot `site.getuserbase()` at module load (daemon-side, real HOME) and inject it as `PYTHONUSERBASE` in the MCP env block. Python honours `PYTHONUSERBASE` over HOME-derived paths, so the spawned subprocess resolves user-site to the daemon's real base — `.pth` files processed, editable installs honoured.
- `cli-docker` deliberately excluded (container has its own Python tree, host paths are meaningless inside).
- Windows unaffected in practice (Windows user-site reads `APPDATA`, not `HOME` — the adapter doesn't touch `APPDATA`), but injection is harmless there too.
- Bump to 0.7.4.

## Test plan

- [x] `pytest tests/` on Windows: 439 passed, 5 skipped (unchanged).
- [x] New test `test_puffo_core_mcp_env_pins_python_user_base` (cli-local injects).
- [x] New test `test_puffo_core_mcp_env_skips_python_user_base_for_docker` (cli-docker skips).
- [ ] Manual: macOS, create a fresh agent → daemon restart → confirm MCP log no longer shows `ModuleNotFoundError: No module named 'mcp'` and `mcp__puffo__send_message` is available in the agent's tool list.
- [ ] Manual: Windows, cli-local + cli-docker both still spawn correctly with v0.7.4 installed (no regression on the platform that wasn't broken).
- [ ] Manual: macOS cli-docker still works — container env should NOT contain `PYTHONUSERBASE` (the host's user-base path would be nonexistent inside the container).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+## [0.7.4] — 2026-05-12
+
+### Fixed
+
+- cli-local: MCP server was crashing on agent startup with
+  `ModuleNotFoundError: No module named 'mcp'` on macOS / Linux
+  installs where `mcp` lives in real-user user-site (e.g. installed
+  via `pip install --user` or as a transitive dep of `pip install
+  -e .`). The adapter rewrites `HOME` (and `USERPROFILE`) on the
+  claude subprocess so each agent has its own `~/.claude`; the MCP
+  subprocess claude spawns inherited that HOME, and Python computed
+  user-site relative to it (`<per-agent-home>/Library/Python/...`),
+  landing on an empty per-agent directory. Result: every
+  `mcp__puffo__*` tool went missing and the agent fell back to
+  plain-text replies. Now `puffo_core_mcp_env` injects
+  `PYTHONUSERBASE` pointing at the daemon's real user base
+  (captured via `site.getuserbase()` at module load time, before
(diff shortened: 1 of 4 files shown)


hellohanchen:
Windows regression test passed (2026-05-12):

- `cli-local`: daemon restarted with v0.7.4 code, agent ran MCP tools end-to-end — no `ModuleNotFoundError`, no fallback to plain-text replies.
- `cli-docker`: agent ran normally; container env unaffected (PYTHONUSERBASE not injected for `runtime_kind == "cli-docker"`, as designed).

macOS verification still pending — the original repro environment. Once that's confirmed by the agent who filed the feedback, this is mergeable.

PR #20, opened 2026-05-15
feat(macos): transparent Keychain credential management (0.9.0a1)

hellohanchen:
## Summary

Daemon-owned OAuth credential lifecycle for macOS, layered on top of `main`'s 0.8.8 `CredentialRefresher` (PUF-221) via a pluggable backend abstraction so Linux/Windows and macOS share the lock + agent fan-out + 401-wake invariants and only diverge on storage.

Force-pushed over the previous 0.9.0a1 / 0.9.0a2 history — the parallel `CredentialManager` class + module-singleton design from those alpha iterations didn't compose cleanly with the post-PUF-221 state of `main`. Rebuilt from `main`@0.8.8 with one unified manager.

**This PR is NOT for merging yet.** Pre-release `0.9.0a3` published to TestPyPI so a macOS colleague can run the diagnostic probes and report back. Once verified, we'll cut 0.9.0 and merge.

## How to test (macOS colleague)

```bash
pip install --index-url https://test.pypi.org/simple/             --extra-index-url https://pypi.org/simple/             puffo-agent==0.9.0a3

puffo-agent test full-probe
```

The probe writes a structured markdown report to `~/.puffo-agent/probe-report.md`. Tokens are redacted (`len=N sha256_prefix=X`); the file is safe to paste back.

Probes:

| Probe | What it checks |
|---|---|
| `keychain-read` | `security find-generic-password` succeeds + parses |
| `keychain-write` | `security add-generic-password -U` roundtrip survives |
| `refresh-flush` | sandboxed `claude --print` actually rotates the cached token |
| `keychain-survives-token-env` | reproduces github.com/anthropics/claude-code#37512 with + without our PATH shim |
| `full-probe` | runs all of the above end-to-end |

## Architecture

`CredentialBackend` Protocol (`portal/credential_refresh.py`) with four methods: `expires_in_seconds`, `refresh` (async, returns `RefreshOutcome.{REFRESHED, UNCHANGED, FAILED}`), `sync_to_agent`, `bootstrap`.

- **`FileBackend`** — preserves bit-identical 0.8.8 behavior on Linux/Windows. Host `~/.claude/.credentials.json` is canonical; refresh runs `claude --print` with `HOME=host_home`; `sync_to_agent` is a symlink (or copy fallback) via `link_host_credentials`. External rotation propagates atomically through the symlink — no external-poll needed.

- **`KeychainBackend`** — macOS path. The Keychain is canonical; the daemon maintains a cache at `~/.puffo-agent/run/claude-credentials.json` (atomic-write JSON, chmod 600). Refresh runs a sandboxed `claude --print` under a tempdir `HOME` seeded from the cache so claude's native refresh path engages and writes the rotated blob back to the sandbox file (we then copy it to the cache). Writeback to Keychain is best-effort so the operator's main CLI sees the new token. `sync_to_agent` is a per-agent file copy (Keychain ACL is keyed on UID + signing identity, not HOME — symlinking the cache wouldn't help, since the per-agent `claude` process still goes through Keychain anyway).

- **`CredentialRefresher`** itself stays platform-agnostic: owns the `asyncio.Lock` (single-writer across all refreshes), the agent home registry, the `_refresh_request` event for the 401-wake (PUF-221 contract, wired via `Worker.__init__` callback — unchanged from main), the 2-minute file-expiry poll, and the post-tick fan-out that calls `backend.sync_to_agent(agent_home)` for every registered agent. Duck-types `poll_external_rotation` on the backend so macOS gets a sibling 5-min Keychain-poll task without `CredentialRefresher` needing to know about the macos module.

- **`daemon.py`** picks the backend at startup based on `is_macos()`. Backwards-compat: `CredentialRefresher(host_home=...)` still works (auto-builds a `FileBackend`) so the 12 `test_credential_refresher.py` tests pin the same contract unchanged.

## What changed vs 0.9.0a1/a2

- **Drops the parallel `CredentialManager` class** + the module-singleton pattern (`set_global_credential_manager` / `trigger_poll_now_if_macos` / `register_agent_home_if_macos` / `unregister_agent_home_if_macos`). On-401 trigger now goes through the same `notify_refresh_needed` callback on both platforms — uniform wiring through `Worker.__init__`, no globals.

- **Linux/Windows path is now the daemon-owned `CredentialRefresher` from PUF-221** (single writer to host `.credentials.json` → no rotating-RT race), not the per-agent `_run_refresh_oneshot` from the 0.9.0a2/a3 era.

- **PATH shim for issue #37512 preserved** in `macos/keychain.py` (verbatim from the previous branch): intercepts `security delete-generic-password "Claude Code-credentials"` while passing every other `security` call through to `/usr/bin/security`.

- **5-min external-rotation poll** + on-401 trigger preserved. Now implemented as a sibling task spawned by `CredentialRefresher.run_loop` when the backend exposes `poll_external_rotation`.

- **Diagnostic CLI (`puffo-agent test ...`)** ported verbatim — 5 probes, redacted markdown report.

## Test plan

- [x] Local unit tests: **726 passed / 7 skipped / 0 failed** on Windows (incl. 12 pre-existing `test_credential_refresher.py` pinned tests passing unchanged thanks to backwards-compat constructor; 34 new `test_macos_credential_manager.py` rewritten against the new shape; 18 new `test_macos_diagnostic.py`; 7 skipped are pre-existing symlink-not-available + permission-mode-pending entries, none new).
- [x] CLI surface compiles, `puffo-agent test --help` renders, `full-probe` on Windows correctly skips every macOS probe via `is_macos()` early-returns.
- [x] `python -c "import puffo_agent.portal.daemon; import puffo_agent.portal.credential_refresh; import puffo_agent.macos.keychain; import puffo_agent.portal.diagnostic"` clean on Windows.
- [ ] **macOS colleague runs `puffo-agent test full-probe` and reports back** ← gate for promotion to 0.9.0.

## Publishing to TestPyPI

After this PR's CI is green, go to Actions → "Publish to TestPyPI" → "Run workflow", select the `feat/macos-transparent-keychain` branch. Uses Trusted Publishing (no token), so the only manual step is clicking "Run workflow".

## History note (force-push)

This branch's force-pushed head (`5cfed64`) discards the previous 0.9.0a1 / 0.9.0a2 / refactor commits (`c3fbc63`, `a9a3dca`, `073dc6b`, `babd97c`, `a444918`, `75e0fa0`). Those are reachable via `git reflog` on the contributor's local machine if any of the prior alpha exploration needs to be revisited. Nothing was tagged or released from those commits — the only published artifact was `0.9.0a1` on TestPyPI, which remains downloadable but is now superseded.

## Related

- github.com/anthropics/claude-code/issues/37512 — unfixed, auto-closed in April 2026
- PUF-221 (`main` 0.8.8) — daemon-owned credential refresh; this PR extends its backend abstraction
- Internal: discussion thread about Keychain popup spam with macOS users

### DIFF
--- a/src/puffo_agent/agent/adapters/local_cli.py
+++ b/src/puffo_agent/agent/adapters/local_cli.py
@@ -23,7 +23,6 @@ import time
 from pathlib import Path
 
 from ...macos.keychain import (
-    install_path_shim,
     is_macos,
     shim_dir,
 )
@@ -252,7 +251,13 @@ class LocalCLIAdapter(Adapter):
         """
         if not is_macos():
             return {}
-        shim_path = install_path_shim(home_dir())
+        # Just compute the shim path — ``KeychainBackend.bootstrap()``
+        # installed the actual script once at daemon start
+        # (credential_refresh.py → keychain.install_path_shim).
+        # Re-installing on every adapter spawn would be wasted I/O
+        # and would race on the same file path when two workers spawn
+        # concurrently.
+        shim_path = shim_dir(home_dir())
         existing_path = os.environ.get("PATH", "")
         return {
             "CLAUDE_CONFIG_DIR": str(Path(self.agent_home_dir) / ".claude"),
--- a/src/puffo_agent/agent/adapters/local_cli.py
+++ b/src/puffo_agent/agent/adapters/local_cli.py
@@ -22,11 +22,17 @@ import shutil
 import time
 from pathlib import Path
 
+from ...macos.keychain import (
+    install_path_shim,
+    is_macos,
+    shim_dir,
+)
 from ...mcp.config import (
     default_python_executable,
     write_cli_mcp_config,
 )
 from ...portal.state import (
+    home_dir,
     link_host_credentials,
     seed_claude_home,
     sync_host_enabled_plugins,
@@ -207,6 +213,7 @@ class LocalCLIAdapter(Adapter):
             "HOME": str(self.agent_home_dir),
             "USERPROFILE": str(self.agent_home_dir),
             **self._permission_hook_env(),
+            **self._macos_credential_env(),
         }
         self._session = ClaudeSession(
             agent_id=self.agent_id,
@@ -221,6 +228,36 @@ class LocalCLIAdapter(Adapter):
             extra_args=extra,
         )
         return self._session
+
+    def _macos_credential_env(self) -> dict[str, str]:
+        """macOS-only env hardening:
+
+        - ``CLAUDE_CONFIG_DIR`` points at the per-agent .claude so the
+          agent's claude reads from ``<agent_home>/.claude/.credentials.json``
+          (which the daemon's ``KeychainBackend.sync_to_agent`` keeps
+          fresh) rather than racing the operator's main CLI for the
+          Keychain entry.
+        - ``PATH`` is prepended with the security-shim dir so the
+          buggy ``security delete-generic-password`` call from Claude
+          Code issue #37512 is intercepted before it can flush the
+          operator's main CLI auth.
+
+        Deliberately does NOT set ``CLAUDE_CODE_OAUTH_TOKEN`` — that
+        env var triggers the same fallback-combiner cleanup path that
+        deletes the Keychain entry. We let claude read its token
+        from the per-agent ``.credentials.json`` like normal.
+
+        Returns ``{}`` on non-macOS so the Linux/Windows spawn env is
+        unchanged.
+        """
+        if not is_macos():
+            return {}
+        shim_path = install_path_shim(home_dir())
+        existing_path = os.environ.get("PATH", "")
+        return {
+            "CLAUDE_CONFIG_DIR": str(Path(self.agent_home_dir) / ".claude"),
+            "PATH": f"{shim_path}{os.pathsep}{existing_path}",
+        }
 
     def _permission_hook_env(self) -> dict[str, str]:
         """Env vars the PreToolUse hook script reads. Claude inherits
--- a/src/puffo_agent/portal/credential_refresh.py
+++ b/src/puffo_agent/portal/credential_refresh.py
@@ -68,10 +68,6 @@ REFRESH_POLL_SECONDS = 120
 REFRESH_SAFETY_MARGIN_SECONDS = 10 * 60
 REFRESH_ONESHOT_TIMEOUT_SECONDS = 120
 
-# Re-export from macos.keychain for callers that want the constant
-# without importing the macos package directly.
-from ..macos.keychain import KEYCHAIN_POLL_INTERVAL_SECONDS  # noqa: E402
-
 
 class RefreshOutcome(enum.Enum):
     """Result of a single backend refresh attempt."""
@@ -151,6 +147,11 @@ class FileBackend:
         ]
         started = time.time()
         try:
+            # cwd=host_home so claude's project-resolution doesn't
+            # drift into the daemon's launch directory (which is
+            # whatever the operator ran `puffo-agent start` from);
+            # the host home is the operator's normal claude
+            # working dir and matches single-process /login UX.
             proc = await asyncio.create_subprocess_exec(
                 *cmd,
                 stdout=asyncio.subprocess.PIPE,
@@ -467,6 +468,12 @@ class CredentialRefresher:
         rotation. Runs as a sibling task to the main run_loop so the
         two cadences don't entangle (file-expiry poll = 2 min;
         external-rotation poll = 5 min)."""
+        # Lazy import — keeps the platform-agnostic module free of a
+        # hard dependency on the macos package, matching the pattern
+        # used inside ``KeychainBackend`` for every other macos
+        # touchpoint.
+        from ..macos.keychain import KEYCHAIN_POLL_INTERVAL_SECONDS
+
         interval = KEYCHAIN_POLL_INTERVAL_SECONDS
         while not stop_event.is_set():
             try:
--- a/src/puffo_agent/portal/credential_refresh.py
+++ b/src/puffo_agent/portal/credential_refresh.py
@@ -1,56 +1,62 @@
-"""PUF-221: daemon-owned Claude OAuth credential refresh.
+"""Daemon-owned Claude OAuth credential refresh.
 
 Replaces the per-agent ``refresh_ping`` from
 ``agent/adapters/base.py``. Every refresh of the on-disk Claude
-credentials file goes through ONE process — the puffo-agent
-daemon — so Anthropic's single-use refresh-token rotation can't
-be raced by N agent workers all reading the same disk RT and
-burning each other's in-memory copies.
+credentials file goes through ONE process — the puffo-agent daemon —
+so Anthropic's single-use refresh-token rotation can't be raced by N
+agent workers all reading the same disk RT and burning each other's
+in-memory copies.
 
-Implementation: the daemon owns one ``CredentialRefresher``
-instance. Its ``run_loop`` coroutine polls
-``~/.claude/.credentials.json`` every ``REFRESH_POLL_SECONDS``
-and triggers a refresh when ``expiresAt - now <
-REFRESH_SAFETY_MARGIN_SECONDS``, OR when something calls
-``notify_refresh_needed()``. Today's two callers: (a) the
-``puffo-agent agent refresh-token`` CLI subcommand, which drops
-a sentinel file the daemon's reconcile loop forwards into the
-in-process ``asyncio.Event``; (b) the per-worker
-``on_auth_failure`` hook fired from
-``worker._handle_suppressed_reply`` when an auth-class leak is
-detected in a turn reply. The event short-circuits the 2-minute
-poll so an operator- or agent-initiated refresh runs within ~1s.
+Backend abstraction
+-------------------
 
-The refresh itself shells out to ``claude --print "ok"`` with
-``HOME=<host_home>`` — mechanically identical to the cli-docker
-``_run_refresh_oneshot`` pattern that's been proven to work on
-Windows + macOS. Single-writer semantics: the daemon's mutex
-ensures only one ``claude --print`` is in flight at a time, so
-Anthropic's RT rotation can't be observed mid-write by another
-caller.
+The platform-agnostic ``CredentialRefresher`` owns the cross-cutting
+concerns: an ``asyncio.Lock`` (single-writer), an agent home registry,
+the ``_refresh_request`` event (notify_refresh_needed wake), the poll
+loop, and the per-tick fan-out to every registered agent. It delegates
+all platform specifics to a pluggable ``CredentialBackend``:
 
-After every successful refresh (or on every poll, as a safety
-net for the operator running ``claude /login`` externally on the
-device), the refresher fans out
-``state.link_host_credentials(host_home, agent_home)`` to each
-registered agent so per-agent symlinks / copy-mode copies stay
-in sync with the host file.
+  - ``FileBackend`` (Linux / Windows): host file
+    ``~/.claude/.credentials.json`` is the canonical store; refresh
+    spawns ``claude --print`` with ``HOME=host_home`` so the atomic
+    tmp+rename lands at the host file; per-agent sync is a symlink (or
+    copy fallback) via ``link_host_credentials``. External rotation
+    (operator running ``claude /login``) propagates atomically through
+    the symlink — no poll needed.
+  - ``KeychainBackend`` (macOS): Keychain is the canonical store
+    (Claude Code 2.x). The daemon maintains a cache at
+    ``~/.puffo-agent/run/claude-credentials.json``; refresh runs a
+    sandboxed ``claude --print`` so claude rotates the token against
+    Anthropic and writes it back; writeback to Keychain is best-effort
+    so the operator's main CLI / VS Code extension see the new token;
+    per-agent sync is a copy (Keychain ACL is keyed on UID + signing
+    identity, not HOME, so the per-agent HOME trick is moot). An extra
+    5-minute Keychain poll picks up rotations done by OTHER processes
+    (operator's main CLI, an agent's own claude self-refreshing on a
+    401) and fans them to running agents.
 
-Future: a follow-up PUF can replace the subprocess shell-out
-with a direct ``aiohttp.post`` to Anthropic's ``/oauth/token``
-endpoint once the endpoint + request shape are operator-
-verified. The subprocess path is safer to ship first because it
-re-uses the exact mechanism cli-docker already proves works.
+Public API (unchanged from the file-backend-only version):
+
+  - ``register_agent`` / ``unregister_agent`` — agent-home set
+  - ``notify_refresh_needed`` — in-process 401 trigger
+  - ``run_loop`` — the daemon's long-lived coroutine
+  - ``expires_in_seconds`` — diagnostics surface
+
+The refactor is invisible to the daemon's ``daemon.py`` wiring
+beyond the choice of backend at construction time.
 """
 
 from __future__ import annotations
 
 import asyncio
+import enum
 import json
 import logging
 import os
 import time
+from dataclasses import dataclass
 from pathlib import Path
+from typing import Optional, Protocol
 
 from .state import link_host_credentials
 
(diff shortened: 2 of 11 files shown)


hellohanchen:
Merged latest `main` into this branch (commit `a444918`) — brings in PUF-200 (0.8.5), PUF-214 (0.8.5), PUF-217 (0.8.5), PUF-206 (0.8.6). Three conflicts, all resolved. Bumped to `0.9.0a3`. Full suite **694 passed / 7 skipped / 0 failed**.

## Conflict resolution

| File | Resolution |
|---|---|
| `pyproject.toml` | `0.9.0a3` (next pre-release after 0.9.0a2 + forward-merge) |
| `CHANGELOG.md` | Added 0.9.0a3 entry at top citing what was forward-merged + analysis below. Kept 0.9.0a2 / 0.9.0a1 entries, then main's 0.8.6 / 0.8.5 / 0.8.4 sections. |
| `src/puffo_agent/agent/adapters/local_cli.py` | Kept HEAD (no-op `_run_refresh_oneshot`) — see analysis below |
| `tests/test_refresh_oneshot_home_env.py` | **Updated** to assert the no-op contract instead of the now-retired legacy-path behaviour — see below |

## Does PUF-214 / PUF-217 invalidate this PR's approach?

### Short answer: **no**. PUF-214 strengthens this PR; PUF-217 is structurally moot here but worth keeping in tree for defense in depth.

### PUF-214 (worker-layer error-string leak suppression) — **complementary, no conflict**

PUF-214 lives in `Worker._run`'s send-fallback path (`portal/worker.py`), not in any of the credential / adapter code this PR touches. It catches the *symptom* (raw `Not logged in. Please run /login` etc. leaking into the channel) when refresh fails for any reason — including any future bug in the CredentialManager.

After this PR, the CredentialManager is the single owner of OAuth refresh, with bounded-exponential backoff capping at 6h cadence on sustained failures (0.9.0a2 addition). That's the *prevention*. PUF-214 is the *containment*: if the CredentialManager ever produces a stale/dead token and Claude CLI emits its error string in response, PUF-214 keeps that string out of the channel and flips `runtime.health=auth_failed`.

These two layers are independent and reinforcing. Nothing in PUF-214's pattern set or backoff logic conflicts with this PR.

### PUF-217 (drop HOME override in `_run_refresh_oneshot`) — **structurally moot on this branch; kept for safety**

PUF-217's fix targeted the legacy per-agent refresh path: when `_run_refresh_oneshot` overrode `HOME=<agent_home_dir>`, Claude CLI's atomic `tmp+rename` replaced the agent's symlinked `.credentials.json` with a regular file, and the next `link_host_credentials` tick overwrote the fresh token with stale host content.

**This PR retires that path entirely.** Per 0.9.0a2's design:
- `LocalCLIAdapter._credentials_expires_in_seconds` returns `None` on every platform, short-circuiting `base.Adapter.refresh_ping`.
- `_run_refresh_oneshot` is a documented no-op.
- The daemon-level `CredentialManager` owns every refresh through `refresh_via_host_oneshot`, which runs in a sandboxed `$HOME` and writes back via Keychain (macOS) or directly to the host file (Linux/Windows).

So the failure mode PUF-217 fixed (`tmp+rename` clobbering an agent symlink) is impossible on this branch — there is no per-agent rename target. The merge resolution **keeps the no-op** since the daemon-owned design supersedes the per-agent fix. The merge commit's docstring on `_run_refresh_oneshot` adds a forward-merge note explaining the relationship and prescribing: if a future change ever re-enables a per-agent refresh, port the `HOME=os.environ` shape from PUF-217 — don't revert to the pre-PUF-217 per-agent `HOME` override.

PUF-217's three Linux-only tests in `test_refresh_oneshot_home_env.py` asserted that the legacy path DID spawn claude with the right env. On this branch those assertions are wrong (the path is a no-op). I replaced the file with a single test that pins the opposite contract: `_run_refresh_oneshot` **must not** spawn a subprocess or touch `auth_healthy`. This catches accidental reintroduction of the FB-88-prone per-agent refresh path.

### PUF-200 / PUF-206 — no interaction

PUF-200 (root_id auto-correction in MCP tools) and PUF-206 (Python-version precheck in `puffo_agent/__init__.py`) touch completely different surface area. No conflict, no design interaction. They merged cleanly.

## Test deltas

- `tests/test_refresh_oneshot_home_env.py`: 3 tests → 1 test. The 3 deleted asserted "the legacy refresh path spawns claude with operator HOME"; that's no longer a contract. The 1 new test asserts "the per-agent path is a no-op."
- `tests/test_worker_error_suppress.py`, `test_export_module.py`, `test_import_module.py`, `test_bridge_export_import.py`, `test_puffo_core_tools.py`, `test_cli_python_version.py`: forward-merged from main, all pass unchanged.
- Full suite: 694 passed / 7 skipped / 0 failed (skips are the existing Windows-no-symlinks gates).

## What this PR still needs

Same as before this merge:
- **macOS colleague runs `puffo-agent test full-probe` and reports back.** The 0.9.0 promotion was always gated on a clean macOS probe report from a real machine.

The merge doesn't move the needle on that — just keeps the branch buildable against current main and forward-folds the orthogonal bug fixes that landed during the alpha window.

If you'd rather publish 0.9.0a3 to TestPyPI first (so the macOS colleague tests against the freshest version with PUF-214's channel-leak safety net active), the publish-testpypi workflow can be re-triggered against this branch.

hellohanchen (commented):
## Solution review — overall verdict: code-side LGTM; ship gates remain (macOS colleague + opt-in `refresh-flush-forced`)

Disclaimer: I ran review on Linux — I can validate the architecture, the FileBackend regression-set, and the test suite, but I cannot exercise the Keychain code paths. The gate on macOS colleague verification via `puffo-agent test full-probe` is exactly the right gate; I have nothing to add there.

### Architecture

The `CredentialBackend` Protocol is the right factoring. The platform-agnostic `CredentialRefresher` keeps every load-bearing invariant from PUF-221 — the `asyncio.Lock`, the `_refresh_request` event, the `register_agent`/fan-out registry, the FIRST_COMPLETED wake-on-event wait, the order of `runtime.health` flip → `on_auth_failure` callback — and only the storage primitives diverge. The backwards-compat constructor (`CredentialRefresher(host_home=...)` still works → `FileBackend` auto-built) lets the 12 pinned `test_credential_refresher.py` tests run unchanged, which is the right proof that Linux/Windows behaviour is identical to 0.8.8.

The macOS sibling-task pattern for `poll_external_rotation` is well-chosen: cadences (file-expiry 2min, Keychain-poll 5min) live as separate tasks rather than entangled in one tick, and `CredentialRefresher` discovers the capability via `hasattr(backend, 'poll_external_rotation')` so it doesn't import `macos.*` from the platform-agnostic file.

### Local Linux verification

- **Full pytest:** `732 passed / 1 skipped / 0 failed` on Linux (vs. operator's `726/7-skip/0-fail` on Windows — delta is 6 Windows symlink skips that Linux runs).
- `tests/test_credential_refresher.py`: 12/12 pre-existing pinned tests pass through the backwards-compat constructor unchanged — that's the core regression check for Linux/Windows behaviour.
- 34 new tests in `test_macos_credential_manager.py` and 18 in `test_macos_diagnostic.py` all pass on Linux (they use `subprocess.run` monkey-patches to fake the `security` binary, so they're meaningful regression cover even off-macOS).
- All consumer imports clean (`Daemon`, `Worker`, `CredentialRefresher`, `FileBackend`, `KeychainBackend`, `local_cli`).

### Findings (none blocking; see line comments for specifics)

1. **`diagnostic.py:4` doc-stale** — module docstring references `puffo_agent.macos.credential_manager` from the 0.9.0a2 era; the actual module is `puffo_agent.macos.keychain`.
2. **`credential_refresh.py:73` import inconsistency** — the unconditional `from ..macos.keychain import KEYCHAIN_POLL_INTERVAL_SECONDS` at module load breaks the lazy-import pattern that every other macos-touchpoint in this file uses.
3. **`local_cli.py:255` redundant shim install** — `install_path_shim(home_dir())` runs synchronously on every adapter spawn, despite `KeychainBackend.bootstrap()` already installing it once at daemon start.
4. **`credential_refresh.py:159` regression vs main** — the cwd justification comment that landed in PUF-221 PR #32 round 4 was dropped in the FileBackend refactor.
5. **`diagnostic.py:622` `CLAUDE_CODE_OAUTH_TOKEN` via env** — briefly visible to `ps auxe`; intentional because the probe is reproducing #37512, but worth one acknowledging line.
6. **`diagnostic.py:243` vs `keychain.py:276` parallel impls** — diagnostic re-implements the sandbox claude-oneshot rather than calling into production's `_run_claude_oneshot`. If the prod sandbox env shape changes, the diagnostic drifts.

### What to do before promoting from TestPyPI

- [ ] macOS colleague runs `puffo-agent test full-probe` and reports back (gate per PR body).
- [ ] Optional: address the 6 nits above (any subset; none are blockers).
- [ ] Optional: opt-in `puffo-agent test refresh-flush-forced --yes` once on a willing host to definitively verify `refresh_via_oneshot` rotates against Anthropic + writeback survives — passive `refresh-flush` only confirms the no-op path while the token is still fresh.

Not approving in the GitHub sense since this is held back from merge pending TestPyPI verification, but code-side this is ready when the probes come back green. Nice work on the backend abstraction; the duck-typed `poll_external_rotation` discovery is particularly clean.

src/puffo_agent/portal/diagnostic.py
```diff
@@ -0,0 +1,856 @@
+"""``puffo-agent test ...`` — macOS Keychain integration probes.
+
+Designed to be run by a colleague on a real macOS host so we can
+verify the assumptions baked into ``puffo_agent.macos.credential_manager``
```
hellohanchen: Doc-stale: the actual module is `puffo_agent.macos.keychain`. `credential_manager` was the parallel class name from the 0.9.0a2 era that this PR rewrote out. Same lag pattern as the round-4 docstring nit on PUF-221 PR #32 — fix is a one-word swap.

src/puffo_agent/portal/credential_refresh.py
```diff
@@ -62,72 +68,423 @@
 REFRESH_SAFETY_MARGIN_SECONDS = 10 * 60
 REFRESH_ONESHOT_TIMEOUT_SECONDS = 120
 
+# Re-export from macos.keychain for callers that want the constant
+# without importing the macos package directly.
+from ..macos.keychain import KEYCHAIN_POLL_INTERVAL_SECONDS  # noqa: E402
```
hellohanchen: Inconsistency: this unconditional import pulls the macos package into `credential_refresh`'s import graph at module load on every platform. Every other macos touchpoint in this file uses lazy imports (`from ..macos.keychain import ...` inside method bodies — see L245, L268, L325, L342) precisely so the platform-agnostic module doesn't depend on the macos module being importable.

Options to pick from:
1. Move `KEYCHAIN_POLL_INTERVAL_SECONDS = 5 * 60` here (the only consumer is `_external_rotation_loop` in this file). The keychain module then re-imports it from here if it still needs the public-API surface, or drops it.
2. Inline-import inside `_external_rotation_loop` at L470, matching the pattern of the other macos touchpoints.
3. Hardcode `300` at L470 with a one-line comment.

Not blocking — the macos module is importable on Linux/Windows since `is_macos()` guards everything that calls `security` — but the pattern inconsistency makes the file harder to read.

src/puffo_agent/agent/adapters/local_cli.py
```diff
@@ -222,6 +229,36 @@ def _ensure_session(self) -> ClaudeSession:
...
+        env var triggers the same fallback-combiner cleanup path that
+        deletes the Keychain entry. We let claude read its token
+        from the per-agent ``.credentials.json`` like normal.
+
+        Returns ``{}`` on non-macOS so the Linux/Windows spawn env is
+        unchanged.
+        """
+        if not is_macos():
+            return {}
+        shim_path = install_path_shim(home_dir())
```
hellohanchen: Redundant install: `KeychainBackend.bootstrap()` already installs the shim once at daemon start (credential_refresh.py:330 → keychain.py:install_path_shim). This adapter-spawn-time install then re-writes the same file via `write_text` + `chmod` on every worker spawn.

Two concerns:
1. **Wasted I/O** on every adapter spawn. The shim body is static.
2. **Concurrent-write race**: two adapter spawns happening simultaneously (e.g., daemon reconciler starting two workers in parallel) hit the same file path with `write_text`. Not catastrophic — both write identical bytes — but it's a write-write race the bootstrap path doesn't have.

Suggestion: replace this call with `shim_dir(home_dir())` (read-only, just compute the path). Trust `KeychainBackend.bootstrap()` to have installed it. If the adapter needs to handle the edge case where bootstrap hasn't run yet (e.g., for tests, or if `is_macos()` flips at runtime — which it can't), make the call cheap-idempotent: only re-install if the file doesn't exist.

src/puffo_agent/portal/credential_refresh.py
```diff
@@ -62,72 +68,423 @@
...
+            "ok",
+        ]
+        started = time.time()
+        try:
+            proc = await asyncio.create_subprocess_exec(
+                *cmd,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.PIPE,
+                env=env,
+                cwd=str(self.host_home),
```
hellohanchen, line 160: Regression vs 0.8.8 main: the cwd-justification comment that landed in PUF-221 PR #32 round 4 was dropped in the FileBackend refactor. The 0.8.8 line read:

```
# cwd=host_home so claude's project-resolution doesn't
# drift into the daemon's launch directory (which is
# whatever the operator ran `puffo-agent start` from);
# the host home is the operator's normal claude
# working dir and matches single-process /login UX.
cwd=str(self.host_home),
```

This was added because the choice of `cwd=host_home` is genuinely non-obvious — `claude --print --max-turns 1 "ok"` writes a project transcript file into cwd, so it matters where you point it. Worth restoring.

src/puffo_agent/portal/diagnostic.py
```diff
@@ -0,0 +1,856 @@
...
+        "  A) PATH=$PATH (no shim), CLAUDE_CODE_OAUTH_TOKEN set\n"
+        "  B) PATH=<shim>:$PATH, CLAUDE_CODE_OAUTH_TOKEN set\n"
+        "Re-reads keychain after each. We expect A to delete (or be a\n"
+        "no-op if Anthropic fixed #37512), and B to never delete.",
+    )
+
+    def _run_oneshot(path_value: str) -> tuple[int, str]:
+        env = {
+            **os.environ,
+            "CLAUDE_CODE_OAUTH_TOKEN": access_token,
```
hellohanchen, line 638: Side-effect note: passing the user's real `accessToken` through env makes it briefly visible to `ps auxe` (and to any other process the user can inspect their own env from) for the duration of the `claude` subprocess. This is **intentional** — the probe is specifically reproducing #37512 which is triggered by `CLAUDE_CODE_OAUTH_TOKEN` — but the side effect isn't called out in the docstring.

Suggest a one-line comment here:

```python
# The token is briefly visible to `ps auxe` while claude runs;
# accepted because this probe's purpose is to repro #37512,
# which only fires when CLAUDE_CODE_OAUTH_TOKEN is set.
```

This matters for anyone running the probe on a shared host (CI box, dev container, etc.). On a personal macOS workstation it's a non-issue.

src/puffo_agent/portal/diagnostic.py
```diff
@@ -0,0 +1,856 @@
...
+            "verify-read",
+            VERDICT_NEEDS_ATTENTION,
+            "post-write blob DIFFERS from pre-write blob — could be a "
+            "concurrent refresh by your main CLI.",
+        )
+        rpt.summary = "Roundtrip OK but content changed — likely benign."
+    return rpt
+
+
+def _run_sandboxed_claude_oneshot(
```
hellohanchen, line 243: Parallel implementation: this sync `_run_sandboxed_claude_oneshot` re-does the same sandbox-HOME dance that `puffo_agent.macos.keychain._run_claude_oneshot` (L276) + `refresh_via_oneshot` (L325) do in async/production.

They diverge today only in sync-vs-async, but the env shape is duplicated:

- `HOME`/`USERPROFILE` = sandbox
- `CLAUDE_CONFIG_DIR` = sandbox/.claude
- `PATH` = shim_dir + existing PATH
- claude args identical

If prod adds (say) a new env var or `--config` flag, the diagnostic silently drifts and stops validating the prod path. Two cleanup options:

1. Lift the env construction into a shared helper in `macos/keychain.py` (`def build_sandbox_env(sandbox, shim_path) -> dict[str, str]`) and have both diagnostic + production call it.
2. Leave as-is but add a `# keep in sync with macos.keychain.refresh_via_oneshot` comment on both sides so future-you (or future Calculation) notices the duplication.

Non-blocking — this PR is correct as-is — but the diagnostic loses its load-bearing value the moment the two drift.

hellohanchen:
Iter pushed at b07ea38. All 6 review comments addressed; the CodeQL bot alert is a deferred false positive (see below).

**diagnostic.py:4 (doc-stale)** — one-word swap: ``credential_manager`` → ``keychain``. Module path has been ``puffo_agent.macos.keychain`` since the 0.9.0a3 rebuild.

**credential_refresh.py:73 (unconditional macos import)** — dropped the top-level ``from ..macos.keychain import KEYCHAIN_POLL_INTERVAL_SECONDS``; the constant is now lazy-imported inside ``_external_rotation_loop`` itself, matching the pattern every other macos touchpoint in this file already uses. ``credential_refresh.py`` no longer pulls the macos package into its import graph on platforms where ``poll_external_rotation`` will never be called.

**credential_refresh.py:159 (regression vs 0.8.8)** — restored the ``cwd=host_home`` justification comment verbatim from PUF-221 PR #32 round 4. Apologies for losing it in the refactor — the agent that did the port preserved the call but not the rationale block. Now back where it was.

**local_cli.py:255 (redundant shim install)** — adapter spawn now calls ``shim_dir(home_dir())`` (path-only, read) instead of ``install_path_shim(home_dir())`` (writes the bash script + chmods on every spawn). ``KeychainBackend.bootstrap()`` already does the install once at daemon start. Removes the per-spawn ``write_text`` + ``chmod`` overhead and closes the concurrent-spawn write-write race you flagged.

**diagnostic.py:622 (token in env / ps auxe)** — added the suggested comment verbatim. Explicit that this is intentional for the #37512 repro and that production refresh paths deliberately don't set ``CLAUDE_CODE_OAUTH_TOKEN``. Worth saying out loud since the probe is run by colleagues on potentially-shared hosts.

**diagnostic.py:243 + macos/keychain.py:331 (parallel impl)** — paired ``keep in sync`` warning comments on both sides: ``diagnostic._run_sandboxed_claude_oneshot`` (sync) and ``keychain.refresh_via_oneshot`` (async). They share env shape + claude args; the diagnostic loses its load-bearing value the moment they drift. Went with the minimal option (cross-referenced comments) rather than extracting a shared ``build_sandbox_env`` helper — the diagnostic is sync + cross-platform-stubbable and the prod path is async + macOS-gated, so the shared helper would have to wrap both behaviors and the win is marginal. Happy to extract if you'd rather; flagging the decision.

**CodeQL: "weak hashing on sensitive data" at diagnostic.py:110** — false positive. The flagged code is:

```python
def _redact_token(s: str) -> str:
    """Show length + sha256 prefix, never raw token."""
    digest = hashlib.sha256(s.encode("utf-8")).hexdigest()[:12]
    return f"len={len(s)} sha256_prefix={digest}"
```

This is a 12-char token *fingerprint* for log redaction, not a password hash. The point is to give operators something they can grep for in logs ("did the same token surface in three different probe sections?") without leaking the actual token. No password material, no key derivation, no auth boundary — it's a redaction primitive. Same class of FP as the ``_summarise_credentials`` "presence boolean printed near accessToken= literal" alert that was dismissed on puffo-server PR #32. Will dismiss in the GitHub UI as ``Won't Fix → False Positive`` with a one-line comment.

Tests still **726 passed / 7 skipped / 0 failed** locally on Windows.

PR #22, opened 2026-05-16
fix(mcp): channel→space resolves from event-driven cache, not space walk

hellohanchen:
## Symptom

Two related symptoms when an agent operates in a channel that isn't in its "home" \`cfg.space_id\`:

1. \`list_channel_members\` returns the wrong space's roster (or a 404 / 403) — the URL was hard-coded with \`cfg.space_id\` regardless of where the channel actually lived.
2. \`send_message\` to a freshly-joined channel either silently sent into the wrong space (pre-FB-76) or paid a multi-call walk over \`GET /spaces\` + \`/spaces/<sp>/channels\` (post-FB-76, slower / still racy).

The diagnostic thread is on the prior session — symptom 1 was reported on a brand-new user; symptom 2 was FB-76.

## Root cause

The channel→space cache that backs \`lookup_channel_space\` was only populated from inbound **messages**. Membership events carrying \`(channel_id, space_id)\` — \`accept_channel_invite\`, \`invite_to_channel\`, \`create_channel\` — landed on the WS but were never recorded. So a channel the agent had just been admitted to remained \"unknown\" until the first message arrived.

## Fix

\`puffo_core_client._handle_event\` now feeds the cache from every membership event that proves the agent can reach the channel:

| Event kind | Admission gate |
|---|---|
| \`invite_to_channel\` | \`payload.invitee_slug == self.slug\` |
| \`accept_channel_invite\` | \`event.signer_slug == self.slug\` |
| \`create_channel\` | always — server only fans these to space members |

\`_accept_invite\` also writes the mapping synchronously after posting the accept, so the immediately-following intro-nudge send doesn't race the WS echo back.

With events feeding the cache, the FB-76-era \`_discover_channel_space\` walker is unnecessary. Replaced by a thin \`_resolve_channel_space\` helper that's purely a cache lookup; on miss the MCP tool raises a clear \"agent has no record of channel\" error.

Three MCP tools simplified to the same cache-lookup-or-raise pattern:
- \`send_message\` — drops the two-stage resolver.
- \`send_message_with_attachments\` — drops the silent \`or cfg.space_id\` fallback.
- \`list_channel_members\` — drops the \`cfg.space_id\` hardcoding.

\`cfg.space_id\` is now genuinely just metadata for the agent's home space; no MCP code path falls back to it.

## Test plan

- [x] New: \`_handle_event\` cache admission tests per kind + signer gate
  - \`test_handle_event_synthetic_accept_records_channel_space\`
  - \`test_handle_event_create_channel_records_mapping\`
  - \`test_handle_event_invite_to_channel_for_self_records_mapping\`
  - \`test_handle_event_invite_to_channel_for_other_skips_cache\`
  - \`test_handle_event_accept_signed_by_other_skips_cache\`
- [x] Rewritten: cross-space \`send_message\` test pre-caches the mapping
- [x] New: \`send_message\` / \`list_channel_members\` cache-miss tests assert no HTTP at all
- [x] \`pytest tests/\` — 542 passed, 3 pre-existing skips (Docker / symlinks / permission-mode)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,13 +77,129 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   resolution stack for little real benefit given the server is
   authoritative for membership state anyway).
 
+### Changed
+
+- **MCP tool surface for channel discovery is now three explicit
+  tools instead of one implicitly-cross-space `list_channels`.**
+
+  - `list_spaces()` — every space the agent is a member of.
+    `GET /spaces` is server-filtered, so the result reflects
+    authoritative permissions: anything in the list is a space
+    the agent can write to.
+  - `list_channels_in_space(space_id)` — channels in one named
+    space. `space_id` is required (empty → MCP tool error) so
+    the LLM can't accidentally fall back on the legacy
+    `cfg.space_id` for routing.
+  - `list_channels_in_all_spaces()` — convenience: walks
+    `GET /spaces` plus one `GET /spaces/<sp>/channels` per
+    space, grouped output. Same behaviour as the old
+    `list_channels` cross-space rewrite, renamed for clarity.
+
+  The old `list_channels` walked `cfg.space_id`'s event stream
+  and returned `(no space configured)` when that field was
+  empty, both of which made it impossible for an agent
+  operating in multiple spaces to enumerate its true channel
+  surface. Three-tool split gives the LLM precise vocabulary
+  for "which spaces" vs "which channels in space X" and removes
+  the last MCP-surface dependency on `cfg.space_id`. Primer
+  (`shared_content.py`), tool allowlist (`mcp/config.py`), and
+  log-message tool-name examples (`local_cli.py` /
+  `docker_cli.py`) are all updated to point at the new names.
+
+### Hardened
+
+- **`send_fallback_message` no longer silently routes to
+  `self.space_id` on cache miss.** The daemon's reply-when-LLM-
+  skipped-`send_message` path used to fall back to the legacy
+  home space when the in-memory channel→space map didn't have
+  the inbound channel — under cross-space deployments or after
+  a kick/cascade that evicted the cache, that sent the reply
+  to the wrong space (server 403) or worse to a space the
+  agent had just been removed from. Now drops the reply with a
+  clear log line; legacy `self.space_id` slot stays in the
+  constructor but no code path consults it for routing.
+
+- **Membership-exit eviction now drops the persistent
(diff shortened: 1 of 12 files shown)


hellohanchen:
Reviewed against current main (merged via \`6b7912e\`) and the three requirements you laid out. Follow-up changes in commit \`4099872\`.

## What worked already

**Req 1 (membership ↔ permission sync)**: PR #22's event-driven channel→space cache + the recent 0.8.7 \`leave_space\` / \`remove_from_space\` / \`remove_from_channel\` / \`leave_channel\` / \`cancel_*\` handlers (merged from main) cover both the entry and exit edges. Pre-fix gaps in the entry events (\`invite_to_channel\`, \`accept_channel_invite\`, \`create_channel\`) are closed by your \`_maybe_cache_channel_space\`.

**Req 2 (wrong target → correct error)**: \`_resolve_channel_space\` raising on miss is the right shape — clear MCP error message back to the LLM.

## What was leaking — fixed in 4099872

1. **\`send_fallback_message\` silent fallback to \`self.space_id\`** (req 3 leak). The daemon's reply-when-LLM-skipped-send_message path was still routing unknown channels into the legacy home space. Now drops + logs.

2. **\`_evict_*_caches\` only touched in-memory state**. The persistent \`channel_space_map\` SQLite table — which the MCP subprocess reads via \`lookup_channel_space\` — wasn't being cleaned on kick/cascade. The LLM would happily resolve a channel-we're-no-longer-in to its old space, pay an RTT, get a 403. Added \`MessageStore.unmark_channel_space\` + \`unmark_channel_space_for_space\`; wired into the eviction helpers (now async).

3. **\`list_channels\` MCP tool was home-space-only** (req 3 leak — the more visible one). Used \`cfg.space_id\` + \`/spaces/<sp>/events\` replay. Rewritten to enumerate \`GET /spaces\` + one \`GET /spaces/<sp>/channels\` per space, grouped output. \`cfg.space_id\` is intentionally not consulted. New test \`test_list_channels_ignores_cfg_space_id\` anchors req 3 against future regressions.

## Tests

\`pytest tests/\`: **681 passed, 9 skipped** (skips are Windows-symlink-related, unrelated to spaces). New tests:
- \`test_membership_events.py\`: +2 — persistent map eviction (whole-space + single-channel paths)
- \`test_worker_integration.py\`: +1 — \`send_fallback_message\` drop-on-unknown
- \`test_puffo_core_tools.py\`: -3 (legacy pagination tests no longer applicable) +4 — cross-space enumeration, empty-agent, cfg.space_id non-influence, per-space-\`str\` tight-race tolerance.

Ready for review.

PR #24, opened 2026-05-18
fix(PUF-200): auto-correct non-root thread_root_id in send_message

hellohanchen:
## Summary

Agents calling `mcp__puffo__send_message` (and `send_message_with_attachments`) with `root_id` set to a reply's envelope id — instead of the thread's true root — used to land in a sub-thread the human client doesn't surface, with no failure signal returned. We hit this twice live on 2026-05-18 (clone-report `msg_38364760` and build-test-report `msg_9e8f1a83`, both threaded under root `msg_610fec10`) and both messages vanished from the operator's view.

This PR adds a `_resolve_root_id(root_id, data_client)` helper next to `_coerce_root_visibility` that looks up the supplied envelope, walks to the real `thread_root_id` (capped at depth 4 with cycle detection), substitutes it before encrypting, and appends a correction note to the tool response so the agent learns the rule on the spot. Lookup miss / transport error falls through with a soft warning rather than failing the send.

Wired into both `send_message` (line 397) and `send_message_with_attachments` (line 894). The attachments-path display string ("in thread …") also reflects the resolved root, not the original wrong id.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-200.md`

## Test plan

- [x] Unit matrix on `_resolve_root_id` (8 tests): empty / whitespace, true root, single-level reply, depth-2 walk, lookup miss (None return), transport error, real `DataNotFound`, cycle defensive cap.
- [x] Integration on `send_message` parametrized over the **two real 2026-05-18 live failures** + 3 single-case integrations: real-root unchanged, unknown-root fall-through with warning, root-level skips lookup entirely.
- [x] Integration on `send_message_with_attachments`: same auto-correction; display string reflects the resolved root.
- [x] Full pytest in worktree: **596 passed / 1 intentional skip / 0 failed**.
- [x] Independent QA round 2 by Solution after iteration on docstring trim + DataNotFound coverage + parametrize ask.

### DIFF
--- a/src/puffo_agent/mcp/puffo_core_tools.py
+++ b/src/puffo_agent/mcp/puffo_core_tools.py
@@ -177,6 +177,97 @@ def _coerce_root_visibility(
     return is_visible_to_human, ""
 
 
+_RESOLVE_ROOT_MAX_DEPTH = 4
+
+
+async def _resolve_root_id(
+    root_id: str, data_client: Any,
+) -> tuple[Optional[str], str]:
+    """Resolve a caller-supplied ``root_id`` to the actual thread
+    root.
+
+    Agents sometimes pass an envelope's ``post_id`` (a specific reply)
+    as ``root_id`` expecting the new message to thread under that
+    reply. But ``thread_root_id`` on the wire must point at the *root*
+    of the thread — when it points at a non-root reply the resulting
+    message lands in a sub-thread or is dropped silently by the client
+    renderer. The bug is invisible to the agent: ``send_message``
+    returns success either way. We learned that the hard way during
+    the team's first hour on 2026-05-18.
+
+    Behaviour:
+    - Empty / whitespace ``root_id`` → ``(None, "")``. No lookup;
+      the envelope's ``thread_root_id`` stays ``None`` (root-level).
+    - Looked-up message's ``thread_root_id`` is ``None`` → the
+      supplied id IS a root → return ``(root_id, "")``.
+    - Looked-up message has a non-null ``thread_root_id`` → the
+      supplied id was itself a reply; substitute the real root and
+      return a correction note naming both ids so the agent learns.
+    - Defensive: if the substituted root is also a reply (malformed
+      data; should never happen for well-formed messages), walk one
+      more level. Capped at ``_RESOLVE_ROOT_MAX_DEPTH`` to prevent
+      infinite traversal on cycles.
+    - Lookup miss / transport failure → fall through with the
+      original ``root_id`` plus a soft-warning note. The send still
+      completes; the agent is told we couldn't verify the root.
+
+    Returns ``(resolved_root_or_None, note)``. ``note`` is empty
+    when no correction was needed (and no warning was emitted),
+    mirroring ``_coerce_root_visibility``'s shape.
+    """
+    if not root_id.strip():
+        return None, ""
+
+    current = root_id
+    seen: set[str] = set()
+    walked = False
+
+    for _ in range(_RESOLVE_ROOT_MAX_DEPTH):
+        if current in seen:
+            break
+        seen.add(current)
+        try:
+            msg = await data_client.get_message_by_envelope(current)
+        except DataNotFound:
+            msg = None
+        except Exception as exc:
+            logger.warning(
+                "resolve_root_id: lookup transport error for %s: %s",
+                current, exc,
+            )
+            return root_id, (
+                f"\nnote: could not verify root_id {root_id} is a thread "
+                "root (lookup failed); sent as-is. If this lands in the "
+                "wrong thread, pass the metadata block's thread_root_id, "
+                "not post_id."
+            )
+        if msg is None:
+            return root_id, (
+                f"\nnote: could not verify root_id {root_id} is a thread "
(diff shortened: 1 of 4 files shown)


src/puffo_agent/mcp/puffo_core_tools.py
```diff
@@ -177,6 +177,76 @@ def _coerce_root_visibility(
...
+                "not post_id."
+            )
+        if msg is None:
+            return root_id, (
+                f"\nnote: could not verify root_id {root_id} is a thread "
+                "root (message not in local store); sent as-is. If this "
+                "lands in the wrong thread, pass the metadata block's "
+                "thread_root_id, not post_id."
+            )
+        parent_root = msg.thread_root_id
```
hellohanchen, line 245: Quick schema-level sanity check: the data model has both `thread_root_id` (true thread root) and `reply_to_id` (immediate parent — chain-able). Per `message_store.py`'s schema, `thread_root_id` should always point at a true root (i.e. its own `thread_root_id IS NULL`) — so on healthy data this loop is one-iter-and-out and `parent_root.thread_root_id is None` always on the second hop.

The depth-2 walk test (`msg_root → msg_mid → msg_leaf`) is therefore exercising a data shape the relay shouldn't ever produce. That's fine as corruption defense, but it'd be worth a docstring line clarifying that — otherwise next reader will wonder why we walk a chain that, by schema contract, shouldn't exist.

src/puffo_agent/mcp/puffo_core_tools.py
```diff
@@ -177,6 +177,76 @@ def _coerce_root_visibility(
...
+        walked = True
+        current = parent_root
+
+    return current, (
+        f"\nnote: root_id {root_id} was a reply, not a root — "
+        f"auto-corrected to {current} after walking "
+        f"{_RESOLVE_ROOT_MAX_DEPTH} levels. Pass the metadata block's "
+        "thread_root_id, not post_id."
+    )
+
```
hellohanchen, line 271: Cosmetic only — this branch fires for two distinct cases: cycle detected (`current in seen` break) or depth cap exhausted (loop falls through naturally). The note text always says "after walking 4 levels", which is correct for the depth-cap case but slightly misleading for the cycle case (where we may have walked just 2 levels before detecting the loop). Real-world impact is zero since both cases require corrupt data; just flagging in case the wording bothers you.

Also: if the cycle is `A → B → A`, `current` ends up back at `A` (the original `root_id`), and the note reads "auto-corrected to msg_a" — i.e. corrected to itself. Weird-looking output for an impossible-in-practice scenario.

src/puffo_agent/mcp/puffo_core_tools.py
```diff
@@ -807,20 +884,22 @@ async def send_message_with_attachments(
...
+            thread_root_id=resolved_root,
             content_type=ATTACHMENT_CONTENT_TYPE,
             content=body_content,
             recipients=devices,
         )
         envelope = encrypt_message(inp, signing_key)
         await cfg.http_client.post("/messages", envelope)
         names = ", ".join(t.name for t in targets)
-        thread_note = f" in thread {root_id}" if root_id else ""
+        thread_display = resolved_root if resolved_root else root_id
```
hellohanchen: Tiny nit: `_resolve_root_id` only returns `None` when `root_id.strip()` is empty (per its contract at L196). So when `resolved_root is None`, `root_id` is also empty, and the `else root_id` fallback is a no-op. Could simplify to `thread_display = resolved_root` and drop the conditional — but it doesn't matter for correctness.

hellohanchen (commented):
Looks solid overall. Tight scoped fix for a real production bug, the helper mirrors `_coerce_root_visibility` cleanly, and the test coverage is unusually thorough — parametrising the integration test on the two **actual** 2026-05-18 live-failure envelope IDs is a really nice touch (gives the test a date-stamped regression anchor instead of a synthetic case). Soft-failure on lookup miss / transport error is the right call — better to land in the wrong thread than to drop the message.

Three minor observations inline, all in the "if you're touching it" bucket — none blocking.

One broader thing worth a quick confirm before merge: `_coerce_root_visibility` runs **before** `_resolve_root_id`, and both consume the original `root_id`. That's correct behaviour (the agent meant "threaded reply with hidden visibility" — the auto-correction just retargets the thread, doesn't change the reply-vs-root intent), but it means a reply with `is_visible=false` whose `root_id` is auto-corrected stays hidden. Worth a sentence in the helper docstring if you want future readers to not have to re-derive that.

Not a regression risk for messages — every message still goes out, just possibly with a different `thread_root_id`. ✅ from me, will defer to your final review.

hellohanchen:
Re-reviewed 459557f. All three observations addressed — and the L248 response is actually stronger than what I suggested: instead of just rewording the corruption-fallthrough note, the helper now preserves the original `root_id` (with a loud "could not resolve" warning) rather than auto-correcting to a value the corrupt chain can't justify. That's the right call. Docstring on `_resolve_root_id` also now spells out the ordering vs `_coerce_root_visibility` and explicitly frames the multi-step walk as corruption defense.

Pushed `chore(release): 0.8.5` (28cf02a) onto this branch with the changelog entry — should now be ready for your final review + merge.

PR #25, opened 2026-05-18
fix(PUF-214): suppress worker-error leaks before send_fallback_message

hellohanchen:
## Summary

Worker-layer egress filter that intercepts known error-string leaks before they reach `client.send_fallback_message(...)` and surface to the user in chat. Pre-fix, when Claude CLI's OAuth died mid-session (FB-159 / Sheri / Yasushi class) or the Anthropic API returned an authentication/rate-limit error (FB-105 class), the raw error string was posted into the channel as if it were the agent's reply. Operators saw their personal/family DMs polluted with `Not logged in. Please run /login` and `[puffo-agent system message] session errored on rate limiting…`.

This PR adds anchored regex patterns matching only the leak signatures (legit agent prose mentioning "rate limit" or "Not logged in" passes through), a shared `_handle_suppressed_reply` helper that logs + populates `runtime.error` + (on auth-class) flips `runtime.health=auth_failed` so the signal lands on `puffo-agent status` instead of in the channel, and wires both `send_fallback_message` call sites in `Worker._run()` through the filter.

**Defense-in-depth pair with PUF-207 (startup OAuth verify) and PUF-213 (adaptive credential refresh).** Together they close the silent-can't-recover surface: PUF-207 catches at startup, PUF-213 prevents staleness during runtime, this catches the leak if either misses.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-214.md`

## Test plan

- [x] Full pytest: **601 passed / 1 intentional skip / 0 failed**.
- [x] 19 tests in `tests/test_worker_error_suppress.py`:
  - 5 positive filter matches (Claude CLI Not-logged-in single + multi-line, kick-text echo, `authentication_error`, `rate_limit_error` with all delimiter variants).
  - 5 overreach guards (helpful login/rate-limit/authentication prose, empty reply, normal agent message — all pass through unchanged).
  - 4 `_looks_like_auth_error` sanity cases.
  - 4 `_handle_suppressed_reply` contract tests (return value, runtime.error population, persistence, scope-specific recovery message, rate-limit-doesn't-flip-health, legit-prose-no-side-effects).
  - 3 call-site contract tests using a `_RecordingClient` mock — assert `send_fallback_message` is NOT called on suppression / IS called once on legit prose / NOT called on empty reply. The test-file helper mirrors the exact `if reply and not _handle_suppressed_reply(...): await client.send_fallback_message(...)` shape from `Worker._run()` so a future call-site edit that forgets the guard breaks the test.
- [x] Independent QA with one iteration on the mock-client assertion + comment trim.

## Flag for reviewers

The pattern `\brate[_ -]limit[_ -]error\b` with the **space** variant will suppress an agent saying "I got a rate limit error from the API, retrying" — verbatim use of that phrase in agent prose is uncommon, but if it happens the message is suppressed (runtime.error populated, no channel post for that turn). The trade-off favours suppression given FB-105's recurrence volume, but if you'd prefer to drop the space variant (limit the regex to `_` and `-` only), it's a one-character change in `worker.py`'s pattern list.

### DIFF
--- a/src/puffo_agent/portal/worker.py
+++ b/src/puffo_agent/portal/worker.py
@@ -376,31 +376,41 @@ def _handle_suppressed_reply(
     agent_id: str,
     *,
     scope: str,
-    treat_auth_as_health: bool,
 ) -> bool:
     """Shared landing for a suppressed worker-error leak. Returns
     True when the reply was suppressed and the caller should skip
-    ``send_fallback_message``; False when the reply is clean. On
-    suppression: log the truncated payload, populate
-    ``runtime.error`` with a scope-tagged message, and (if
-    ``treat_auth_as_health`` and the leak looks auth-class) flip
-    ``runtime.health="auth_failed"``."""
+    ``send_fallback_message``; False when the reply is clean.
+
+    On suppression: log the truncated payload, populate
+    ``runtime.error`` with a scope-tagged + leak-class-tagged
+    message, and (if the leak is auth-class) flip
+    ``runtime.health="auth_failed"`` — that signal is definitive
+    regardless of which scope surfaced it."""
     safe_reply = _suppress_worker_error_leak(reply)
     if safe_reply is not None:
         return False
+    is_auth = _looks_like_auth_error(reply)
     logger.warning(
         "agent %s: suppressed worker-error leak in %s reply: %s",
         agent_id, scope, reply[:200],
     )
-    if treat_auth_as_health and _looks_like_auth_error(reply):
+    if is_auth:
         runtime.health = "auth_failed"
     if scope == "api-error-retry":
-        runtime.error = (
-            "Worker emitted an auth/rate-limit error string after an "
-            "API error; suppressed from channel post. Operator should "
-            "check Claude CLI auth and re-run "
-            f"`puffo-agent agent resume {agent_id}` once recovered."
-        )
+        if is_auth:
+            runtime.error = (
+                "Worker emitted an auth-error string after an API "
+                "error; suppressed from channel post. Re-run "
+                "`claude /login`, then "
+                f"`puffo-agent agent resume {agent_id}`."
+            )
+        else:
+            runtime.error = (
+                "Worker emitted a rate-limit error string after an "
+                "API error; suppressed from channel post. Usually "
+                "self-recovers — investigate the daemon log if "
+                "persistent."
+            )
     else:
         runtime.error = (
             "Worker emitted an auth/rate-limit error string instead of "
@@ -742,7 +752,6 @@ class Worker:
                 self.runtime,
                 agent_id,
                 scope="fallback",
-                treat_auth_as_health=False,
             ):
                 # Fallback reply when the agent skipped both
                 # send_message and [SILENT]. Post to root so the
@@ -781,7 +790,6 @@ class Worker:
                 self.runtime,
                 agent_id,
                 scope="api-error-retry",
-                treat_auth_as_health=True,
             ):
                 # The kick-retry reply is the hottest leak site — see
                 # PUF-214 / FB-88 / FB-159 case-studies (Claude CLI's
--- a/src/puffo_agent/portal/worker.py
+++ b/src/puffo_agent/portal/worker.py
@@ -10,6 +10,7 @@ import asyncio
 import json
 import logging
 import os
+import re
 import shutil
 import sys
 import time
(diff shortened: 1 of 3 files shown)


src/puffo_agent/portal/worker.py
```diff
@@ -320,6 +321,96 @@ def _build_puffo_core_client(
...
+        "agent %s: suppressed worker-error leak in %s reply: %s",
+        agent_id, scope, reply[:200],
+    )
+    if treat_auth_as_health and _looks_like_auth_error(reply):
+        runtime.health = "auth_failed"
+    if scope == "api-error-retry":
+        runtime.error = (
+            "Worker emitted an auth/rate-limit error string after an "
+            "API error; suppressed from channel post. Operator should "
+            "check Claude CLI auth and re-run "
```
hellohanchen: The `api-error-retry` scope's `runtime.error` always tells the operator to **"check Claude CLI auth"** — but this scope fires for both auth-class leaks (`Not logged in`, `authentication_error`) AND rate-limit-class leaks (`rate_limit_error`, kick-text echo). For a rate-limit-class leak there's nothing wrong with auth; the operator action is to wait (or kill the offending traffic source), not re-login.

The helper already calls `_looks_like_auth_error(reply)` four lines up. Could branch on that result here to produce two variants of the message:

```python
if scope == "api-error-retry":
    if _looks_like_auth_error(reply):
        runtime.error = (
            "Worker emitted an auth error string after retrying an "
            "API error; suppressed from channel post. Check Claude CLI "
            f"auth and re-run `puffo-agent agent resume {agent_id}`."
        )
    else:
        runtime.error = (
            "Worker emitted a rate-limit error string after retrying "
            "an API error; suppressed from channel post. Usually "
            "self-recovers; investigate if persistent."
        )
```

Doesn't affect the suppression itself — just makes the operator-facing copy point at the right problem.

src/puffo_agent/portal/worker.py
```diff
@@ -646,11 +737,19 @@ async def on_message_batch(
             # display still reads as "N messages processed."
             self.runtime.msg_count += 1
             self.runtime.last_event_at = int(time.time())
-            if reply:
+            if reply and not _handle_suppressed_reply(
+                reply,
+                self.runtime,
+                agent_id,
+                scope="fallback",
+                treat_auth_as_health=False,
```
hellohanchen: Curious about `treat_auth_as_health=False` here. The `on_message_batch` (fallback) scope is the *normal* reply path — if a `Not logged in` line leaks here, that's also definitive evidence of an auth failure (Claude CLI emitted it because OAuth died), so flipping `runtime.health=auth_failed` is just as useful as in the retry scope.

Is this intentional (empirically, auth errors only manifest at retry-scope because the SDK swallows them on the first attempt and only echoes them after a retry kick) or is it an oversight from the api-error-retry being the case-study site?

If intentional, worth a one-line comment on the call site so future readers don't have to re-derive it. If oversight, `treat_auth_as_health=True` in both places makes the behaviour symmetric.

src/puffo_agent/portal/worker.py
```diff
@@ -320,6 +321,96 @@ def _build_puffo_core_client(
...
+    re.compile(
+        r"^\s*Not logged in[\s\S]*Please run /login",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"^\s*\[puffo-agent system message\]\s+session errored on rate",
+        re.IGNORECASE,
+    ),
+    re.compile(r"\bauthentication_error\b", re.IGNORECASE),
+    re.compile(r"\brate[_ -]limit[_ -]error\b", re.IGNORECASE),
```
hellohanchen, line 358: Just confirming the trade-off you flagged in the PR description — the space variant in `[_ -]` catches the literal English phrase "rate limit error" anywhere in a message. Agent prose like "I got a rate limit error from the API, retrying in a moment" would be suppressed and surface as `runtime.error` instead of going to the channel.

Given FB-105's recurrence volume, suppressing-the-occasional-prose is the safer side of the trade. Agreed.

If you ever want to tighten: keep `[_-]` only and add a separate `rate limit error` pattern that's only matched at the start of the message (anchor with `^`) — that catches the SDK error format (which always appears at start of the leaked line) without catching mid-message prose. But it's a separate exercise; this PR's choice is reasonable as-is.

hellohanchen (commented):
Solid worker-layer egress filter. The pattern set is anchored to real signatures, the four leak strings come from named live incidents (FB-105 / FB-88 / FB-159), and the test matrix covers both positive matches AND the explicit overreach guards Equation flagged. The `_RecordingClient` test that mirrors the exact `if reply and not _handle_suppressed_reply(...): await send_fallback_message(...)` shape from `Worker._run()` is the strongest test in the suite — any future call-site refactor that forgets the guard will break it before prod.

Defense-in-depth framing with PUF-207 (startup verify) and PUF-213 (adaptive refresh) is the right way to think about this — three independent layers, this one being the loudest backstop.

Three non-blocking observations inline, two of which are about wording accuracy rather than behaviour. Plus one note on the rate-limit space variant which you already flagged in the PR description.

✅ from me, will defer to your final review.

hellohanchen:
@hellohanchen asked me to fold two follow-ups into this PR before merge — both grew out of a colleague hitting `You've hit your limit` in production yesterday, which the current 4-pattern set misses entirely.

Pattern recommendations below come from the official [Claude Code error reference](https://code.claude.com/docs/en/errors) (every error string the CLI emits, mapped to a recovery section) and the [Claude API error reference](https://platform.claude.com/docs/en/api/errors) (10 canonical `<type>_error` identifiers). Not invented — quoted verbatim from docs and the column "Message" in the docs' error-lookup table.

---

## 1. Extend the pattern list (docs-driven, not guessed)

### Usage limits — the colleague's case
Subscription-plan quotas surface verbatim per docs §"You've hit your session limit":

```python
# Session / weekly / Opus / Sonnet / "<anything> limit" variants
re.compile(r"^\s*You've hit your\b.*?\blimit\b", re.IGNORECASE),
re.compile(r"^\s*Credit balance is too low\b", re.IGNORECASE),
```

### Server-side 429 / 5xx — distinct from API error identifiers
Per docs §"Server is temporarily limiting requests" and §"API Error: Repeated 529 Overloaded":

```python
re.compile(r"\bAPI Error: Request rejected \(429\)\b", re.IGNORECASE),
re.compile(r"\bAPI Error: Server is temporarily limiting requests\b", re.IGNORECASE),
re.compile(r"\bAPI Error: Repeated 529 Overloaded errors\b", re.IGNORECASE),
re.compile(r"\bAPI Error: 500\b[\s\S]*Internal server error\b", re.IGNORECASE),
```

### OAuth / auth recovery paths (additional auth-class)
Per docs §"OAuth token revoked or expired" and §"Invalid API key" / §"This organization has been disabled":

```python
re.compile(r"^\s*OAuth token (?:revoked|has expired)\b", re.IGNORECASE),
re.compile(r"^\s*Invalid API key\b", re.IGNORECASE),
re.compile(r"\bThis organization has been disabled\b", re.IGNORECASE),
```

These three should also be added to `_looks_like_auth_error` so they flip `runtime.health=auth_failed` (same recovery story as the existing `Not logged in` + `authentication_error`).

### API error identifiers (Anthropic API canonical 10)
Per platform docs §"HTTP errors". We already match `authentication_error` + `rate_limit_error`; add the rest of the safe-to-anchor identifiers:

```python
re.compile(r"\boverloaded_error\b", re.IGNORECASE),
re.compile(r"\bbilling_error\b", re.IGNORECASE),
re.compile(r"\bpermission_error\b", re.IGNORECASE),
re.compile(r"\btimeout_error\b", re.IGNORECASE),
```

**Skip these** — too generic, real-FP risk too high for the volume they'd catch:
- `\binvalid_request_error\b` — agent prose discussing API design could emit it
- `\bnot_found_error\b` — generic phrase
- `\bapi_error\b` — extremely common token, would suppress prose like "If you hit an api_error, retry"

### Skip-list (do NOT add — high FP risk per docs review)
- `Prompt is too long` — agent might explain context limits in prose
- `Request timed out` — common English phrase
- `Unable to connect to API` — agent might describe network states
- `Image was too large`, `PDF too large` — descriptive prose

---

## 2. Randomized backoff (15–60s) instead of auto-pause

Han's framing: auto-pause has too high a recall-risk (a single leak shouldn't ground the agent). Instead, sleep a random 15–60s on detect — drops the error frequency ~100x without taking the agent fully offline. If the underlying condition clears during the sleep, the next batch goes through clean; if it doesn't, we leak again and sleep again, but at 1/30s ~ rate, not at message rate.

### Sketch

```python
import random

# Make the helper return the backoff (0 on clean prose, [15, 60] uniform on suppress).
def _handle_suppressed_reply(
    reply, runtime, agent_id, *, scope, treat_auth_as_health,
) -> tuple[bool, float]:
    safe_reply = _suppress_worker_error_leak(reply)
    if safe_reply is not None:
        return False, 0.0
    # ... existing log + runtime.error + health-flip + save ...
    backoff = random.uniform(15.0, 60.0)
    logger.info(
        "agent %s: suppressed leak in %s — backing off %.1fs before next batch",
        agent_id, scope, backoff,
    )
    return True, backoff

# At both call sites in Worker._run:
suppressed, backoff = _handle_suppressed_reply(
    reply, self.runtime, agent_id,
    scope="fallback",  # or "api-error-retry"
    treat_auth_as_health=False,  # or True
)
if reply and not suppressed:
    await client.send_fallback_message(channel_id, reply, root_id=root_id)
elif backoff > 0:
    await asyncio.sleep(backoff)
```

### Behavioural notes worth thinking through before implementing

- **Heartbeat keeps firing** during the sleep — it's a separate coroutine (`asyncio.create_task(heartbeat())` in `_run`). Good: `puffo-agent status` stays current.
- **Inbound messages during sleep** — they queue at the WS layer; once `sleep` returns, the worker picks up the next batch. If the agent is in a hot loop hitting the same limit, the next batch will leak again and sleep again — steady state is one leak per ~37s. That's the design intent.
- **Backoff escalation** — defer for v1. Constant 15–60s uniform is fine. If empirically a single 30s sleep doesn't clear the underlying limit window, add jittered exponential (cap 5min) in a follow-up.
- **Operator's `puffo-agent status` view** — already shows `runtime.error` + `runtime.health=auth_failed`. Adding `last_backoff_at` to RuntimeState would make the backoff visible in status too, but that's a polish ask, not a blocker.

### Tests to add

- `test_handle_suppressed_reply_returns_backoff_in_range` — suppressed → backoff ∈ [15.0, 60.0]
- `test_handle_suppressed_reply_returns_zero_backoff_on_clean_prose`
- `test_call_site_sleeps_after_suppression` — monkeypatch `asyncio.sleep` to capture the duration, assert it's in range and that `client.send_fallback_message` is NOT called
- `test_call_site_no_sleep_on_clean_reply` — monkeypatched sleep should not be called

---

## Suggested ordering

1. Pattern additions first (small, mechanical, fully testable)
2. Backoff change second (touches the call-site signature — keep it a separate commit for review)
3. Both land together in PUF-214 before merge; PR description gets a "Closes the colleague-hit `You've hit your limit` gap" line

Doc citations again for the record:
- [Claude Code error reference](https://code.claude.com/docs/en/errors) — message-to-recovery table is the source of truth for all CLI-emitted strings
- [Claude API errors](https://platform.claude.com/docs/en/api/errors) — the 10 canonical `<type>_error` identifiers

hellohanchen (commented):
Great round-2 work. Everything I flagged is in, executed well, and Calculation caught two things I missed:

**Implementation audit:**
- All 12 doc-cited patterns landed verbatim from the audit; usage limits / 429 wrappers / OAuth recovery / API `<type>_error` identifiers all covered.
- Skip-list (`invalid_request_error`, `not_found_error`, `api_error`, `Prompt is too long` etc.) deliberately excluded per the audit — and there's a parametrized regression guard test for it.
- `_AUTH_ERROR_PATTERNS` extracted as a shared tuple between `_looks_like_auth_error` and `_WORKER_ERROR_LEAK_PATTERNS` — single source of truth, cleaner than what I sketched. Three new auth-class patterns now extend both surfaces from one place.
- Backoff returns `(suppressed, backoff_seconds)`, call sites unpack and `await asyncio.sleep(backoff)` on suppression. Module-level `_SUPPRESSION_BACKOFF_MIN/MAX_SECONDS` constants exported so tests pin against them — change the constants, tests follow automatically.
- Log line now includes `(backoff %.1fs)` so the throttling is visible in daemon logs.

**Catches I missed in the original audit:**
- Trailing `\b` on `\bAPI Error: Request rejected \(429\)\b` was unreachable — `)` is non-word and the following char (space / em-dash) is also non-word, so the boundary never has a transition to fire on. Dropping it doesn't reduce specificity (the `\bAPI Error:` prefix carries all the anchoring) and removes a phantom never-true check. Nice catch.

**Test quality:**
- 16-case parametrized positive-match for the new patterns + 8-case skip-list negatives → strongest pattern coverage in the file.
- Backoff distribution smoke at 100 samples asserts both range AND `len(set(samples)) >= 90` — catches a degenerate-random regression.
- New `test_call_site_sleep_intercepts_send_under_real_asyncio` monkeypatches `asyncio.sleep` to a no-op coroutine and exercises the production-shape harness end-to-end. This is the strongest test in the file now.

**One non-blocking nit** — `test_round2_skip_list_passes_through` has an `if "timeout_error" not in prose:` guard that silently skips the assertion for the `"Discussed timeout_error handling..."` case. The case IS suppressed (correctly — `\btimeout_error\b` is the anchored pattern, the prose just happens to use the literal identifier in a sentence about API handling). The test's parametrize includes it but doesn't assert anything for it, so it's effectively a no-op invocation. Either parametrize on `(prose, expect_suppress)` and branch the assertion, or drop the timeout_error case from the list. Doesn't matter for correctness; just makes the test honest about what it's testing.

---

✅ from me. Closes the colleague's `You've hit your limit` gap + structural fix for the tight-loop leak frequency in one PR. CI green. Ready for your final review + merge.

PR #27, opened 2026-05-18
fix(PUF-206): pre-check Python >= 3.11 at cli.py module load

hellohanchen:
## Summary

Users running `puffo-agent` against Python 3.10 / 3.9 hit a deep `SyntaxError` / `ImportError` from inside the submodule chain at `cli.py:21-49` — the actual cause (wrong Python version) was buried under the wrong-looking message, and users frequently mis-attributed it as "my Python is broken" (Shiva, FB-149).

This PR adds a `_require_python_311()` precheck at the top of `cli.py`, called at module load between `from __future__ import annotations` and the first submodule import. On Python <3.11 it writes a single actionable line to stderr — naming the detected version + three concrete upgrade paths (pyenv / Homebrew / python.org) — and exits with code 1. On 3.11+ it returns silently and the rest of the module loads normally.

`pyproject.toml`'s `requires-python = ">=3.11"` is still the canonical gate; this is the runtime safety net for users on venvs whose interpreter doesn't match the metadata.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-206.md`

## Test plan

- [x] Full pytest: **586 passed / 1 intentional skip / 0 failed** (582 pre-existing + 4 new).
- [x] 4 new tests in `tests/test_cli_python_version.py` mock `sys.version_info` + `sys.stderr` and cover the matrix: rejects 3.9.18, rejects 3.10.12, passes 3.11.0, passes 3.14.4. Reject branches assert on exit code, version-in-message, and presence of at least one of the upgrade-path hints.
- [x] Error message is ASCII-only (`>= 3.11`, no `≥`) per Equation's Windows-terminal compatibility nit.
- [x] Independent QA with one iteration to slim the helper docstring.

## Out of scope

- Pre-install hook / install-script side: `pyproject.toml` already pins via `requires-python`; this is the runtime safety net.
- Auto-detect-and-suggest a `python3.12` on PATH: explicitly post-Tuesday per the ticket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,24 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   code 1 + version-in-message + presence of at least one of
   `pyenv` / `brew` / `python.org` upgrade hints. (PUF-206)
 
+- **PUF-217 test fixture: `os.rename` → `os.replace` + symlink-skip
+  guard for Windows.** The PUF-217 test fixture's fake-claude
+  subprocess used `os.rename(tmp_target, target)` to mimic Claude
+  CLI's atomic tmp+rename. On POSIX `os.rename` atomically replaces
+  an existing target, but on Windows it raises `FileExistsError`;
+  the cross-platform atomic-replace primitive is `os.replace`. Two
+  `os.rename` call sites in `test_refresh_oneshot_home_env.py`
+  swapped to `os.replace`. Additionally, the two tests that assert
+  on `agent_creds.is_symlink()` now gate on
+  `_symlinks_available(tmp_path)` and `pytest.skip` when the host
+  can't create symlinks — mirrors the existing pattern in
+  `test_host_credentials.py` (Windows without Developer Mode falls
+  through `link_host_credentials`'s copy-mode branch, so the
+  symlink-distribution invariant these tests assert on doesn't
+  apply there). Linux CI continues to exercise the full assertions;
+  Windows now skips cleanly instead of failing. Full suite:
+  **642 passed / 9 skipped / 0 failed**.
+
 ## [0.8.5] — 2026-05-18
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,48 @@ All notable changes to `puffo-agent` are documented in this file. The
 format follows [Keep a Changelog](https://keepachangelog.com/) and
 this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [0.8.6] — 2026-05-18
+
+### Fixed
+
+- **Python-version precheck moved to `puffo_agent/__init__.py` and
+  fires before any submodule of the package is parsed.** Users on
+  Python 3.9 / 3.10 used to see a deep `SyntaxError` / `ImportError`
+  from inside the submodule chain at `cli.py`'s import block — the
+  actual cause (wrong Python version) was buried under the
+  wrong-looking message, and users frequently mis-attributed it as
+  "my Python is broken" (Shiva, FB-149).
+
+  Initial fix (0.8.5-rc, PR #27) added the precheck at the top of
+  `cli.py`. That covered the dominant case but had a subtle
+  weakness: Python parses an entire module before executing any line,
+  so if `cli.py` itself ever used Python 3.11-only syntax (`match` /
+  `case`, PEP 604 unions in expression position, exception groups)
+  the file would fail at parse time **before** the precheck had a
+  chance to run, and the user would be back to "deep SyntaxError"
+  from inside `cli.py`.
+
+  Move it one layer up: `puffo_agent/__init__.py` runs first when
+  Python resolves any `puffo_agent.*` submodule import — including
+  the entry-point's `from puffo_agent.portal.cli import main`. The
+  precheck now fires **before `cli.py` is even parsed**, so a future
+  3.11-only edit to `cli.py` (or any other submodule) can't bypass
+  the guard. `puffo_agent/__init__.py` itself stays deliberately
+  parseable on Python 3.6+ (f-strings only, no PEP 604, no `match`)
+  — documented in the module docstring so the constraint isn't lost.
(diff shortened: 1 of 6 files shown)


hellohanchen (commented):
Clean, well-scoped fix. The placement at the top of `cli.py` between `from __future__ import annotations` and the first submodule import is the right call — submodule imports are where the deep `SyntaxError` / `ImportError` chain originates, and this precheck intercepts before any of them load.

**Implementation audit:**
- Message is ASCII-only (`>= 3.11`, no `≥`) per the Windows-terminal nit.
- Three concrete upgrade paths (pyenv / brew / python.org) — covers the dominant install surfaces. `uv` / `mise` / `conda` missing but listing them all gets noisy; current set is fine.
- Recommends `3.12` not `3.11` in the install commands — sensible middle ground (3.12 is the most-supported, 3.13 is recent).
- `SystemExit(1)` + `stderr.write` is the right shape — `print(..., file=sys.stderr)` would also work, but `write` is one less syscall and the explicit `\n` is honest about what's going on the wire.
- `import sys` correctly moved to the top of the file (between `from __future__` and the precheck) and removed from the standard-library import block below.

**Test coverage:**
- 4 cases cover the matrix (3.9 reject, 3.10 reject, 3.11 pass, 3.14 pass). `_VersionInfo` tuple-subclass with `major`/`minor`/`micro` named attrs handles both the tuple comparison and the message-formatting access correctly.
- Asserts version-in-message + at-least-one-upgrade-hint, which is the right shape (defensive against future message rewording without losing the "concrete next step" guarantee).

---

**One non-blocking observation** worth noting in the team's mental model:

This precheck protects against `SyntaxError` / `ImportError` from **submodules imported by `cli.py`**, but not against `cli.py` itself becoming unparseable on Python 3.9 / 3.10. Python parses the whole module before executing any line — so if a future edit adds e.g. a `match/case` block, a PEP 604 union in expression position (`x: int | str = 5` without `from __future__ import annotations` cover), or an exception group, Python 3.9 would fail at parse time and the precheck wouldn't fire.

Today `cli.py` is parseable on 3.9 (annotations are lazy under `from __future__ import annotations`, no `match`, no `|`-unions in expression position). The discipline of "keep `cli.py` body 3.9-parseable" needs to be maintained manually — there's no automated guard.

**Stronger placement** would be `src/puffo_agent/__init__.py` (currently empty). Python imports the package init *before* parsing the submodule, so the precheck fires even if `cli.py` itself becomes unparseable. Five-line addition; doesn't conflict with this PR. Worth considering as a follow-up if the team ever needs the precheck to be bulletproof — but the current placement covers the dominant case and is fine to ship as-is.

---

✅ from me. Closes the FB-149 / Shiva mis-attribution surface; ready for merge whenever you're ready.

hellohanchen:
Self-applied the non-blocking observation from the round-1 review + bumped to 0.8.6. Want your eyes on it before merge.

**What changed in 76b822b:**
- Precheck moved from `cli.py` to `puffo_agent/__init__.py`. Python loads the package init *before* parsing any submodule, so the guard now fires before `cli.py` is even parsed — a future 3.11-only edit anywhere in any `puffo_agent.*` submodule can't bypass it.
- `puffo_agent/__init__.py`'s module docstring spells out the discipline ("stays parseable on Python 3.6+; no PEP 604 in expression position, no `match/case`") so the constraint isn't lost.
- `cli.py` cleaned up: precheck function + call removed, `import sys` returned to the stdlib import block where it was originally.
- Test import path updated: `from puffo_agent import _require_python_311` instead of `from puffo_agent.portal.cli import ...`. Same 4-case matrix, all green.
- `pyproject.toml` → `0.8.6`, CHANGELOG.md gets a 0.8.6 entry explaining the move + the parse-order rationale.

**Found while running the local suite (NOT caused by this change):**
PUF-217's tests `test_refresh_oneshot_write_lands_at_host_path_visible_via_agent_symlink` + `test_refresh_oneshot_does_not_create_regular_file_at_agent_path` fail on Windows with `FileExistsError [WinError 183]` at `os.rename(tmp_target, target)`. Cause: Windows' `os.rename` doesn't atomically replace existing files — `os.replace` does. `git stash + re-run` confirmed the failures are pre-existing on the unmodified branch (CI passes because Linux `os.rename` does replace). Easy follow-up: swap `os.rename` → `os.replace` in the two fixture lines. Not blocking this PR.

CI on Linux should still be green for this PR — only changed are the precheck location, the test's import path, and the version/changelog files.

Ready for your final review + merge.

hellohanchen:
Folded the PUF-217 test-fixture Windows-portability fix into this same PR per your ask (commit `2ba9f0c`).

**Changes:**
- Two `os.rename(tmp_target, target)` call sites in `test_refresh_oneshot_home_env.py` → `os.replace`. POSIX `os.rename` atomically replaces an existing target but Windows raises `FileExistsError`; `os.replace` is the cross-platform atomic primitive. (`os.rename` was a real bug, just masked by Linux's POSIX behaviour.)
- Two tests that load-bearingly assert `agent_creds.is_symlink()` (`test_refresh_oneshot_write_lands_at_host_path_visible_via_agent_symlink` + `test_refresh_oneshot_does_not_create_regular_file_at_agent_path`) now gate on `_symlinks_available(tmp_path)` + `pytest.skip`. Mirrors the established pattern in `test_host_credentials.py` line 35. On Windows without Developer Mode, `link_host_credentials` falls through its copy-mode branch, so the symlink-distribution invariant simply doesn't apply.
- The third PUF-217 test (`test_refresh_oneshot_inherits_operator_home`) doesn't depend on symlinks and continues to run everywhere.

Local Windows results: previously 2 failed, now 0 failed + 2 skipped (with the same "symlinks unavailable on this host" message as the existing skipped tests). Linux CI continues to exercise the full assertions — `os.replace` is a no-op semantics change on POSIX, and `_symlinks_available` returns True there.

Full suite now: **642 passed / 9 skipped / 0 failed**. CHANGELOG.md 0.8.6 entry updated to mention this addition.

Ready for final review + merge.

PR #28, opened 2026-05-18
fix(PUF-207): probe Claude OAuth at agent startup, auto-pause on fail

hellohanchen:
## Summary

When an agent's persisted-session adapter spawned with a dead Claude CLI OAuth token, `Worker._run()` previously sailed past `warm()` into the first turn — at which point Claude CLI emitted `Not logged in. Please run /login` and that string leaked into the channel as if it were the agent's reply (FB-159 — Sheri / Yasushi mis-attributed as "my Python is broken"). The startup path had no signal at all that auth was broken.

This PR wires the existing `adapter.refresh_ping()` probe to fire in `Worker._run()` between init and `warm()`. If `auth_healthy` comes back explicitly `False`, the new `_check_startup_auth_or_pause(...)` helper mutates the runtime to `status=paused, health=auth_failed`, writes a multi-line recovery prompt to `runtime.error` (covers all four FB-159 case-study surfaces: separate-Terminal hint, full-path `puffo-agent` hint, `agent resume` verb, agent id in both opening prose and the command template), saves, and bails before spawning. Adapters that don't carry a credential TTL (sdk / chat-only) leave `auth_healthy=None` and proceed normally.

**Defense-in-depth pair with PUF-213 (adaptive refresh tick prevents staleness during runtime) and PUF-214 (suppress error-string leaks if either misses).** Together they close the silent-can't-recover surface for FB-159 / FB-105.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-207.md`

## Test plan

- [x] Full pytest: **587 passed / 1 intentional skip / 0 failed** (582 pre-existing + 5 new).
- [x] 5 new tests in `tests/test_worker_startup_auth.py`:
  - `auth_healthy=False` → pauses, populates message, persists to disk; recovery message includes `claude` + `/login` + `Terminal` + `puffo-agent` + `agent resume <id>`; agent id appears in **both** the opening prose and the command (`count >= 2` asserted so a future trim doesn't silently regress).
  - `auth_healthy=None` (sdk / chat-only) → proceeds, runtime untouched.
  - `auth_healthy=True` → proceeds, runtime untouched.
  - Not sticky: subsequent `auth_healthy=True` via the existing periodic tick at `worker.py:705-728` can flip back to `health=ok`.
- [x] Independent QA with one iteration to slim docstring + wiring comment + delete a dead test helper.

## Out of scope (per Equation)

- **FB-105 / PUF-214** (worker-error-suppression for the runtime-leak case where auth expires *during* a session) ships separately — already in this review queue.
- **FB-88 / PUF-213** continuous-runtime refresh — already in this review queue too.
- **Install-time check (FB-151, P2)** — different timing window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,58 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   Windows now skips cleanly instead of failing. Full suite:
   **642 passed / 9 skipped / 0 failed**.
 
+### Added
+
+- **Agent startup auto-pauses when the OAuth probe says "auth is
+  dead."** When a cli-local agent's persisted session re-spawned
+  with an expired or revoked OAuth token, `Worker._run()` used to
+  proceed straight into `warm()` and the first turn — at which
+  point Claude CLI emitted `Not logged in. Please run /login` and
+  that string leaked into the channel as if it were the agent's
+  reply (FB-159 / Sheri / Yasushi mis-attributed as "my Python is
+  broken"). The startup path had no signal at all that auth was
+  broken before the first message hit production.
+
+  Now `Worker._run()` fires `adapter.refresh_ping()` between init
+  and `warm()`. The new `_check_startup_auth_or_pause()` helper
+  reads `adapter.auth_healthy` afterwards:
+  - `None` (sdk / chat-only adapters — no credential file) or
+    `True` (probe succeeded) → proceed unchanged.
+  - `False` (probe reported auth failure) → mutate the runtime to
+    `status=paused, health=auth_failed`, populate `runtime.error`
+    with a four-step recovery prompt (open a separate Terminal,
+    not from within an agent's own shell, run `claude /login`,
+    then `puffo-agent agent resume <id>`; mentions the venv full-
+    path caveat), persist, and bail before `warm()` ever spawns
+    the doomed claude subprocess. `self._warm_done.set()` is
+    still called on the early-return path so any caller awaiting
+    `wait_for_warm()` doesn't hang.
+
+  Not sticky: the existing periodic refresh tick at
+  `worker.py:737-748` can flip `runtime.health` back to `ok` if
+  the operator re-authenticates and the next probe succeeds. The
+  operator still needs to manually `puffo-agent agent resume <id>`
+  (auto-resume would risk a tight pause/resume loop if auth
+  flaps); the message in `runtime.error` is the recovery script.
+
+  Defense-in-depth triad with PUF-213 (adaptive refresh tick
+  prevents staleness during runtime) and PUF-214 (worker-layer
+  error-string leak suppression if either misses) — together they
+  close the silent-can't-recover surface for the FB-159 / FB-105
+  class.
+
+  Tests in `test_worker_startup_auth.py` (7 cases): the helper
+  matrix (`auth_healthy=False` pauses with the full recovery
+  prompt incl. agent id ≥2 occurrences; `None` / `True` proceed;
+  persistence to disk; not-sticky on probe-success), plus two
+  call-site tests using a `_startup_call_site` helper that
+  mirrors the exact pause-or-warm-then-set production block from
+  `Worker._run()` — asserts `warm_done.is_set()` on BOTH the
+  pause early-return path AND the proceed-then-warm path, catching
+  any future refactor that drops the `self._warm_done.set()` line
+  before the early return. Same call-site-mirroring pattern as
+  PUF-214's `_fallback_call_site` helper. (PUF-207)
+
 ## [0.8.5] — 2026-05-18
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


hellohanchen (commented):
Clean defense-in-depth fix. Wiring `refresh_ping` between init and `warm()` is the right cut — `warm()` is the heavy operation (spawns the claude subprocess), so probing first means we don't burn that startup cost when auth is already known dead. The `auth_healthy: bool | None` tri-state contract from `base.py:73` matches the helper's `is not False` guard precisely; sdk / chat-only adapters that leave `auth_healthy=None` correctly proceed.

**Implementation audit:**
- Call site at `worker.py:519-523` follows the established "_warm_done.set() on early-return" convention from the init-failure path at line 505-506. Callers waiting on warm-done won't hang.
- Recovery message hits all four FB-159 case-study surfaces (separate-Terminal hint, full-path `puffo-agent` hint, explicit `agent resume` verb, agent id in prose + command template).
- `try/except Exception` around `refresh_ping()` correctly degrades to "proceed with `auth_healthy` unchanged" on transient transport errors — the periodic tick at line 737-748 catches it later. `CancelledError` propagates as expected (it's a `BaseException`).
- Not sticky — auto-pause only sets `status=paused` + `health=auth_failed` + `error`; the periodic tick's flip-to-`ok` path is untouched. `test_check_startup_auth_not_sticky` smoke-tests this contract.

**Test quality:**
- 5 cases cover the full helper matrix (False / None / True / disk persistence / not-sticky).
- `runtime.error.count("test-agent-001") >= 2` assertion is clever defensive cover — guards against a future "trim duplicate id" refactor silently losing one of the agent_id surfaces.
- Tests mock `_FakeAdapter` to just carry `auth_healthy`, mirroring the helper's narrow read of the adapter contract. Right abstraction level.

---

**Four non-blocking observations:**

1. **`refresh_ping()` can block startup for up to 120s** (per `REFRESH_ONESHOT_TIMEOUT_SECONDS` in `local_cli.py`). On a network blip at daemon boot, every cli-local agent waits the full timeout before warm() (or fail-open). Worst case under transient network issue: 10 agents × 120s if they serialise, or a single 120s window if they run concurrently (which they do — `Worker._run` is async). Not a blocker since `Worker._run` is async — but worth noting that startup feels slower than before. A follow-up could fire the probe in the background and gate only the *first message* on its completion. Out of scope here.

2. **With PUF-217 merged, `refresh_ping` now activates the operator's MCP servers.** Adds 2-5s startup latency per agent for the probe leg. Same caveat already flagged in PUF-217's PR description. The `--strict-mcp-config` follow-up there will also benefit this path when it lands.

3. **Call-site test gap.** The 5 helper tests cover `_check_startup_auth_or_pause`'s return value + state mutations, but there's no test asserting `self._warm_done.set()` actually fires at the call site on the pause path. If a future refactor accidentally drops the `self._warm_done.set()` line at `worker.py:522`, callers awaiting `wait_for_warm()` (line 380) would hang indefinitely. PUF-214 added a `_RecordingClient`-style test mirroring the production-shape `if/else` block; the same pattern here would catch the regression. Marginal — won't block.

4. **Recovery message: "(not the Codex shell)"** is specific to one harness scenario. The hazard the line is warning against is "don't run `claude /login` from inside an agent's own subprocess context" — which applies to `cli-local` claude / hermes / gemini-cli / codex equally. Slightly more general wording: "(not from within an agent's own shell)" or "(don't run this from inside another puffo-agent subprocess)". Pure copy nit — current wording is fine if your operator population is mostly Codex users.

---

✅ from me. Defense-in-depth with PUF-213 (runtime refresh) + PUF-214 (egress leak suppression) closes the silent-can't-recover surface end-to-end. Ready for final review + merge whenever you're set.

hellohanchen:
Re-reviewed 0032056. Both addressed:

**(3) Call-site warm-gate test (load-bearing) — done.** `_startup_call_site` helper mirrors the exact production shape from `Worker._run()`:
```
if not _check_startup_auth_or_pause(...):
    warm_done.set()
    return
try:
    await warm(...)
finally:
    warm_done.set()
```
Two tests assert `warm_done.is_set()` on **both** the pause early-return path AND the proceed-then-warm path. Catches the regression I flagged — if a future edit drops the `self._warm_done.set()` line on the pause branch, `wait_for_warm()` callers would hang indefinitely. Same pattern as PUF-214's `_fallback_call_site` — consistent codebase style.

**(4) Recovery wording generalised — done.** `(not the Codex shell)` → `(not from within an agent's own shell)`. Covers cli-local / hermes / gemini-cli / codex uniformly. Pure copy change, no behavioral effect.

**(1) startup probe can block up to 120s** and **(2) MCP startup cost from operator's config** — explicitly out of scope per my round-1 framing, intentionally not touched. PR description's PUF-218 deferred follow-up + `--strict-mcp-config` follow-up cover both.

CI green. ✅ from me, ready for final review + merge.

PR #29, opened 2026-05-18
fix(PUF-213): scale credential_refresh tick to TTL so the window can't be missed

hellohanchen:
## Summary

The worker's `credential_refresh()` loop slept a fixed 10-minute tick between probes. When a Claude CLI / Anthropic OAuth access token's lifetime was short enough that its refresh-window (5 minutes before expiry) fell entirely inside one tick interval, the loop could skip refresh entirely — token expired, next turn 401'd, the FB-88 silent-can't-recover surface. Anthropic OAuth access tokens at ~1h TTL invalidated the literal "24h before expiry" framing from the canonical FB-88 ticket too.

This PR replaces the fixed sleep with an adaptive `_next_refresh_tick(expires_in)` helper that scales the sleep to the token's remaining TTL: far from expiry → default 10-minute heartbeat; close enough that the next tick should land inside the refresh window → wake `threshold` seconds before expiry with margin; inside or past the window → 60-second floor to bound retry without dogpiling. `None` TTL (sdk / chat-only adapters with no credentials file) → default tick — the loop is then a pure health heartbeat.

**Defense-in-depth triad with PUF-207 (startup OAuth verify) + PUF-214 (worker-error suppression).** None subsumes the others; this one prevents staleness during runtime, PUF-207 catches it at startup, PUF-214 catches the leak if either misses.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-213.md`

## Test plan

- [x] Full pytest: **590 passed / 1 intentional skip / 0 failed** (582 pre-existing + 8 new).
- [x] 8 new tests in `tests/test_credential_refresh_policy.py` exhaustively cover the helper:
  - `None` TTL → `default_tick` (sdk / chat-only path).
  - Far future TTL → `default_tick` (heartbeat regime).
  - TTL just above the window → `expires_in - threshold` (wake with margin).
  - TTL exactly at the window → `floor` (clamped).
  - TTL below the window → `floor`.
  - Negative TTL (already expired) → `floor` (bounded retry).
  - Custom-thresholds parametric: pins production constants out of the assertion so a future threshold tune doesn't shift these tests.
  - `default_tick` cap on long horizon — even when `expires_in - threshold` exceeds default, return `default_tick`.
- [x] Helper params default to module constants per Equation's call-site simplification ask — production call site is just `_next_refresh_tick(expires_in)`.
- [x] Call-site wiring at the existing `credential_refresh()` loop wraps `adapter._credentials_expires_in_seconds()` in try/except → `None` so any adapter-side error gracefully degrades to the default tick.
- [x] Independent QA with one iteration to slim docstring (let the 8-test matrix be the spec).

## Out of scope

- **5+ day soak AC** — out of unit-test scope; will land as a follow-up PUF once we have a live cohort or a fake-time / Freezegun harness.
- **FB-105 / PUF-214** runtime-leak suppression — separate ticket, also in this review queue.
- **Refresh-token-itself expiry (30+ days)** — falls through to PUF-207's pause-with-recovery path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,60 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   Windows now skips cleanly instead of failing. Full suite:
   **642 passed / 9 skipped / 0 failed**.
 
+- **`credential_refresh` worker loop now scales its sleep to the
+  token's TTL instead of a fixed 10-minute tick.** Pre-fix, the
+  worker slept a flat `CREDENTIAL_REFRESH_TICK_SECONDS = 10 * 60`
+  between probes. When a Claude CLI / Anthropic OAuth access token's
+  remaining lifetime was short enough that its refresh window —
+  `CREDENTIAL_REFRESH_BEFORE_EXPIRY_SECONDS = 5 * 60` from
+  `base.Adapter` — fell entirely inside one tick interval, the loop
+  could skip the refresh entirely. Token expired, next turn 401'd,
+  the FB-88 silent-can't-recover surface. The classic case: probe
+  at T sees TTL=600s → "above 300s threshold, skip"; next tick at
+  T+600 sees TTL=0 — refresh window was T+300 → T+600, fully
+  swallowed.
+
+  New `_next_refresh_tick(expires_in)` helper in `worker.py`
+  computes the next sleep adaptively:
+  - `None` TTL (sdk / chat-only adapters without a credentials
+    file) → fall back to `default_tick = 10 min` — the loop is a
+    pure health heartbeat in this mode.
+  - TTL far in the future → `default_tick` (capped — above the
+    refresh window the loop is just a slow heartbeat).
+  - TTL just above the window → wake `threshold` seconds before
+    expiry so the next tick lands inside the refresh window with
+    margin. Concretely: TTL=600s, threshold=300s → next tick =
+    600 - 300 = 300s, so the helper wakes at TTL=300s right at
+    the threshold — refresh fires.
+  - TTL inside the window OR negative (already expired) → clamp
+    to `CREDENTIAL_REFRESH_TICK_FLOOR_SECONDS = 60` so a
+    sustained refresh failure doesn't dogpile `_REFRESH_LOCK`,
+    but we still tick fast enough to retry promptly.
+
+  The helper is a pure function with module-constant defaults —
+  production call site is just `_next_refresh_tick(expires_in)`;
+  tests override `default_tick` / `threshold` / `floor` to pin
+  synthetic values that decouple them from production-constant
+  drift. The `try / except Exception → None` wrap around
+  `self._adapter._credentials_expires_in_seconds()` at the call
+  site gracefully degrades to `default_tick` if an adapter hook
+  ever throws.
+
+  8 new tests in `test_credential_refresh_policy.py`: None /
+  far-future / just-above-window (returns pre-window margin) /
+  at-window (clamps to floor) / below-window (clamps to floor) /
+  already-expired (clamps to floor) / default-cap on long horizon
+  / parametric custom-thresholds. Adapter-level integration
+  (entire `credential_refresh` loop end-to-end against a real
+  clock) deferred to a 5+ day soak as the PR description notes.
+
+  Defense-in-depth triad with PUF-207 (startup OAuth probe +
+  auto-pause) and PUF-214 (worker-error suppression at the egress
+  boundary). PUF-213 prevents the refresh window from being
+  silently missed during runtime; PUF-207 catches at startup;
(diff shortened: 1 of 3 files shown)


hellohanchen (commented):
Clean fix for the canonical FB-88 "10-min sleep can swallow the entire refresh window" bug. Math walks through cleanly: at TTL=600 with threshold=300, the helper returns `target=300` → wake at TTL=300 (right at the threshold), refresh fires; at TTL=900 the helper caps at `default_tick=600` → wake at TTL=300 with one extra default-tick budget. No edge where the refresh window falls entirely between two sleeps.

**Implementation audit:**
- `target < floor` clamp prevents busy-loop inside the window; `target > default_tick` clamp keeps the loop a pure heartbeat above the window. Monotonic, bounded.
- `None` TTL handling (sdk / chat-only adapters) correctly falls through to `default_tick` — confirmed by `_credentials_expires_in_seconds` returning `None` on the base class as documented at `base.py:197-203`.
- Negative TTL (already-expired token) → floor → bounded retry once a minute. Won't dogpile a sustained outage.
- `try/except Exception` around `self._adapter._credentials_expires_in_seconds()` correctly degrades to `None` → default tick if the adapter's hook throws — same fail-safe philosophy as PUF-207.
- Adaptive recomputation accounts for refresh-ping latency naturally: if the probe takes 30s, the next `expires_in` reading is 30s smaller, so the next sleep adjusts. No drift accumulation.

**Test quality:**
- 8 cases cover the full piecewise function: None, far-future cap, just-above-window margin, at-window floor, below-window floor, already-expired floor, default-cap, custom-thresholds parametric.
- The custom-thresholds test pins synthetic values (7/2/10/1, 1/2/10/1, etc.) decoupling the test from production-constant drift. Pattern is reusable — nice.
- Production call site is just `_next_refresh_tick(expires_in)` per Equation's simplification ask; tests pass overrides directly. Right ergonomics.

---

**Three non-blocking observations:**

1. **`self._adapter._credentials_expires_in_seconds()` reads an underscore-prefixed method.** It's documented as a subclass hook on `base.Adapter:197` (with a default returning `None`), so the underscore is misleading — it's part of the base API surface, not implementation detail. The reach across the layer boundary is fine semantically, just visually surprising. A cleanup PR could rename to `credentials_expires_in_seconds` (drop the underscore) to make the public-hook intent explicit. Not in scope for this PR.

2. **Lazy import comment is slightly thin.** At `worker.py:59` the helper does `from ..agent.adapters.base import CREDENTIAL_REFRESH_BEFORE_EXPIRY_SECONDS` inside the function body with `# Lazy import keeps this helper unit-testable.` But `worker.py` already imports `from ..agent.adapters import Adapter` at module top (line 19), so there's no circular-import risk for adding the constant import alongside. The lazy import works fine — just one extra import lookup per call (cached in `sys.modules` after the first), zero practical cost — but the "unit-testability" justification doesn't quite hold (the test file imports the constant directly without trouble). Could be moved to module-top for clarity; functionally identical.

3. **No integration-level test of the loop wiring.** The 8 helper tests are exhaustive on `_next_refresh_tick` itself, but there's no test asserting `credential_refresh()` actually passes the `expires_in` from the adapter to the helper and uses the result as the `asyncio.wait_for` timeout. PUF-214 added a `production_shape()` test for similar call-site wiring; same pattern could mirror here. That said, the credential_refresh loop is asyncio + adapter + runtime save-heavy — hard to test cleanly without fake-time machinery. The PR description acknowledges this gap as the "5+ day soak AC" follow-up. Acceptable to defer.

---

✅ from me. Defense-in-depth triad with PUF-207 (startup verify) + PUF-214 (egress leak suppression) is the right framing. With those three merged, the silent-can't-recover surface for FB-88 / FB-105 / FB-159 is fully closed. Ready for final review + merge.

hellohanchen:
Re-reviewed 92482f56. Only change since round 1 is a docstring trim — 30-line behaviour matrix collapsed to a one-line contract. Reasonable: the 8-test matrix in `test_credential_refresh_policy.py` already encodes every branch (None / far / pre-window / at-window / below-window / negative / cap / custom-thresholds), and the test names are self-documenting. Trimming was the right call.

My three round-1 non-blocking observations:
- **(1) `_credentials_expires_in_seconds` underscore prefix** — not addressed (you said out-of-scope, still agreed).
- **(2) Lazy import comment** — kept the lazy import, just compressed the comment. The "unit-testable" justification still isn't quite the reason (the test imports the constant fine without it; it's really just "no circular-import risk worth chasing for one constant"), but functionally identical and not worth blocking on.
- **(3) No integration test of the `credential_refresh` loop wiring** — explicitly deferred to the 5+ day soak AC per PR description.

None of these are blockers and you said as much in the round-1 thread. ✅ from me, ready for final review + merge.

PR #30, opened 2026-05-18
fix(PUF-217): drop HOME override in _run_refresh_oneshot

hellohanchen:
## Summary

On cli-local + Linux, `_run_refresh_oneshot` overrode `HOME` to the agent's per-agent home dir, sending claude's `tmp+rename` write through the agent's symlinked `.credentials.json`. The rename replaced the symlink with a regular file at the agent path, leaving the canonical host file stale. The next `_credentials_expires_in_seconds` tick called `link_host_credentials`, whose copy-mode fast-path saw `agent_creds.exists() and not is_symlink()` + mtime mismatch with host and ran `shutil.copy2(host_creds, agent_creds)` — the **stale host file overwrote the fresh-token agent file**. From the daemon's perspective `expiresAt` never advanced; the refresh cycle hit the 60s adaptive-cadence floor and dogpiled indefinitely. Live in-band repro happened on this Linux box at 18:30 today, mid-implementation.

This PR drops the HOME / USERPROFILE override in `_run_refresh_oneshot` so the refresh subprocess inherits the daemon's env (operator's HOME). Claude writes to `/home/<operator>/.claude/.credentials.json` directly; per-agent symlinks distribute the fresh token. The "symlink survives atomic rename writes" claim in `state.py:182-183`'s docstring is now retroactively correct because rename never targets the symlink path. Long-lived `ClaudeSession` agent subprocesses are unaffected — `_ensure_session` still sets `HOME=<agent_home>` for normal turn execution; only the short-lived refresh probe changes scope.

**Defense-in-depth context:** PUF-217 closes the **disk-write side** of the FB-88 refresh cascade. PUF-218 (deferred) will close the disk-read side (long-lived `ClaudeSession` reloads from disk after refresh). With PUF-207 (startup probe), PUF-213 (adaptive cadence), and PUF-214 (leak suppression — also in this review queue), the OAuth-lifecycle compound is fully closed on Linux.

Ticket: `/home/hanchen/.puffo-agent/shared/PUF-217.md`

## Test plan

- [x] Full pytest: **600 passed / 1 intentional skip / 0 failed**.
- [x] 3 new tests in `tests/test_refresh_oneshot_home_env.py`:
  - `test_refresh_oneshot_inherits_operator_home` — env-mutation guard: monkeypatches `asyncio.create_subprocess_exec`, captures `env` kwarg, asserts `env["HOME"]` equals the operator's HOME (and explicitly NOT the agent's home_dir).
  - `test_refresh_oneshot_write_lands_at_host_path_visible_via_agent_symlink` — end-to-end-ish: fake claude subprocess does `tmp+rename` at the env's HOME path; asserts agent symlink still resolves to a file with the fresh `accessToken` AND `_credentials_expires_in_seconds()` reads back a positive TTL.
  - `test_refresh_oneshot_does_not_create_regular_file_at_agent_path` — anti-regression: agent path remains `is_symlink()` after refresh; no stray `.credentials.tmp` at the agent path. Catches a future refactor that reintroduces a HOME override.
- [x] Independent QA on the fast-track schedule given the live cascade on operator's box.

## Operator post-deploy validation

Equation's MCP-startup-cost caveat: dropping the HOME override means the refresh subprocess now activates the **operator's** `.claude.json` MCP servers (Gmail / Drive / Calendar / Notion / PDF + any locally-installed) instead of the agent's. Expect 2–5s of MCP startup overhead per refresh. `REFRESH_ONESHOT_TIMEOUT_SECONDS = 120` so there's ample headroom, but please grep your daemon log for `refresh one-shot rc=0 in N.Ns` post-deploy to confirm the timing is in the expected range. If you see N consistently above ~10s, a `--strict-mcp-config` patch to skip MCP startup in refresh-only invocations would be a clean follow-up.

## Out of scope (PUF-218 — deferred)

Long-lived `ClaudeSession` in-memory token cache. If operator's symptom is fully closed after this lands + a fresh refresh cycle, PUF-218's urgency softens and it can land post-Tuesday. If symptom persists, PUF-218 is next-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,67 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   startup, PUF-213 prevents staleness during runtime, this catches
   the egress leak if either misses. (PUF-214)
 
+- **OAuth-refresh probe no longer clobbers the agent's credentials
+  symlink on Linux.** On cli-local + Linux, `_run_refresh_oneshot`
+  used to override `HOME` to the agent's per-agent home dir, sending
+  Claude CLI's atomic `tmp+rename` write through the agent's
+  symlinked `.credentials.json`. The rename **replaced the symlink
+  with a regular file** at the agent path, leaving the canonical
+  host file stale. The next `_credentials_expires_in_seconds` tick
+  called `link_host_credentials`, whose copy-mode fast-path detected
+  `agent_creds.exists() and not is_symlink()` plus an mtime mismatch
+  and ran `shutil.copy2(host_creds, agent_creds)` — the **stale host
+  file overwrote the fresh-token agent file**. From the daemon's
+  perspective `expiresAt` never advanced, the adaptive-cadence floor
+  kicked in at 60s, and the refresh cycle dogpiled indefinitely.
+  Live in-band repro happened on the operator's Linux box at 18:30
+  on 2026-05-18, mid-implementation.
+
+  Fix drops the `HOME` / `USERPROFILE` override in
+  `_run_refresh_oneshot` so the refresh subprocess inherits the
+  daemon's env (the operator's HOME). Claude writes to
+  `/home/<operator>/.claude/.credentials.json` directly; the
+  per-agent symlinks distribute the fresh token via read-through.
+  The "symlink survives atomic rename writes" claim in `state.py`'s
+  `link_host_credentials` docstring is now retroactively correct
+  because rename never targets the symlink path. Long-lived
+  `ClaudeSession` agent subprocesses are unaffected —
+  `_ensure_session` still sets `HOME=<agent_home>` for normal turn
+  execution; only the short-lived refresh probe changes scope.
+
+  Operational caveat (flagged for post-deploy): dropping the HOME
+  override means the refresh subprocess now activates the
+  *operator's* `.claude.json` MCP servers (Gmail / Drive / Calendar
+  / Notion / PDF + any locally-installed) instead of the agent's.
+  Expect 2–5s of MCP startup overhead per refresh.
+  `REFRESH_ONESHOT_TIMEOUT_SECONDS = 120` so there's ample headroom,
+  but watch for `refresh one-shot rc=0 in N.Ns` log lines staying
+  under ~10s; a `--strict-mcp-config` follow-up will skip MCP
+  startup in refresh-only invocations if that becomes a problem in
+  practice.
+
+  Tests in `test_refresh_oneshot_home_env.py`:
+  `test_refresh_oneshot_inherits_operator_home` (env-mutation guard
+  — monkeypatches `asyncio.create_subprocess_exec` and asserts
+  `env["HOME"]` equals the operator's HOME, NOT the agent's
+  home_dir); `test_refresh_oneshot_write_lands_at_host_path_visible_via_agent_symlink`
+  (end-to-end-ish — fake claude subprocess does `tmp+rename` at the
+  env's HOME path, asserts agent symlink still resolves to a file
+  with the fresh `accessToken` AND
+  `_credentials_expires_in_seconds()` reads back a positive TTL);
+  `test_refresh_oneshot_does_not_create_regular_file_at_agent_path`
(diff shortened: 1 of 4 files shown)


hellohanchen (commented):
**Looks good for fast-track merge.** Bug analysis is precise — the full causal chain (HOME override → tmp+rename at agent symlink → `link_host_credentials` copy-mode sees regular-file + mtime mismatch → stale host content overwrites fresh token) holds up under scrutiny. The fix is minimal-blast-radius (single env-dict change), the docstring update on `state.py:182-183` makes the symlink-survival invariant retroactively correct, and the test in particular `test_refresh_oneshot_write_lands_at_host_path_visible_via_agent_symlink` actually runs `os.rename` so it would have caught the pre-fix behaviour cleanly.

The `_credentials_expires_in_seconds` (L205 in this PR) calling `link_host_credentials` as a side effect — that's load-bearing for both seeding the symlink and for the test setup. Worth keeping in mind for PUF-218 if that touches read-side caching.

**One important follow-up question** I'd want answered before declaring the cascade fully closed — see inline. Two minor nits in summary.

Given the live in-band repro at 18:30 today and the urgency, ✅ — would merge.

---

## Follow-up question (potentially load-bearing)

`_ensure_session` (currently `local_cli.py:300-302`, NOT touched by this PR) still launches the long-lived session subprocess with `HOME=<agent_home_dir>` + `USERPROFILE=<agent_home_dir>`. **If** that long-lived claude session ever persists a token refresh to disk mid-session (which it likely does — Claude Code refreshes OAuth tokens internally as they near expiry, and persists for subsequent invocations), then that write would hit the agent symlink path the same way refresh-oneshot did pre-fix. Same `os.rename` → same symlink-clobbering → same `link_host_credentials` copy-mode fallback → same stale-token amplification.

PUF-217 closes the refresh-oneshot write path. PUF-218 (deferred) per your description is about the **read side** (long-lived session reloads from disk). But the **write side** of the long-lived session is currently uncovered by either.

Three ways this plays out:

1. Long-lived `--print --output-format stream-json` claude sessions don't write credentials during their lifetime (only refresh-oneshot does). In which case nothing to do. ← worth confirming with claude-code internals doc / quick strace.
2. They do write. The current PR closes the refresh-oneshot cascade but the session-write cascade can still trigger the same dogpile (probably less frequent — session refreshes only fire when token nears expiry mid-session). Worth a PUF-218.x ticket.
3. They write, but `link_host_credentials` is reliably called between the session-write and the next read, restoring the symlink. Less likely given today's repro, but possible.

Not a blocker for this PR — your urgency on the live cascade dominates. But the PUF-218 description in your PR body specifically calls out the read-side; whoever picks it up should also verify (1) or treat the session-write path as scope.

## Two minor nits in summary

1. **MCP startup cost moves to operator's config.** The PR body flags this — and correctly identifies `--strict-mcp-config` as the eventual fix. One subtle wrinkle: if the operator's MCP config is broken (one server package removed / npm cache cleared / etc), refresh-oneshot now fails with a non-zero exit and the daemon treats it as a refresh failure → adaptive cadence accelerates → dogpile re-emerges via a different cause. Defensive option is to harden `--strict-mcp-config + --mcp-config /dev/null` (or equivalent) into THIS PR, but that's scope creep on a hot-fix. Fine to defer if you're keeping a close eye on `refresh one-shot rc=` log lines post-deploy.

2. **Tests bypass `_verify()`** by flipping the private `_verified` flag (`adapter._verified = True`). Convenient but creates a coupling to a private state variable — a future refactor renaming that flag silently disables the gating. Could replace with `monkeypatch.setattr(adapter, '_verify', lambda: None)` for the same effect with less hidden coupling. Marginal — not a blocker.

hellohanchen:
Re-reviewed 43069d6 — clean fix for the private-flag nit. Tests now ride on the public `_verify` method via `monkeypatch.setattr` rather than coupling to the cache attribute name. CI green.

The session-side-writes question + MCP-startup-cost trade-off from round 1 are intentionally deferred per your PR description (PUF-218 / follow-up ticket). No new concerns.

✅ from me — ready for final review + merge as part of the 0.8.5 cut.

PR #32, opened 2026-05-19
fix(PUF-221): daemon-owned credential refresh (single writer, agents stay readers)

hellohanchen:
## Summary

- Centralizes Claude OAuth credential refresh at the **daemon** instead of per-agent — single writer to the host `~/.claude/.credentials.json` so Anthropic's single-use refresh-token rotation can no longer be raced by N agent workers all reading the same disk RT.
- New `CredentialRefresher` (portal-level) polls every 2 min for upcoming expiry (10 min safety margin), refreshes by shelling out to `claude --print "ok"` with `HOME=<host_home>`, and fans out per-agent view-sync via `link_host_credentials` so symlinks / copy-mode copies stay current.
- Replaces per-agent `agent refresh-ping <id>` CLI with a single `agent refresh-token` subcommand that drops a sentinel file the daemon's reconcile loop forwards into an in-process `asyncio.Event`, short-circuiting the 2-min poll so operator-initiated refresh runs within ~1s.
- **Retires** PUF-207 (`auth_healthy` flag + `_check_startup_auth_or_pause` startup probe) and PUF-213 (`_next_refresh_tick` adaptive cadence) — daemon-level refresh + atomic disk write makes per-agent startup probes redundant. Supersedes PUF-217 + PUF-218.
- Net **+623 / −1279** = ~656 LOC removed: `Adapter.refresh_ping` / `_run_refresh_oneshot` (both adapters) / `_credentials_expires_in_seconds` / `auth_healthy` / `_REFRESH_LOCK` / `_check_startup_auth_or_pause` / `_next_refresh_tick`, plus 4 whole test files (`test_refresh_ping.py`, `test_credential_refresh_policy.py`, `test_refresh_oneshot_home_env.py`, `test_worker_startup_auth.py`) + 2 surgical removals from `test_cli_session_recovery.py`.
- Closes PUF-221. Retires PUF-207 / PUF-213 / PUF-217 / PUF-218.

## Architecture

**Today's failure mode (the trigger):** Anthropic's OAuth uses single-use refresh tokens — every successful `/oauth/token` rotate issues a new `refresh_token` and invalidates the previous one. On cli-local Linux, long-lived `claude` subprocesses refresh **in memory only** (never writing back to disk until process exit). When the daemon's per-agent `claude --print "ok"` then fires to force a write, the in-memory rotation has already invalidated the disk RT — 401, nothing to write, every other agent reading disk gets 401 too.

**Fix:** one writer, N readers, one shared file.

- Daemon owns one `CredentialRefresher` instance. Its `run_loop` polls `~/.claude/.credentials.json` every 2 min, triggers refresh when `expiresAt - now < 10 min`, OR when `notify_refresh_needed()` is called.
- `notify_refresh_needed()` is the in-process trigger today fed by the `agent refresh-token` CLI sentinel file (`refresh_token_request_path()` — symmetric with `stop_request` pattern in `state.py`). The daemon's reconcile loop watches the sentinel and forwards into the in-process `asyncio.Event`.
- Single-writer semantics: daemon's `asyncio.Lock` ensures only one `claude --print` is in flight at a time. Anthropic's RT rotation can't be observed mid-write by another caller.
- After every successful refresh (or on every poll, as a safety net for the operator running `claude /login` externally), the refresher fans out `state.link_host_credentials(host_home, agent_home)` to each registered agent so per-agent symlinks / copy-mode copies stay in sync.

## Test plan

- [x] Unit tests on `CredentialRefresher` — 12 tests in `tests/test_credential_refresher.py`: disk-read variants, register/unregister, no-refresh-when-fresh, refresh-when-close-to-expiry, view-sync fan-out, `run_loop` stop, `run_loop` wake-on-event (<1s latency), sentinel-file round trip.
- [x] Full pytest suite: **676 passed / 1 skipped / 0 failed** post-cleanup (was 706 pre-cleanup; -30 from 4 deleted test files + 2 surgical removals).
- [x] Negative grep on removed-symbol set (`auth_healthy|refresh_ping|_run_refresh_oneshot|_check_startup_auth|_next_refresh_tick|_credentials_expires_in_seconds|_REFRESH_LOCK`) returns only 3 historical/architectural notes in docstrings + comments — zero live code references.
- [x] Sanity imports of all consumer modules (`Daemon`, `Worker`, `CredentialRefresher`, `Adapter` base, `LocalCliAdapter`, `DockerCliAdapter`) clean after the ~1100 LOC delete.
- [ ] **Live deploy validation:** run all 3 agents continuously past one access-token lifetime on operator's host; assert zero `auth_failed` transitions. (Post-merge step.)
- [ ] **Recovery validation:** manually `claude /login` mid-session, verify daemon picks up the operator-rewritten file on next poll without daemon restart. (Already empirically observed 2026-05-19 05:34–05:37 UTC when operator's `claude /login` rewrote the host file and both Solution + Equation recovered automatically on next inbound message.)


hellohanchen:
Architecturally the right move. "One writer, N readers, one shared file" is the correct mental model for Anthropic's single-use refresh-token rotation, and pulling the writer into the daemon process is the only place where `asyncio.Lock` actually gives you the guarantee you need — a per-worker mutex never could, because the in-memory RT lived inside the long-running cli-local `claude` subprocess that the daemon doesn't own.

**Local verification:**
- New `tests/test_credential_refresher.py` — **12/12 passed** in 0.76s.
- Full suite — **670 passed, 7 skipped** (the 7 are pre-existing symlink-related skips on this Windows host, unrelated to this PR).
- `python -c "import puffo_agent.portal.credential_refresh; import puffo_agent.portal.daemon; import puffo_agent.portal.cli"` — clean.

---

**Per-requirement audit:**

### Req 1: refresh runs against the HOST file, then fans out to every agent ✅

- `CredentialRefresher._refresh_now` (credential_refresh.py:178) sets `env = {**os.environ, "HOME": str(self.host_home)}` and `cwd=str(self.host_home)` before spawning `claude --print "ok"`. So the rotated tokens land in `~/.claude/.credentials.json` (host), not in any agent's per-agent home. This is exactly the PUF-217 lesson preserved (`HOME=agent_home` rewrites the symlink into a regular file — never do that).
- `_sync_views` (credential_refresh.py:233) iterates `self._agent_homes` and runs `link_host_credentials(self.host_home, agent_home)` for each.
- `Daemon._reconcile_once` registers each agent home when its worker starts (daemon.py:182, daemon.py:194) and unregisters on stop (daemon.py:209), so the fan-out set stays accurate.
- The test `test_tick_refreshes_when_close_to_expiry` (test_credential_refresher.py:113) explicitly asserts `env_home == str(tmp_path)`; `test_tick_fans_view_sync_to_registered_agents` asserts both registered agents get `link_host_credentials` called.

### Req 2: 3 triggers — host expiry, per-agent expiry, 401 ⚠️ **(401 path incomplete)**

- **Host expiry** ✅ — `_tick` (credential_refresh.py:140) reads `expires_in_seconds()` from disk and triggers when `expires_in <= REFRESH_SAFETY_MARGIN_SECONDS` (10 min). Poll every 2 min, so worst-case reaction time = 2 min + the small claude --print latency.
- **Per-agent expiry** ✅ functionally — per-agent paths are symlinks or copies of the same host file, so "per-agent expiry" *is* "host expiry" once you collapse the design. Symlink-mode agents see the refresh atomically; copy-mode (Windows) agents get the `_sync_views()` re-copy every tick. This is the intentional simplification of the architecture.
- **401 from a worker** ❌ — **this is the gap.** `notify_refresh_needed()` exists at credential_refresh.py:82 and the docstring says *"In-process trigger from an agent that just saw a 401"*, but the only production caller is the daemon's reconcile loop reading the CLI sentinel file (daemon.py:98-101). The worker DOES detect auth failures — `_AUTH_ERROR_PATTERNS` + `_looks_like_auth_error` at worker.py:326-358 — and it flips `runtime.health="auth_failed"` (worker.py:434), but it never calls `self.refresher.notify_refresh_needed()`. It also can't reach `self.refresher` today; the daemon owns it, the worker doesn't have a handle.

  So the *effective* flow today is:
  - Worker sees 401 → `runtime.health=auth_failed` → operator reads status → operator runs `puffo-agent agent refresh-token` → sentinel file → daemon → refresher fires.
  
  That's *operator-mediated*, not automatic. To meet your stated requirement #2, the worker needs an in-process callback into the refresher. Cheapest version: pass a `notify_refresh_needed: Callable[[], None]` down through `Worker.__init__` (the daemon constructs it as `self.refresher.notify_refresh_needed`), and call it from `_handle_suppressed_reply` (worker.py:421) right after the `is_auth = _looks_like_auth_error(reply)` branch sets `runtime.health="auth_failed"` at line 434. Test `test_run_loop_wakes_early_on_refresh_request` (line 213-252) already verifies the receiving end works end-to-end — it just needs a real producer.

  **Note:** the test at line 213 explicitly simulates this by calling `r.notify_refresh_needed()` from outside, comment says *"When an agent reports a 401 via notify_refresh_needed"* — i.e. the design intent matches the requirement, only the wiring is missing.

### Req 3: sync to all agents regardless of refresh result, as long as host is valid ✅

- `_tick` at credential_refresh.py:153 calls `self._sync_views()` unconditionally — runs whether `_refresh_now` succeeded, failed, returned-early (already fresh), or was skipped (no expiry trigger). This is the load-bearing line for requirement 3.
- The fan-out also runs on the `expires_in is None and not triggered_by_agent` branch (line 145), so even if the host file is corrupt mid-tick, the loop still attempts sync (the per-agent `link_host_credentials` will fail with a logged warning, not crash the daemon — covered by the `try/except` at credential_refresh.py:239-245).
- Subtle edge case: `test_tick_no_refresh_when_fresh_and_no_agent_trigger` (line 91) verifies the subprocess isn't spawned in the fresh-token path, but doesn't explicitly assert `link_host_credentials` IS called. Worth adding `monkeypatch.setattr(credential_refresh, "link_host_credentials", fake_link)` + `r.register_agent(a)` + `assert a in synced` to that test so the contract is locked down by an explicit assertion. Today it's only verified by the separate `test_tick_fans_view_sync_to_registered_agents` test.

---

**CI: CodeQL alert at `cli.py:643` is a false positive — safe to dismiss.**

The flagged line is `print(f"  {_summarise_credentials(host_creds)}")` inside `cmd_agent_refresh_token`. `_summarise_credentials` (cli.py:595-619) returns a string containing only:
- file size + mtime
- `expiresAt=<timestamp> (+/-N seconds from now)`
- `accessToken=yes/no` (presence boolean — not the token value)
- `refreshToken=yes/no` (presence boolean — not the token value)
- `scopes=['user:inference', ...]` (OAuth scope names, not secrets)

CodeQL's `py/clear-text-logging-sensitive-data` is a keyword-based heuristic — the literal substrings `accessToken=` and `refreshToken=` in the f-string trip it. Same class of FP as the SHA256-fingerprint case in earlier PRs. Dismiss as "Won't Fix → False Positive" with a one-line comment that only presence booleans are printed, no secret material crosses the print boundary.

---

**Non-blocking cleanup observations:**

1. **Dead imports / dead constants** left over after `_run_refresh_oneshot` deletion:
   - `local_cli.py:37` — `looks_like_auth_failure` is imported but no longer referenced anywhere in that file (the only call site was inside the deleted `_run_refresh_oneshot`).
   - `docker_cli.py:51` — same dead import.
   - `local_cli.py:44` — `REFRESH_ONESHOT_TIMEOUT_SECONDS = 120` is unreachable.
   - `docker_cli.py:85` — same dead constant.
   
   The constant is now defined in `credential_refresh.py:60` where it belongs. Drop the four leftover declarations + imports in a small follow-up. `looks_like_auth_failure` itself in `base.py:122` is similarly unused after this PR — *all* its call sites lived inside the deleted `_run_refresh_oneshot` paths. Could potentially be deleted too, though it's a small/cheap utility that future code may want back.

2. **PR description and the 401 path.** The architecture summary says "OR when `notify_refresh_needed()` is called" — strictly true at the API level, but the PR doesn't wire any production caller besides the CLI sentinel. Worth either (a) actually wiring the worker→refresher 401 callback as part of this PR (per Req 2 audit above) — cheap, ~10 lines — or (b) explicitly carving the 401 path out of scope in the PR description and filing the follow-up so reviewers don't pattern-match on the docstring promise. (a) is the better choice — it's small and it's literally what requirement 2 asks for.

3. **Subprocess cwd choice.** `_refresh_now` uses `cwd=str(self.host_home)` (credential_refresh.py:192). This works but is unusual — by convention claude --print's cwd is typically a workspace dir. Probably fine since `--max-turns 1` + `"ok"` prompt does no file I/O, but if claude ever starts writing transcripts to the cwd this would land them in `~/`. Worth a brief comment on the choice or — defensively — a per-daemon scratch dir like `~/.puffo-agent/refresher-cwd/`.

4. **The "exit=0 but expiresAt didn't advance" branch** (credential_refresh.py:219-227) is the right defensive check, but it only logs and returns — it doesn't escalate to anything visible to the operator. On affected claude builds this means silent refresh failure for hours until the operator notices their bot is dead. Consider bumping to `logger.error` and/or writing a daemon-wide `runtime.health=refresh_broken` flag that the next `puffo-agent status` surfaces, so the operator gets a fast-fail signal instead of waiting for the inevitable 401-cascade.

---

**Verdict:** ✅ for the core architecture and req 1 + req 3.

⚠️ **Change requested for req 2**: the 401-from-worker auto-trigger path isn't wired. The `notify_refresh_needed` plumbing exists and is tested end-to-end, but no production code calls it on the worker's 401 detection. Adding the callback through `Worker.__init__` and firing it from `_handle_suppressed_reply` is ~10 lines and closes the third requirement cleanly. Without it, "401 → trigger refresh" is operator-mediated, which is what the previous architecture already provided.

Once req 2 is wired + the small cleanup items (dead imports/constants) are addressed, this is mergeable. The CodeQL fail is dismissable.

hellohanchen:
Round-2 review — all three round-1 items addressed cleanly.

**Local verification:**
- Targeted tests (`test_credential_refresher.py` + `test_worker_error_suppress.py`) — **61/61 passed** in 1.29s.
- Full suite — **673 passed, 7 skipped** (+3 vs. round-1, all 3 new tests in `test_worker_error_suppress.py` are the on_auth_failure callback contract).
- CI: CodeQL **pass**, Analyze actions **pass**, Analyze python **pass**.

---

### Req-2 (401-from-worker auto-trigger) — closed ✅

The wire matches what I sketched in round-1, with one improvement I didn't suggest but is the right call:

- `Worker.__init__` takes `notify_refresh_needed: Optional[Callable[[], None]]` as a **keyword-only** arg (worker.py:482). Kwarg-only is correct here — it's a daemon-injected dependency, not part of the agent's identity.
- `Daemon._reconcile_once` threads `self.refresher.notify_refresh_needed` in at **both** Worker construction sites — fresh start (daemon.py:180) AND config-change restart (daemon.py:196). Both sites cover; nothing leaks through a restart.
- `_handle_suppressed_reply` grows `on_auth_failure: Optional[Callable[[], None]]` (worker.py:408), fires it **only on the auth-class branch** after `runtime.health = "auth_failed"`. Symmetric — a non-auth leak (429 / quota / 5xx) does NOT wake the refresher, which is correct because credential rotation isn't the fix for an Anthropic outage.
- **Callback exceptions are swallowed** (worker.py:440-446) — a broken hook can't break the suppression flow. This is the bit I didn't think to ask for and it's load-bearing: if `notify_refresh_needed` somehow raised, the worker would still flip health + return the suppression signal. Good defensive ordering — `runtime.health = "auth_failed"` happens BEFORE the callback call, so even a guaranteed-throw callback can't leave health in a torn state.
- **Both leak-suppression call sites wired** — fallback (worker.py:822) and api-error-retry (worker.py:864). The api-error-retry one is the hot path (FB-88 / FB-159 case-studies note in the comment), so missing it would have been a real bug.

Three new tests (`test_worker_error_suppress.py`):
- `test_handle_suppressed_reply_fires_notify_on_auth_class` — auth-class leak fires the callback exactly once + `runtime.health == "auth_failed"`.
- `test_handle_suppressed_reply_skips_notify_on_non_auth_leak` — rate-limit leak suppresses BUT does not fire callback AND does not flip health to auth_failed. Locks the symmetry contract.
- `test_handle_suppressed_reply_swallows_notify_exception` — raising callback → health still flips, suppression still returns True. Locks the defensive swallow.

This is the right test trio: positive, negative, defensive.

### Other round-1 items also folded ✅

- **Dead imports / dead constants** — `_AUTH_FAILURE_SIGNATURES` + `looks_like_auth_failure` removed from `base.py` (no callers post-rip). Dead imports + `REFRESH_ONESHOT_TIMEOUT_SECONDS` removed from both `local_cli.py` and `docker_cli.py`. Constant now lives only in `credential_refresh.py:60` where it's actually used. Clean.
- **Module docstring** — `credential_refresh.py:12-22` updated in `54970bc` to list **both** callers of `notify_refresh_needed()`: (a) CLI sentinel, (b) per-worker `on_auth_failure` hook. Future readers won't have to grep to find the second caller. Small but the right kind of small.
- **cwd=host_home rationale** — explicit comment at `credential_refresh.py:185-189` justifying the choice (matches operator's `claude /login` working dir, not whatever cwd `puffo-agent start` was launched from). Closes round-1 #3.
- **Silent-refresh-failure escalation** — `logger.warning` → `logger.error` at `credential_refresh.py:223` for the "exit=0 but expiresAt didn't advance" branch. That's the operator-facing fast-fail signal I asked for. Could go further (`runtime.health=refresh_broken` daemon-wide flag) but bumping the log level is the minimum-viable fix and stops the silent-death class of bug.
- **View-sync assertion in fresh-token path** — `test_tick_no_refresh_when_fresh_and_no_agent_trigger` now explicitly registers an agent + monkeypatches `link_host_credentials` + asserts `sync_calls == [(tmp_path, agent_a)]`. Req-3's contract is now pinned by an explicit assertion in the no-refresh path, not just inferred from the separate fan-out test.

### CodeQL

Passing now. The previous "1 high" at `cli.py:643` (presence-boolean false positive) appears to have been dismissed / re-categorized — the CodeQL job is fully green without touching `cli.py`. No action needed.

---

### Nothing left blocking

All three of your requirements are now wired in code AND covered by tests:

| Req | Path | Test |
|---|---|---|
| 1: refresh host, fan-out to agents | `_refresh_now` HOME=host_home + `_sync_views` | `test_tick_refreshes_when_close_to_expiry` + `test_tick_fans_view_sync_to_registered_agents` |
| 2a: host expiry | 2-min poll, 10-min margin | `test_tick_refreshes_when_close_to_expiry` |
| 2b: per-agent expiry | collapsed into 2a via symlink/copy + `_sync_views` every tick | covered by 1 + 2a |
| 2c: 401 from worker | `_handle_suppressed_reply` → `on_auth_failure` → `notify_refresh_needed` → `_refresh_request.set()` | `test_handle_suppressed_reply_fires_notify_on_auth_class` + `test_run_loop_wakes_early_on_refresh_request` (end-to-end) |
| 3: sync regardless of refresh result | unconditional `self._sync_views()` at `_tick:153` | `test_tick_no_refresh_when_fresh_and_no_agent_trigger` (now with explicit `link_host_credentials` assertion) |

✅ from me — ship it.

PR #33, opened 2026-05-19
fix(PUF-227): pin per-message channel context in batch render

hellohanchen:
## Summary

- Fixes per-worker prompt-assembly metadata corruption where a thread queue's batch could render one envelope's `channel_id` / `channel_name` / `space_id` / `space_name` using a **stale batch-level cached `channel_meta`** captured at first enqueue.
- Root cause (`agent/core.py:127-146` pre-fix): `handle_message_batch` read every msg's channel/space fields from the batch-level `channel_meta` dict, ignoring per-msg fields already present in each envelope's decoded dict. When the queue's append-into-existing-batch path landed a second message on the same `root_id` from a DIFFERENT channel, the second envelope's prompt block carried the FIRST message's channel context.
- Three-step fix in `agent/core.py`: `handle_message_batch` + `handle_api_error_retry` now read every channel/space field off each `msg` dict directly. The post-batch route + assistant-log channel tag uses the LAST batch msg's channel_name (matches what the agent is most likely replying about).
- `agent/puffo_core_client.py:_ThreadEntry` docstring reconciliation: explicitly states `channel_meta` is a batch-level summary, NOT load-bearing for per-message rendering. Line 837 (`entry.channel_meta = channel_meta` on reopen-after-dispatch) stays — direction-of-corruption analysis showed it's not the load-bearing path; the old docstring just incorrectly claimed channel_meta was invariant for a root.
- Closes PUF-227.

## Architecture / decision notes

**Why per-msg fields are authoritative.** Each msg dict already carries `channel_id` / `channel_name` / `space_id` / `space_name` from the envelope decode (`puffo_core_client.py:722-728` builds them per-envelope). The pre-fix code was effectively a downgrade — choosing to read from a cached summary instead of the canonical per-envelope source.

**Why daemon-side render-from-msg over upstream "fix same-root-spanning-two-channels."** Daemon-side render-from-msg makes the agent robust regardless of how the root_id collision arises (server bug, UI bug, agent-reply-system, anything else upstream). Equation flagged the upstream investigation as a possible follow-up audit ticket if customers continue reporting cross-channel pollution after this lands.

## Test plan

- [x] **Load-bearing assertion** (`tests/test_core_handle_batch.py:test_batch_renders_each_msg_with_own_channel_id`) — reproduces Scout's exact symptom shape: batch with msg-A from `#General` and msg-B from `#gtm`, stale batch-level `channel_meta` pointing at `#General`. Asserts block-B carries `channel: gtm` + `channel_id: ch_gtm`, **and critically** that `channel_id: ch_general` does NOT leak into block-B. Pre-fix this assertion fails — that's what makes it load-bearing.
- [x] **Symmetric retry path** (`test_api_error_retry_fallback_renders_each_msg_with_own_channel`) — the rate-limit retry fallback chunks each go through `_format_user_block`; assert both channel ids appear in the assembled fallback text.
- [x] **Queue-layer source of truth** (`tests/test_thread_queue.py:test_admit_same_root_two_channels_preserves_per_msg_fields`) — the canonical untested case the intake flagged. Direct precondition for the prompt-layer fix to work.
- [x] **Docstring contract** (`tests/test_thread_queue.py:test_admit_same_root_reopen_resets_channel_meta_to_new_msg`) — pins behavior so future refactors can't accidentally invariant-make `channel_meta`.
- [x] **Full pytest suite**: **685 passed / 1 skipped / 0 failed** (was 679 on main → +6 net new tests; 1 skip is the pre-existing permission-proxy-mode entry, unrelated).
- [x] **Negative grep:** zero `channel_meta.get(` reads in `handle_message_batch` or `handle_api_error_retry` post-fix.
- [ ] **Live verification post-merge:** confirm Scout no longer corrupts cross-channel mentions; spot-check from the puffotest cluster.

### DIFF
--- a/src/puffo_agent/agent/core.py
+++ b/src/puffo_agent/agent/core.py
@@ -127,12 +127,22 @@ class PuffoAgent:
         if not batch:
             return None
         for msg in batch:
+            # PUF-227: render each message with ITS OWN channel/space
+            # context, not the batch-level cached channel_meta. The
+            # admit-into-existing-batch path at puffo_core_client.py
+            # appends new messages without updating entry.channel_meta;
+            # if a thread root receives messages from two different
+            # channels in the same in-queue batch, channel_meta would
+            # carry the FIRST message's context for every subsequent
+            # render. Each msg dict already carries channel_id /
+            # channel_name / space_id / space_name from the envelope
+            # decode — use them.
             self._append_user(
-                channel_meta.get("channel_name", ""),
+                msg.get("channel_name", ""),
                 msg.get("sender_slug", ""),
                 msg.get("sender_email", ""),
                 msg.get("text", ""),
-                channel_id=channel_meta.get("channel_id", ""),
+                channel_id=msg.get("channel_id", ""),
                 root_id=root_id,
                 attachments=msg.get("attachments") or [],
                 sender_is_bot=msg.get("sender_is_bot", False),
@@ -140,8 +150,8 @@ class PuffoAgent:
                 post_id=msg.get("envelope_id", ""),
                 create_at=msg.get("sent_at", 0),
                 followups=None,
-                space_id=channel_meta.get("space_id", ""),
-                space_name=channel_meta.get("space_name", ""),
+                space_id=msg.get("space_id", ""),
+                space_name=msg.get("space_name", ""),
                 sender_display_name=msg.get("sender_display_name", ""),
                 is_visible_to_human=msg.get("is_visible_to_human", True),
             )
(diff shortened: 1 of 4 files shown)


hellohanchen:
Closing per operator decision (msg_8f26dbeb in PUF-227 thread). The render-layer-only fix is being superseded by the stricter wipe-to-null-on-cross-channel-invalid approach (Q1=(a)) split into PUF-227-A (agent) + PUF-227-B (web app). PR #33's tests + per-msg-render insight remain useful reference for the PUF-227-A receiver-side admit path.

PR #34, opened 2026-05-20
fix(PUF-227-A): strict cache+channel validation on send and admit

hellohanchen:
## Summary

- Enforces the operator-locked strict-cache invariant (PUF-227 Q1=(a)) on both ends of the agent's wire: a `thread_root_id` / `reply_to_id` is valid **only if** the parent envelope is in the local message store AND its channel/space match the outbound (sender) or incoming (receiver) envelope. Mismatch → wipe to `None` (top-level), not preserve-and-render-wrong.
- **Sender** (`mcp/puffo_core_tools.py`): new `_validate_root_same_channel` runs after `_resolve_root_id` in both `send_message` and `send_message_with_attachments`. On wipe, the tool response carries a human-readable warning so the agent can self-correct on the next compose.
- **Receiver** (`agent/puffo_core_client.py`): new `_validate_incoming_parent_id` runs in `handle_envelope` BEFORE `store.store(...)`. Wiped values propagate through downstream consumers (db row, admit-batch `root_id` routing, `msg_dict.root_id` to the agent prompt, invite-DM inline-handler intercept).
- This is the structural fix for the upstream cause that PUF-227 PR #33 was a render-layer band-aid for. Replaces (closes) #33's approach. Closes PUF-227-A.

## Architecture / decision notes

**Why both sender and receiver.** Each side enforces the invariant against its own local view. Either side alone catches the bug; together they're defense-in-depth — a sender on a stale cert can still poison a receiver that has the parent locally (and vice versa) if either skips the check. Independent ship cadence with PUF-227-B (web app) since each repo's enforcement is self-sufficient.

**Why wipe-to-`None` (not reject/return-error).** Operator's Q1=(a) framing: client should only ever see parent-ids in its local cache. An invalid id from upstream is corrupted data, not a fatal error — the right repair is to demote it to a top-level message (which is structurally always valid) rather than drop the message entirely. Receiver storage is the source of truth.

**Why same-channel AND same-space check.** Channel-mismatch is the load-bearing Scout-class symptom. Space-mismatch is belt-and-braces: a cross-space thread linkage is also nonsense and worth catching at the same gate.

**DM context handling.** Incoming DMs have no `channel_id`; receiver helper short-circuits the channel check but keeps the cache-presence check. Sender's `_validate_root_same_channel` is called with `expected_channel_id=None` for DMs — same short-circuit. Tests cover both paths.

**Transport-error policy.** Sender lookup transport error → wipe + warning (safe-fail). Receiver lookup exception → wipe (also safe-fail, log at warning). The strict invariant treats "I can't verify" identically to "I verified and it's invalid" — both routes ship a top-level message rather than risk admitting a bad thread linkage.

## Test plan

- [x] **Sender (7 new in `test_puffo_core_tools.py`):** pass-through (no root, same-channel parent), wipes (cross-channel / not-in-cache / cross-space / transport-error), DM envelope.
- [x] **Receiver (8 new in `test_puffo_core_client_admit_validate.py`):** symmetric matrix.
- [x] **Load-bearing assertion** (`test_validate_wipes_when_parent_in_different_channel`): reproduces Scout's symptom on the receiver side. Pre-fix the parent's channel context would leak into the new envelope's batch; post-fix the wiped id forces a fresh per-envelope `root_id` so admit-batch routing can't contaminate.
- [x] **Updated PUF-200 test** (`test_send_message_unknown_root_id_wiped_to_null_with_warning`): the strict-(a) behavior change is the load-bearing diff vs PUF-200's older fall-through semantics. Renamed to make the change explicit.
- [x] **Full pytest suite:** **693 passed / 1 skipped / 0 failed** (was 685 → +8 net new tests).

## Related

- **PUF-227 (parent)** — the cross-channel-same-root_id mystery and original render-layer fix attempt; PUF-227-A is the structural correction in the agent repo.
- **PUF-227-B (web app)** — symmetric strict-cache enforcement in the web client; independent ship cadence.
- **PUF-227 PR #33** — closed without merge; superseded by this PR.
- **PUF-200** — original `_resolve_root_id` ancestor walk; its test was updated to reflect strict-(a) wipe behavior.

## Known limitation

**Out-of-order delivery → permanent wipe (acceptable per operator's review).** If a WS reconnect-replay (or upstream server reordering) pushes a child envelope before its parent envelope lands in local cache, the child's `thread_root_id` / `reply_to_id` will be wiped to `None` at admit time (cache-presence check fails). Once the parent later arrives, the child does NOT get re-linked retroactively — there's no backfill mechanism. The child stays a top-level envelope in local state forever.

Acceptable trade-off given:
- WS reconnect-replay is rare on the puffo-core stream (server replays in arrival order).
- The structural property is "agents never honor a thread linkage they can't verify locally" — preserved.
- Loss is graceful (child becomes top-level, still visible + replyable) rather than catastrophic.

If this surfaces as a customer-visible issue, a follow-up could add a deferred-resolution queue (hold ids until either timeout or parent-arrival) — not in scope here.


hellohanchen:
## Review notes

**Overall: solid, ship-ready modulo the items below.** Design is right (defense-in-depth sender + receiver, strict cache-(a) wipe-to-None), validation runs before `store.store(...)`, all 4 downstream sites (`store row`, admit-batch `root_id`, prompt `msg_dict.root_id`, invite-handler intercept) pick up the wiped value via `payload_thread_root_id` rebind. Tests reproduce Scout's symptom directly.

### Suggested changes before merge

**1. (medium) No `reply_to_id`-specific test.** The helper handles `reply_to_id` (both `_validate_incoming_parent_id` and `_validate_root_same_channel` are generic), and both callsites invoke it twice — once for `thread_root_id`, once for `reply_to_id` — but every test in `test_puffo_core_client_admit_validate.py` exercises only the `thread_root_id` path. Recommend adding at minimum a `test_validate_reply_to_id_uses_same_helper` so a future refactor that re-introduces `payload.reply_to_id` somewhere is caught by CI.

**2. (low-medium) Sender wipe has no `logger.info`.** `_validate_root_same_channel` only emits a warning into the tool response. The receiver helper logs at `logger.info` on every wipe so an operator can grep daemon log for "how often is this happening". Sender should be symmetric — add `logger.info("validate_root_same_channel: wiped %s — ...", resolved_root)` in each wipe branch. (Transport-error branch already has `logger.warning`, which is correct.)

**3. (low) No CHANGELOG entry.** 0.9.0 just shipped to PyPI. Recommend adding a `## [0.9.1] — TBD` entry (or `## [Unreleased]`) for the PUF-227-A fix. Keeps the version-bump-on-next-release trivial.

### Out-of-scope / worth documenting

**4. (low) Out-of-order delivery → permanent wipe.** If a WS reconnect-replay (or server bug) pushes a child envelope before its parent, the child gets wiped to top-level and there's no backfill mechanism to re-link it once the parent arrives. The receiver test docstring already calls out "forward-traffic enforcement, no migration of historical state" — recommend lifting that into the PR description as an explicit known-limitation note, since the loss is permanent. Probably acceptable given how rare this should be in practice, but worth being explicit about.

**5. (low, nit) Helper placement asymmetry.** Receiver is an instance method (`self.store`); sender is a module-level free function (`data_client` param). The logic is near-identical (cache-presence + same-channel + same-space + exception → wipe). Could DRY into a shared `validate_parent_id(parent_id, expected_channel, expected_space, lookup_fn) -> tuple[Optional[str], str]` utility. Not load-bearing — fine to defer.

**6. (very low) Transport-error wipe is invisible at the metric layer.** Already logged at WARNING in both helpers, so visible in daemon log. But if a sqlite flake spike happens (e.g., disk pressure), there's no counter to make "wipe rate spiked" easy to chart. Could be a follow-up if we ever instrument the agent with metrics.

### Compatibility note

`_resolve_root_id`'s ancestor walk (from PUF-200) still runs first; its result then feeds `_validate_root_same_channel`. If the walk finds a true root that's in local cache + same channel, it passes through — same-channel invariant preserved. If the walk falls through with the original id, the new strict-(a) check now wipes it (the explicit intent). PUF-200's permissive semantics are correctly overridden, and the test rename (`test_send_message_unknown_root_id_falls_through_with_warning` → `_wiped_to_null_with_warning`) makes this explicit.

CI is green (Analyze actions / python / CodeQL all pass).

PR #39, opened 2026-05-22
feat(PUF-238): fill /v1/agents/{id}/log to read audit.log with tail + since

hellohanchen:
## Summary

Fills the pre-existing `GET /v1/agents/{id}/log` stub that was returning `{lines: []}` since the route was scaffolded.

- Reads NDJSON `~/.puffo-agent/agents/<agent_id>/workspace/.puffo-agent/audit.log` (populated by `local_cli.py` / `docker_cli.py`).
- Query params: `tail=N` (default 200, capped at `_LOG_MAX_TAIL=2000`); `since=<byte-offset>` for delta polling.
- Malformed JSON lines surface as `{event: "_raw", msg: <raw text>}` so the operator can still see what the agent emitted instead of silently dropping.
- Missing audit.log → empty `lines` + `note` so the client can distinguish "agent hasn't written yet" from "agent wrote then we caught up via delta."
- **Cursor past EOF resets to 0** — rotation-safe (file shrunk below `since` means the audit.log was rotated/archived; client gets a fresh initial tail on next poll).
- Auth middleware (paired identity) was already in place — non-paired callers get 401.
- Server-first ship order per intake. Client back-compat: old client polling against new endpoint works; new client polling against old daemon also works (stub returned `{lines: [], note}` previously).
- Closes server half of PUF-238.

## Test plan

- [x] **`tests/test_log_endpoint.py` (11 new):** missing-file, tail (basic), tail-default (200), tail capped at `_LOG_MAX_TAIL`, invalid `tail` (non-numeric / negative) → fallback to default, `since` byte-offset delta, `since` at EOF → empty delta, `since` past EOF → reset to 0 (rotation safe), malformed JSON line → `_raw` event wrapper, unknown agent_id → 404, unpaired caller → 401.
- [x] **Full pytest:** **819 passed / 1 skipped / 0 failed** (was 808 → +11 new).

## Related

- **PUF-238** — agent logs tab feature.
- **Client side**: `puffo-core-han-group#<TBD>` (PR opening in parallel).

### DIFF
--- a/src/puffo_agent/portal/api/handlers.py
+++ b/src/puffo_agent/portal/api/handlers.py
@@ -10,8 +10,10 @@ are allowed.
 
 from __future__ import annotations
 
+import json
 import logging
 import os
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 
@@ -1000,6 +1002,17 @@ _LOG_DEFAULT_TAIL = 200
 # Hard cap on ``tail`` — protects against accidental huge fetches
 # from a chatty agent. 2000 lines * ~300 bytes / line ≈ 600 KB.
 _LOG_MAX_TAIL = 2000
+# Cap on bytes returned in delta mode. Prevents a busy agent + slow
+# poller combination from delivering a multi-MB response in one
+# tick. When the delta exceeds this, we serve a partial window +
+# advance ``next_cursor`` to the partial offset so the client picks
+# up the rest on the next poll.
+_LOG_MAX_DELTA_BYTES = 256 * 1024
+# Reverse-seek chunk size — read this many bytes per pass off the
+# end of the file when building a tail response. Keeps memory
+# bounded for unbounded ``audit.log`` files until rotation lands
+# at the writer side.
+_LOG_TAIL_CHUNK_BYTES = 64 * 1024
 # Marker event used when a line in audit.log can't be parsed as
 # JSON. Surfaces the raw text so the operator can still see what
 # the agent emitted instead of dropping the line silently.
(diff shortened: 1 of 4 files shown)


src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
+        next_cursor = offset + len(content)
+        for raw in content.decode("utf-8", errors="replace").splitlines():
+            if raw.strip():
+                lines.append(_parse_log_line(raw))
+    else:
+        # Tail mode — read the whole file (the MAX_TAIL bound on
+        # the response makes the memory hit predictable) and slice
+        # the last N lines. NDJSON files in practice stay small
+        # (rotation lives at the daemon layer, not here).
+        decoded = log_path.read_bytes().decode("utf-8", errors="replace")
```
hellohanchen: 🟡 Tail mode reads the whole file into memory.

The inline comment says rotation lives at the daemon layer, but I checked `cli_session.AuditLog.write` — it just `open("a")`s, no rotation anywhere. So `audit.log` grows unbounded for the life of an agent. A chatty agent at 100 events/turn × ~300 B/line over a few months hits 100s of MB; a single `tail=200` request then allocates that twice (`read_bytes()` + `decode()`) plus a list of N strings.

Reverse-seek pattern is the usual fix — read backwards in chunks until you have `tail+1` newlines, then slice. Or enforce rotation at the writer side and update this comment. Either way the current comment overstates the safety.

src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
+        # that's shrunk (rotated / archived) below the cursor
+        # resets to 0 so the next response carries the full window.
+        offset = since if since <= file_size else 0
+        with log_path.open("rb") as f:
+            f.seek(offset)
+            content = f.read()
+        next_cursor = offset + len(content)
+        for raw in content.decode("utf-8", errors="replace").splitlines():
+            if raw.strip():
+                lines.append(_parse_log_line(raw))
```
hellohanchen, line 1173: 🟡 Delta mode has no upper bound on bytes returned.

`tail` is enforced for the no-`since` path but not here — if the agent wrote 20 MB between polls, the response carries all 20 MB. Worth capping (e.g., `content[:MAX_DELTA_BYTES]`) and advancing `next_cursor` to the partial offset so the client picks up the rest next poll. Otherwise slow poller + busy agent → daemon memory thrash.

src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
-        "note": (
-            "per-agent log capture is not implemented yet — daemon "
-            "currently writes a single combined stream. follow-up PR."
-        ),
-    })
+        "lines": lines,
+        "next_cursor": next_cursor,
+    }
+    if not lines:
+        body["note"] = "no log entries yet"
```
hellohanchen: 🟡 PR description says `note` distinguishes "agent hasn't written yet" from "caught up via delta", but the string is identical (`"no log entries yet"`) in both branches — line 1063 (missing file) and line 1098 (empty result).

The distinction is only in `next_cursor` (0 vs file_size), which is brittle for a UI to switch on. Suggest two distinct notes (`"audit log not yet created"` vs `"no new entries since cursor"`) or a separate `state` field.

src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
+            "agent_id": agent_id,
+            "lines": [],
+            "next_cursor": 0,
+            "note": "no log entries yet",
+        })
+
+    file_size = log_path.stat().st_size
+
+    lines: list[dict[str, Any]] = []
+    if since is not None:
```
hellohanchen, line 1149: 🟢 When both `tail` and `since` are passed, `tail` is silently ignored. No test for the combo. Either ignore intentionally + add a sentence saying so, or 400 on the conflict — silent ignore is a debug-time gotcha.

src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
+
+def _parse_log_line(raw: str) -> dict[str, Any]:
+    """Decode one NDJSON line from audit.log. Falls back to a
+    ``_raw`` event wrapper if the line isn't valid JSON so the
+    client can still display the bytes."""
+    import json
+    try:
+        return json.loads(raw)
+    except (json.JSONDecodeError, ValueError):
+        return {"ts": "", "event": _LOG_MALFORMED_EVENT, "msg": raw[:1024]}
```
hellohanchen: 🟢 Malformed line gets `ts=""`. If a future client sorts by `ts` (common in log UIs), empty strings collate to the top so all `_raw` events bunch regardless of when they were written.

Minor — using `datetime.now(timezone.utc).isoformat(...)` for the ingestion time would keep ordering sane. The raw bytes are already preserved in `msg`, so a synthesized `ts` loses nothing.

src/puffo_agent/portal/api/handlers.py
```diff
@@ -993,21 +993,110 @@ async def archive_agent(request: web.Request) -> web.Response:
...
+# JSON. Surfaces the raw text so the operator can still see what
+# the agent emitted instead of dropping the line silently.
+_LOG_MALFORMED_EVENT = "_raw"
+
+
+def _parse_log_line(raw: str) -> dict[str, Any]:
+    """Decode one NDJSON line from audit.log. Falls back to a
+    ``_raw`` event wrapper if the line isn't valid JSON so the
+    client can still display the bytes."""
+    import json
```
hellohanchen: 🟢 `import json` inside the helper is fine functionally (Python caches modules), just unusual style — `json` is already imported at module top elsewhere in this file (line 1491). Lifting this one up keeps imports in one place.

hellohanchen (commented):
Solid PR. 10 tests cover the major paths, rotation-safety via past-EOF reset is a nice touch, `_raw` preservation > silent drop. Comments are all 🟡 perf / 🟢 clarity — no blockers.

hellohanchen (commented):
Round 2 LGTM (would APPROVE if GitHub let me — self-PR). All 6 prior comments addressed:

- 🟡 tail-mode whole-file read → `_read_tail_bytes` reverse-seek in 64 KB chunks, covered by `test_log_tail_uses_reverse_seek_on_a_large_file`
- 🟡 unbounded delta → `_LOG_MAX_DELTA_BYTES = 256 KB` + newline-boundary trim + partial-cursor advance, covered by the new cap test and the partial-completion follow-up test
- 🟡 ambiguous `note` → distinct `state` field (`never_written` / `up_to_date` / `empty`) with matching note copy
- 🟢 `tail` + `since` combo → 400 + `"mutually exclusive"` error
- 🟢 malformed `ts=""` → synthesized ingestion time via `datetime.now(timezone.utc).isoformat(...)`
- 🟢 `import json` lifted to module top

No new issues introduced. Follow-up nit (not for this PR): `audit.log` writer side still doesn't rotate — reverse-seek makes reads safe at scale but disk usage grows unbounded. Worth filing once the read surface is in production.

Nice turnaround.

PR #45, opened 2026-05-26
fix(PUF-252): surface api-error-abandon via runtime.health so silent failures stop being invisible

hellohanchen:
## Summary

Fixes **PUF-252 bug-1** — the auto-recovery wedge Sam diagnosed at `msg_11de3d4e` (relayed via Grande). Scout went silent on a DM after a Claude rate-limit fired; the consumer-loop's kick-retry path exhausted + abandoned the batch silently, leaving `runtime.health == "ok"` while the actual state was "abandoned". Sam had to manually restart Scout.

This PR ships the **state-honesty data layer** — new `on_api_error_abandon` callback fired exactly once when kick-retries exhaust; worker callback flips `runtime.health → "api_error_abandoned"` + populates `runtime.error` with a human-readable summary. Bug-2 (UI affordance: status dot + refresh/restart lever) is **scoped out** — Nova canonical-side already split it into FB-197 + FB-198 (Operator Action Panel cluster); this PR ships the data those UI surfaces will consume.

## Discrimination outcome (Stage-3 trace)

**(α) confirmed via code-read:** Sam's reply path was `agent/core.py:_run_turn_and_route` raising `AgentAPIError` → kick-retry in `_consume_queue` → exhaust → silent return.

**(β) ruled out:** PUF-214's `_handle_suppressed_reply` was NOT on Sam's path (his reply contained literal "API Error" so the upstream raise caught it first; the error-translation path Calculation looked at didn't intercept the abandon).

## Architectural alignment

Per the just-updated `feedback_dedup_triage_policy.md` (PUF-249 revision: user-action via UI is the right recovery path, not platform auto-recovery), this PR ships **state visibility only** — no auto-restart, no auto-resume. The UI affordance (FB-198 restart lever) is the correct recovery surface; this PR gives it a signal to render.

Error summary copy explicitly cites the user-action path: *"The agent has gone silent on this thread until a new message arrives OR the agent is refreshed/restarted."*

## What changed

- **`portal/state.py`**: `RuntimeState.health` enum extended with `"api_error_abandoned"` alongside existing `"ok"` / `"auth_failed"` / `"unknown"`. Comment block documents the bug-2 read-path so a future UI dev sees where this state surfaces.
- **`agent/puffo_core_client.py`**: new `on_api_error_abandon: Optional[Callable[..., Coroutine]]` parameter on `listen()`, threaded through `_consume_queue` → `_do_api_error_retries`. Fired exactly once at three exit points:
[...]

PR #48, opened 2026-05-27
fix(PUF-258): clear runtime.health="auth_failed" on credential refresh-success

hellohanchen:
## Summary

Wires the EXIT edge for ``runtime.health = "auth_failed"`` — symmetric to PUF-255's ``on_turn_success`` clear-path for ``api_error_abandoned``, but at the credential-refresh-success event boundary instead of the turn-success boundary.

Pre-PUF-258 the daemon set ``runtime.health = "auth_failed"`` (worker.py:479, PUF-221's lane) but **nothing in the codebase cleared it**. Once flipped, the flag stayed sticky until process restart, which made ``audit.log`` misleading for power-users / operators checking agent health post-recovery (operator surfaced the gap via the PR #45 grep observation).

## Changes

- ``CredentialRefresher`` gets ``register_on_refresh_success(cb)`` / ``unregister_on_refresh_success(cb)`` callbacks
- ``_refresh_now`` fires after ``backend.refresh()`` succeeds (lock released first so callbacks can't deadlock the next refresh)
- ``_external_rotation_loop`` fires after detected rotation + ``_sync_views`` (covers the macOS Keychain external-rotation path)
- Failure path stays silent — no fire on ``backend.refresh()`` raising
- Daemon ``_register_with_refresher(agent_cfg, worker)`` wires Worker's ``_clear_auth_failed_if_recoverable`` staticmethod; callback identity stashed on ``worker._refresh_success_callback`` so ``_stop_worker`` can unregister symmetrically from both refreshers
- ``Worker._clear_auth_failed_if_recoverable`` lifted staticmethod (mirrors PUF-255's ``_clear_api_error_abandoned_if_recoverable``):
  - Only clears ``auth_failed``. Leaves ``api_error_abandoned`` alone (PUF-255's lane; symmetric partition)
  - No-op on ``ok`` / ``unknown`` to skip the save churn on the steady-state hot path
  - **Optimistic-clear**: if refresh succeeds but the next request still 401s, ``_handle_suppressed_reply`` re-flips on the next leak detection
- ``state.py`` docstring updated to reflect the now-symmetric ENTER/EXIT lifecycles for both ``api_error_abandoned`` (PUF-252 set / PUF-255 clear) and ``auth_failed`` (PUF-221 set / PUF-258 clear)

## Tests

15 new tests in ``tests/test_credential_refresh_clears_auth_failed.py``:

**Helper-policy (5):**
- flips ``auth_failed`` → ``ok``; no-op on ``ok``; no-op on ``unknown``; leaves ``api_error_abandoned`` alone (symmetric partition); optimistic-clear-then-re-set sequence

**CredentialRefresher wiring (6):**
- register/unregister roundtrip; multi-subscriber fan-out (order preserved); callback-exception isolation with caplog at WARNING; ``_refresh_now`` fires on success; ``_refresh_now`` does NOT fire on failure; skip-branch (sufficient TTL) doesn't fire

**External-rotation E2E (1):**
- Monkeypatches ``backend.poll_external_rotation`` + shrinks ``KEYCHAIN_POLL_INTERVAL_SECONDS`` + runs ``_external_rotation_loop`` briefly; asserts ``_sync_views`` called AND callback fired. Pins the rotated-branch ``_fire_refresh_success`` site so a macOS Keychain regression breaks the test.

**Daemon-side wiring (3):**
- ``_register_with_refresher`` binds callback to ``worker.runtime`` (closure-binding pinned)
- Codex-harness routes to ``codex_refresher`` (not ``self.refresher``)
- ``_stop_worker`` unregisters from both refreshers (cross-refresher cleanup symmetry)

**Test count:** 883 passed, 1 skipped (was 879 pre-PR; +15 across both commits; usual permission-mode skip).

## Scope & discipline

- **PR-256 close discipline preserved:** PUF-258 does NOT add a new propagation surface to puffo-server / UI. The ``runtime.health`` field stays local-honest. Per the user-action-invariance pre-check: this is owner-side cleanup of the local field PUF-256 closed as not-worth-propagating. Audit.log diagnostic value justifies the work per the "observability-for-operators IS legitimate" edge-case.
- **Symmetric partition with PUF-255:** each helper's first line guards on its own ``runtime.health`` value; pinned by ``test_clear_leaves_api_error_abandoned_alone``.
- **Ship-to-host:** daemon-side fix; agent restart picks it up. No server schema changes, no client changes.

## Out of scope

- Per-thread granularity for ``runtime.error`` (same coarseness as PUF-255; PUF-253 design lane)
- Central state-machine registry (Nova's "(1)-vs-(3) scope-choice" confirmed-out by Equation intake)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,30 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **PUF-258: ``runtime.health = "auth_failed"`` no longer sticky after
+  credential refresh-success.** The daemon set the flag in
+  ``_handle_suppressed_reply`` (PUF-221) but nothing cleared it
+  anywhere — once flipped it stayed until process restart, leaving
+  ``audit.log`` dishonest about post-recovery agent state.
+
+  ``CredentialRefresher`` now exposes
+  ``register_on_refresh_success(cb)`` /
+  ``unregister_on_refresh_success(cb)`` and fires after both regular
+  ``_refresh_now`` success AND ``_external_rotation_loop`` detected
+  rotation. Fire happens outside the refresh lock so callbacks can't
+  deadlock the next cycle; gated on
+  ``outcome is RefreshOutcome.REFRESHED`` so PUF-265's UNCHANGED
+  case (exit=0 but token didn't actually rotate) doesn't oscillate
+  ``auth_failed → ok → auth_failed``. Callbacks isolated — a
+  subscriber raising logs at WARNING and the loop continues.
+
+  ``Worker._clear_auth_failed_if_recoverable`` (lifted staticmethod
+  matching PUF-255's pattern) flips ``"auth_failed" → "ok"`` and
+  clears ``runtime.error``. Leaves ``api_error_abandoned`` to
+  PUF-255's ``on_turn_success`` lane (symmetric partition). Optimistic
+  semantics: if the agent's next request still 401s,
+  ``_handle_suppressed_reply`` re-flips on the next leak detection.
+
 - **PUF-264: request-too-large no longer infinite-retries + no longer
   leaks the raw Anthropic API error to chat.** When a user uploaded
   10 PDFs (msg_a42f8cbb), the agent inlined all of them into one
--- a/src/puffo_agent/portal/credential_refresh.py
+++ b/src/puffo_agent/portal/credential_refresh.py
@@ -628,11 +628,6 @@ class CredentialRefresher:
             self.host_credentials = backend.host_credentials
         self._refresh_request = asyncio.Event()
         self._agent_homes: set[Path] = set()
-        # PUF-258: per-Worker callbacks fired after a successful refresh
-        # (regular tick OR external-rotation detection) so each Worker
-        # can clear its ``runtime.health == "auth_failed"`` optimistically.
-        # Mirrors the ``register_agent`` shape but operates on the
-        # in-memory Worker reference instead of a disk path.
         self._on_refresh_success: list[Callable[[], None]] = []
         self._lock = asyncio.Lock()
 
@@ -643,13 +638,6 @@ class CredentialRefresher:
         self._agent_homes.discard(Path(agent_home))
 
     def register_on_refresh_success(self, callback: Callable[[], None]) -> None:
-        """PUF-258: subscribe to refresh-success notifications.
-
-        Fired after every successful refresh OR external-rotation
-        detection so subscribers (Workers) can clear an
-        ``auth_failed`` flag optimistically. Callbacks are sync; any
-        exception is caught + logged so a buggy subscriber doesn't
-        break the refresh loop."""
         self._on_refresh_success.append(callback)
 
     def unregister_on_refresh_success(self, callback: Callable[[], None]) -> None:
@@ -659,12 +647,7 @@ class CredentialRefresher:
             pass
 
     def _fire_refresh_success(self) -> None:
-        """Dispatch refresh-success to every registered subscriber.
-
-        Callback exceptions are swallowed + logged -- the refresh
-        loop has already done its real work (rotated credentials,
-        synced views); a Worker callback raising here mustn't
-        cascade into the daemon's refresh cadence."""
+        # list(...) defensive copy: callback may (un)register during dispatch.
         for cb in list(self._on_refresh_success):
             try:
                 cb()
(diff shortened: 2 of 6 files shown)


hellohanchen:
Review LGTM in shape — symmetric ENTER/EXIT pattern matches PUF-255 cleanly, multi-layer test coverage (helper policy + refresher wiring + external-rotation E2E + daemon-side wiring) is solid, scope discipline is correct (no propagation surface added, PR-256 close discipline preserved). **One mandatory integration item with PR #52 before either ships.**

## 1. Must-fix: conflict with PR #52 (PUF-265) in `_refresh_now`

Both PRs change the same lines of `_refresh_now`. The conflict isn't just textual — the semantic resolution matters.

**PR #48 current:**
```python
try:
    await self.backend.refresh()
    refreshed = True
except Exception as exc:
    logger.warning("backend refresh errored: %s", exc)
...
if refreshed:
    self._fire_refresh_success()
```

**PR #52 current:**
```python
try:
    outcome = await self.backend.refresh()
except Exception as exc:
    logger.warning("backend refresh errored: %s", exc)
    outcome = RefreshOutcome.FAILED
self._propagate_outcome(outcome)
```

Semantic problem: PR #48's `refreshed = True` triggers on "no exception", which **includes the `UNCHANGED` outcome** (the exit=0-but-expiresAt-didn't-advance silent-failure case PUF-265 catches). If `UNCHANGED` fires `_fire_refresh_success`:

1. Worker's `auth_failed` clears to `ok`
2. Next request still 401s (token wasn't actually rotated)
3. `_handle_suppressed_reply` re-flips to `auth_failed`
4. Operator sees `auth_failed → ok → auth_failed → ok` oscillation in `audit.log`

Regardless of merge order, the resolution should be:

```python
async with self._lock:
    ...
    try:
        outcome = await self.backend.refresh()
    except Exception as exc:
        logger.warning("backend refresh errored: %s", exc)
        outcome = RefreshOutcome.FAILED
    self._propagate_outcome(outcome)        # PUF-265
if outcome is RefreshOutcome.REFRESHED:
    self._fire_refresh_success()            # PUF-258 — only on REAL success
```

The `_external_rotation_loop` fire path is unaffected — that runs only when the macOS Keychain has actually rotated, so the credentials are by definition fresh there.

Suggested new test in `test_credential_refresh_clears_auth_failed.py`:

```python
@pytest.mark.asyncio
async def test_refresh_now_does_not_fire_on_unchanged(tmp_path, monkeypatch):
    """UNCHANGED outcome (PUF-265: claude exited 0 but expiresAt
    didn't advance) must NOT fire refresh-success callbacks —
    credentials didn't actually rotate, so optimistically clearing
    auth_failed would oscillate auth_failed → ok → auth_failed every
    poll cycle until the underlying refresh recovers."""
```

## 2. Verified positives

- **Lock-release-then-fire** in `_refresh_now` is the right call — prevents callback re-entry from deadlocking the next refresh
- **Callback exception isolation** (`list(self._on_refresh_success)` defensive copy + try/except per callback + WARNING log) — `test_callback_exception_does_not_break_refresh_loop` uses double-registered `good_cb` to pin both "good fires before bad" AND "loop continues past bad", which is exactly the right shape
- **Precedence partition** — `_clear_auth_failed_if_recoverable`'s first-line guard `if runtime.health != "auth_failed": return` cleanly leaves `api_error_abandoned` to PUF-255's `on_turn_success` lane, with explicit `test_clear_leaves_api_error_abandoned_alone` coverage. Symmetric to PUF-255's leaves-auth-failed-alone
- **Cross-refresher cleanup** — `_stop_worker` unregisters from BOTH `refresher` and `codex_refresher` (idempotent-on-both, same pattern as `unregister_agent`); `test_daemon_stop_worker_unregisters_callback_from_both_refreshers` defends against single-side regression
- **Codex routing** — `test_daemon_register_codex_harness_routes_to_codex_refresher` pins the `harness="codex"` branch goes to `codex_refresher` with independent callback list
- **Hot-path no-op** — `health == "ok"` skip avoids `runtime.save` churn on steady-state cycles; `test_clear_no_op_on_ok` pins
- **External-rotation E2E** — `test_external_rotation_loop_fires_on_detected_rotation` pins the second `_fire_refresh_success` site so a macOS Keychain regression breaks the test

## 3. Optional cleanups (not blocking)

- **Docstring trim**: `register_on_refresh_success` / `_fire_refresh_success` / `_clear_auth_failed_if_recoverable` carry 5-15-line WHY-rationale docstrings. Default-zero per minimal-comments preference. Defer to rebase pass.
- **`Worker._refresh_success_callback` not in `__init__`**: currently set via setattr in `_register_with_refresher`, read via `getattr(..., None)` in `_stop_worker`. Safe but informal. Initialise to `None` in `Worker.__init__` for clarity.
- **`_make_fake_worker` style**: `worker = type("FakeWorker", (), {})()` reads cleaner as `types.SimpleNamespace`. Pure cosmetic.

## Suggested action

1. **Must**: integrate the `_refresh_now` resolution above (fire only on `RefreshOutcome.REFRESHED`) + add the `does_not_fire_on_unchanged` test.
2. **Coordinate**: decide PR #48 / #52 merge order with the team. Either order is fine as long as whichever ships second carries the integrated `_refresh_now` shape.
3. **Optional**: docstring trim + `__init__` declaration during the rebase pass.

PR #50, opened 2026-05-29
fix(PUF-263): enforce paused-only on /v1/agents/export

hellohanchen:
## Summary

Operator spec (msg_9d0aaa27, item 1a): only paused agents can be exported. A running agent may be mid-write (memory updates, `cli_session.json` refresh, in-flight skill state) so the snapshot would be inconsistent and could either silently lose data or import into a broken state on the other side. The existing `/v1/agents/export` multi-agent migration endpoint already produced `.puffoagent` archives; this adds the paused-only guard.

## Changes

`src/puffo_agent/portal/api/handlers.py::agents_export`:
- Before delegating to `exp.pack`, load each requested agent's `AgentConfig` and 409 if `cfg.state != "paused"`
- **Whole-batch reject** — if any one agent in the request is running, the bundle isn't a consistent snapshot even partially, so we don't return a half-populated archive
- Reuses `AgentConfig.load` + `cfg.state` enum (`"running"` | `"paused"`) from `portal/state.py`
- New `_conflict` helper next to existing `_bad` / `_not_found` (one-liner, status=409)

## Tests

2 new + 1 helper-default change:
- `_seed_agent` now defaults to `state="paused"` (the new canonical export state); existing happy-path tests already exercise that shape and continue to pass unchanged
- `test_bridge_export_rejects_running_agent` — single agent seeded running → 409, body mentions "running" + agent id so the UI can fingerprint the cause
- `test_bridge_export_rejects_whole_batch_if_any_running` — mixed batch with one paused + one running → 409, body fingers the offending (non-paused) agent

**Test results:** 8/8 focused + 870 passed, 1 skipped (unrelated permission-mode test) full suite.

## Paired client PR

This is the server-side half of PUF-263. Web client PR (puffo-core-han-group) adds the Export button on the agent card + password modal + blob download. Ship server first (defines 409 surface); client tolerates a v1 daemon by hiding the button.

## Out of scope

- Import-side UI not in this PR (existing `/v1/agents/import` endpoint is unchanged)
- Multi-agent batch export from the UI — server supports it; just no UI for now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,13 +36,17 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   first request without an extra ``/devices/subkeys`` round-trip.
   ``_revoke_old_device`` reuses that pre-registered subkey instead
   of POSTing a fresh one, so total server traffic is unchanged
-  (old subkey for enrol + new subkey for revoke). After a successful
+  (old subkey for enrol + new subkey for revoke). After
   ``_commit_staging`` — whether revoke succeeded cleanly or got
-  shelved as ``pending_revoke.json`` — the imported ``agent.yml`` is
-  patched from ``state: paused`` (inherited from the export gate) to
-  ``state: running`` so the operator doesn't have to click Resume on
-  the new machine. State flip is best-effort: a yaml write failure
-  is logged and the agent stays paused but otherwise functional.
+  shelved as ``pending_revoke.json`` — ``AgentConfig.save`` patches
+  ``state: paused`` (inherited from the export gate) to ``running``
+  so the operator doesn't have to click Resume on the new machine.
+  Subkey registration and the state flip are both best-effort: a
+  401 on ``/devices/subkeys`` (chain-validation lag right after
+  enrol) is logged and the import proceeds without persisting a
+  session — the worker rotates one on first request, same as a
+  fresh install. A yaml write failure on the state flip leaves the
+  agent paused but otherwise functional.
 
 - **Codex agent archive/delete failing with ``Permission denied`` on
   ``.codex/tmp/.../.lock`` (Windows).** The codex CLI holds an
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,21 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   loop (which respects the paused-by-operator flag). Tighten with a
   per-agent lock if a regression surfaces.
 
+- **PUF-263: import flow now self-contained, lands the agent in
+  running state.** ``import_bundle`` registers the new device's
+  subkey immediately after enrol and persists it as a session under
+  ``keys/<slug>.session.json`` so the agent worker can sign its
+  first request without an extra ``/devices/subkeys`` round-trip.
+  ``_revoke_old_device`` reuses that pre-registered subkey instead
+  of POSTing a fresh one, so total server traffic is unchanged
+  (old subkey for enrol + new subkey for revoke). After a successful
+  ``_commit_staging`` — whether revoke succeeded cleanly or got
+  shelved as ``pending_revoke.json`` — the imported ``agent.yml`` is
+  patched from ``state: paused`` (inherited from the export gate) to
+  ``state: running`` so the operator doesn't have to click Resume on
+  the new machine. State flip is best-effort: a yaml write failure
+  is logged and the agent stays paused but otherwise functional.
+
 - **Codex agent archive/delete failing with ``Permission denied`` on
   ``.codex/tmp/.../.lock`` (Windows).** The codex CLI holds an
   exclusive file lock on ``.codex/tmp/arg0/codex-<id>/.lock`` for the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,27 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **PUF-263: ``/v1/agents/export`` enforces paused-only.** A running
+  agent may be mid-write (memory updates, ``cli_session.json``
+  refresh, in-flight skill state) so a snapshot would be inconsistent
+  and could either silently lose data or restore into a broken state
+  on the other side. ``agents_export`` now loads each requested
+  agent's ``AgentConfig`` and returns 409 (new ``_conflict`` helper)
+  when ``cfg.state != "paused"``. Whole-batch reject — a single
+  non-paused agent in a multi-agent bundle fails the request,
+  preserving "either everything in the bundle is a consistent
+  snapshot, or nothing is." Unknown agent ids return 404 with the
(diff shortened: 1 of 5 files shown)


hellohanchen:
Review LGTM in shape — paused-only is the canonical fix for snapshot consistency, whole-batch reject preserves the "consistent snapshot or nothing" invariant, and the test seed helper's new `state="paused"` default is the right place to absorb the canonical-state-change. Three things to address + a couple of minor cleanups.

## 1. Must-fix: CHANGELOG.md entry missing

The 0.9.6 unreleased section gets entries for PUF-258 / PUF-264 / PUF-265 / PUF-266 already. PUF-263 needs one too — matches the existing pattern (root cause → fix → behavior change → known limits/scope).

Suggested entry:

```markdown
- **PUF-263: `/v1/agents/export` enforces paused-only.** A running
  agent may be mid-write (memory updates, `cli_session.json` refresh,
  in-flight skill state) so a snapshot would be inconsistent and could
  either silently lose data or restore into a broken state on the
  other side. ``agents_export`` now loads each requested agent's
  ``AgentConfig`` and returns 409 (new ``_conflict`` helper) when
  ``cfg.state != "paused"``. Whole-batch reject — a single non-paused
  agent in a multi-agent bundle fails the request, preserving "either
  everything in the bundle is a consistent snapshot, or nothing is."
  Paired with the web client's Export button (gated on paused) so the
  409 only fires on a race where the agent flipped between gate and
  submit.
```

## 2. Must-fix: TOCTOU race between paused check and `exp.pack` — acknowledge it

```python
for aid in raw_ids:
    cfg = AgentConfig.load(aid)
    if cfg.state != "paused":
        return _conflict(...)
# ← agent state could flip to running HERE
blob = exp.pack(raw_ids, password, ...)
```

The window is small (microseconds) but real: the daemon's reconcile loop, another API call, or the operator clicking Resume between gate and pack would let `exp.pack` snapshot a half-running agent. The 409-gated UI doesn't protect this — daemon-side caller bypassing the UI hits it raw.

Two paths:

**(a) Accept the race + document it.** Operator's intent + small window + no auto-unpause makes the real-world risk near-zero. Add a one-line comment above the guard naming the assumption:

```python
# PUF-263: paused-only export. NOTE: there is a small TOCTOU window
# between this check and exp.pack — if the agent gets resumed mid-pack
# the snapshot is partially-inconsistent. Acceptable for P0 because
# the only resume paths are operator-driven (visible) or the reconcile
# loop (which respects the paused-by-operator flag). Tighten with a
# per-agent lock if a regression surfaces.
```

**(b) Fix with a per-agent lock.** `exp.pack` acquires the same lock the state-flip path uses. Bigger change; out of scope for this PR.

Either is fine — I'd take (a) for now. Just don't leave the race unmentioned.

## 3. Should-fix: `FileNotFoundError` → 404 branch not tested

The new code has:

```python
try:
    cfg = AgentConfig.load(aid)
except FileNotFoundError:
    return _not_found(f"agent not found: {aid}")
```

No test exercises this branch. A future refactor that drops the try/except (or that swaps `FileNotFoundError` for a different exception type) would silently regress to a 500 stack-trace at the user. Cheap to add — request export for `"does_not_exist"`, assert 404 + body fingers the id.

## 4. Minor: two loops over `raw_ids`

```python
for aid in raw_ids:
    if not is_valid_agent_id(aid):
        return _bad(f"invalid agent id: {aid!r}")
# ...
for aid in raw_ids:
    try:
        cfg = AgentConfig.load(aid)
    ...
```

Distinct concerns (cheap syntactic check before any disk read), so the two-loop structure is justifiable. Could merge into one loop with both checks inline, but the cost is a slightly less readable failure mode (a bad-id error after a partial disk read instead of cleanly-rejected upfront). Leave as-is unless you want the consolidation.

## 5. Minor: `from ..state import AgentConfig` mid-function

Inside `agents_export` rather than module-top. Probably a circular-import avoidance — if so, worth a one-line comment. If no circular risk, move to the top with other imports.

---

Test quality is good — the new seed-helper `state="paused"` default is the right place to absorb the change without churn across the existing happy-path tests. The mixed-batch test specifically asserting the offending agent (`beta`) is named, not the paused one (`alpha`), is exactly the right precision to defend against the "first agent in the list" lazy implementation. Both 409 tests check both status AND body content for the relevant id.

CI green, ship order with web PR #163 (server first) is correct per the description.

hellohanchen:
第二轮 review — 之前 3 个 must-fix 全部到位，新增覆盖也合理。LGTM，2 个 minor polish 项你看下要不要在 merge 前一起带上。

## 已解决 ✅

- **CHANGELOG entry** — 0.9.6 unreleased 加了完整段落，含 TOCTOU 已知限制 + 缓解理由。
- **`_conflict` 409 helper** — `_bad` / `_not_found` 那一组里塞进去，pattern 一致。
- **Paused-only guard** — `agents_export` 在 `exp.pack` 之前对每个 agent 跑 `AgentConfig.load` + 校验 `state == "paused"`，whole-batch reject 保留了「整个 bundle 一致或者全失败」的不变量。
- **测试覆盖** — running 单 agent 拒绝、混合 batch 拒绝、unknown id → 404 三个新 case，足够锁住合同。
- **TOCTOU 文档化** — code comment + CHANGELOG 都说明了风险 + 为什么 P0 可以接受（resume 路径要么操作员驱动要么 reconcile 尊重 paused flag）。

## Minor polish（非阻塞）

### 1. `AgentConfig.load(aid)` 只 catch `FileNotFoundError`

`handlers.py` 1727 行只处理文件不存在的情况。如果 agent dir 存在但 `agent.yml` 损坏 / 字段缺失，`AgentConfig.load` 会抛 yaml/validation 异常，这条会冒到 aiohttp default handler 当 500 给用户。

建议改成更宽：

```python
try:
    cfg = AgentConfig.load(aid)
except FileNotFoundError:
    return _not_found(f"agent not found: {aid}")
except Exception as exc:
    return _conflict(f"agent {aid} config unreadable: {exc}")
```

或者最少把 `Exception` 兜一层 500 → 4xx，避免 stack trace 漏给用户。

### 2. `raw_ids` 双重遍历

handlers.py 上面已经有一遍 `for aid in raw_ids: is_valid_agent_id(aid)` 校验，新增的 paused-check 又遍历了一遍。合一次：

```python
for aid in raw_ids:
    if not is_valid_agent_id(aid):
        return _bad(f"invalid agent id: {aid!r}")
    try:
        cfg = AgentConfig.load(aid)
    except FileNotFoundError:
        return _not_found(f"agent not found: {aid}")
    if cfg.state != "paused":
        return _conflict(f"agent {aid} is {cfg.state!r}; pause it before exporting")
```

可读性 + 一次磁盘往返。纯 style，不阻塞。

## 注意点（已 OK，FYI）

- `_seed_agent` 默认 `state="paused"` 的变更我跑了所有调用点，3 个未传 state 的 test (lines 151/247/286) 用 agent 的方式都跟 export 兼容，没静默 regression。

PR #52, opened 2026-05-29
fix(PUF-265): capture RefreshOutcome + flip runtime.health=refresh_broken on streak [P0 defensive]

hellohanchen:
## Summary (P0 defensive fix)

`CredentialRefresher._refresh_now` used to call `backend.refresh()` **without capturing the returned `RefreshOutcome`**, so `UNCHANGED` outcomes (the exact "claude exited 0 but expiresAt didn't advance" case detected at `credential_refresh.py:215-223`) fell through silently. The daemon went back to its 2-min poll loop, `runtime.health` stayed `"unknown"` fleet-wide, and operators only noticed once individual agents 401'd hours later — by which point the fleet had been quietly running on a dead refresh mechanism.

**This is the defensive PR** (Equation msg_57311718 + Calculation msg_5641e7d0 finding). Stops the silent-loop blind spot so operators see "stuck" instead of "stuck-without-knowing." Does NOT fix the underlying refresh — that's the next PR (gated on the captured stderr/stdout from the next UNCHANGED event in the wild to discriminate sub-hypothesis 2a/2b).

## Changes

- `credential_refresh.py`:
  - `_refresh_now` now captures `outcome = await self.backend.refresh()` (was dropping it); backend-raises wrapped as `RefreshOutcome.FAILED`
  - New `_propagate_outcome(outcome)`: REFRESHED resets streak + calls `_clear_refresh_broken`; UNCHANGED/FAILED increments + flips at `REFRESH_BROKEN_THRESHOLD=2` (~4 min at 120s poll)
  - `_flip_refresh_broken` walks every registered agent_home, loads runtime, flips `health="refresh_broken"` with operator-actionable error message (`"Run `claude /login` then `puffo-agent agent resume <id>`"`)
  - **Precedence guard**: skips agents already in `auth_failed` / `api_error_abandoned` (those are stronger downstream signals; same operator recovery path)
  - FileBackend UNCHANGED branch now dumps stdout/stderr tails (400 chars each) for next-incident sub-hypothesis 2a/2b discrimination
- `state.py`: `RuntimeState.health` enum + docstring extended for `"refresh_broken"`
- `cli.py:531`: `agent list` surfaces `[refresh_broken]` alongside `[auth_failed]` / `[api_error_abandoned]`

## Tests

10 new in `tests/test_credential_refresher.py` (+23 total in file, all pass):
- Outcome counter mechanics: REFRESHED resets / UNCHANGED + FAILED increment
- Flip fires at exactly `REFRESH_BROKEN_THRESHOLD=2`, not 1
- Subsequent REFRESHED clears `refresh_broken` → `ok` + resets counter
- **Precedence**: `refresh_broken` does NOT overwrite `auth_failed` OR `api_error_abandoned`
- **Smoking-gun regression guard**: `_refresh_now` captures outcome instead of dropping it (this is the bug class that caused the P0)
- `_refresh_now` exception path: backend that raises is `FAILED` (not silently swallowed)
- FileBackend UNCHANGED branch logs both stdout + stderr tails
- `_flip_refresh_broken` fans out across all registered agents

**Test results:** 23/23 focused + 879 passed, 1 skipped (pre-existing permission-mode test) full suite.

## Known limits

- Does NOT fix the underlying refresh — that's the next PR (real-fix), gated on the captured stderr/stdout from the next UNCHANGED event in the wild
- Web/UI consumers (FB-197/198) already read `runtime.health`; surfacing `refresh_broken` in the bubble / agent-detail UI is its own ticket
- The 2-tick threshold (~4 min) is the smallest fast-feedback window that won't false-positive on a single transient subprocess hiccup. If operators want even faster signaling, drop to 1.

## Ship cadence

Operator-directed live-test branch (msg_5bcd284d): operator will start puffo-agent from this branch to verify the visibility flip fires correctly in the real environment before merge. PR opened for diff review in parallel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,31 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   + raw API error. No ``api_error_abandoned`` flip — request-too-large
   is a permanent input-side failure class, not a transient.
 
+- **PUF-265: ``CredentialRefresher`` no longer silently runs on a dead
+  refresh mechanism.** ``_refresh_now`` used to call
+  ``await self.backend.refresh()`` without capturing the returned
+  ``RefreshOutcome``, so the "claude exited 0 but expiresAt didn't
+  advance" case (UNCHANGED) fell through unnoticed. The daemon went
+  back to its 2-minute poll loop, ``runtime.health`` stayed
+  ``"unknown"`` fleet-wide, and operators only noticed once individual
+  agents 401'd hours later.
+
+  ``_refresh_now`` now captures the outcome (exception → ``FAILED``)
+  and feeds ``_propagate_outcome``, which tracks a consecutive
+  non-success streak. After ``REFRESH_BROKEN_THRESHOLD = 2`` ticks
+  (~4 min @ 120s poll), every registered agent's ``runtime.health``
+  flips to ``"refresh_broken"`` with an operator-actionable error
+  (``"Run `claude /login` then `puffo-agent agent resume <id>`"``).
+  Cleared unconditionally on the next REFRESHED tick — daemon
+  restarts can still unstick agents stuck on the previous instance's
+  flipped state. Does not overwrite ``"auth_failed"`` /
+  ``"api_error_abandoned"`` (stronger downstream signals; same
+  operator recovery). ``agent list`` surfaces ``[refresh_broken]``.
+
+  ``FileBackend``'s UNCHANGED branch dumps stdout + stderr tails
+  (400 chars each) at ERROR level for forensic discrimination of the
+  underlying root cause.
+
 - **Codex agent archive/delete failing with ``Permission denied`` on
   ``.codex/tmp/.../.lock`` (Windows).** The codex CLI holds an
   exclusive file lock on ``.codex/tmp/arg0/codex-<id>/.lock`` for the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,21 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   (400 chars each) at ERROR level for forensic discrimination of the
   underlying root cause.
 
+  v2 hardening (post-2026-05-29 09:08 UTC Anthropic-side rate-limit
+  incident): the refresh probe is now pinned to ``claude-haiku-4-5``
+  via ``--model`` (Haiku has higher per-model limits than the
+  operator's interactive Opus/Sonnet default, so the probe stops
+  fighting model-specific rate windows). A new
+  ``RefreshOutcome.RATE_LIMITED`` variant is returned when the
+  probe's stderr/stdout matches a canonical Anthropic rate-limit
+  signature (six anchored patterns: 429 / "temporarily limiting
+  requests" / ``rate_limit_error`` / 5h-quota / 529 overloaded).
+  RATE_LIMITED counts toward the refresh_broken streak (same as
+  FAILED — the rotation really didn't happen) AND schedules a
+  randomised ``[5, 15]`` s fast retry via ``_refresh_request.set()``
+  instead of parking on the natural 120s poll. Back-to-back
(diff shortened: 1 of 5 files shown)


hellohanchen:
Review LGTM as a P0 defensive fix — root cause (uncaptured `RefreshOutcome` in `_refresh_now`) is correctly addressed, regression guard test pins the contract, threshold + precedence rules are sound. One latent stale-state bug worth fixing before merge:

## 1. Daemon-restart leaves agents stuck at `refresh_broken` (needs fix)

`_propagate_outcome` only calls `_clear_refresh_broken()` when the in-memory streak counter is non-zero:

```python
if outcome is RefreshOutcome.REFRESHED:
    if self._consecutive_non_success > 0:   # ← this guard
        logger.info(...)
        self._clear_refresh_broken()
    self._consecutive_non_success = 0
    return
```

**Scenario:**
1. Daemon flips N agents to `refresh_broken` on disk
2. Operator runs `claude /login`, fixes the token, but doesn't restart the daemon yet
3. Daemon restarts (deploy / OOM / unrelated) → `_consecutive_non_success = 0` (fresh init)
4. Next tick returns `REFRESHED` → guard is False → `_clear_refresh_broken()` is skipped
5. Disk state stays `refresh_broken` forever; `agent list` shows stale `[refresh_broken]` indefinitely

**Fix:** drop the `> 0` guard. `_clear_refresh_broken` already short-circuits internally via `if rs.health != "refresh_broken": continue`, so calling it every REFRESHED tick is at worst one file-load per registered agent — cheap and idempotent.

```python
if outcome is RefreshOutcome.REFRESHED:
    if self._consecutive_non_success > 0:
        logger.info(
            "credential refresh recovered after %d non-success tick(s) ...",
            self._consecutive_non_success,
        )
    self._clear_refresh_broken()   # always — survives daemon restart
    self._consecutive_non_success = 0
    return
```

Keep the log line gated on the streak (only meaningful when there was an in-memory streak); but always perform the disk-state cleanup.

**Add test:**

```python
def test_refresh_broken_cleared_after_daemon_restart(tmp_path, monkeypatch):
    """Disk state 'refresh_broken' must be cleared on the next REFRESHED
    outcome even when the in-memory counter is 0 (post-restart case
    where the prior daemon instance flipped the agent before going
    down)."""
    # 1) Seed an agent at refresh_broken on disk directly
    # 2) Build a fresh refresher (counter == 0)
    # 3) Dispatch REFRESHED
    # 4) Assert health flipped back to "ok"
```

## 2. Optional: confirm log destination for stdout/stderr dump

The new UNCHANGED branch logs `claude --print` stdout + stderr tails (400 chars each) at ERROR level for sub-hypothesis 2a/2b discrimination. The PR description calls this out as intentional forensics — fine. Just worth a one-line check from the operator that daemon ERROR logs aren't forwarded to a third-party observability platform where `claude` CLI output might be inappropriate to ship. If it's local systemd journal / internal-only file, ignore.

---

Everything else looks good:

- Root-cause regression guard (`test_refresh_now_captures_outcome_instead_of_dropping`) uses a spy to pin "outcome must be forwarded" — exactly the right shape of test for this bug class
- Exception path correctly counted as `FAILED` instead of silently swallowed (`test_refresh_now_treats_backend_exception_as_failed`)
- Precedence guards (both flip and clear directions) preserve `auth_failed` / `api_error_abandoned` — the stronger downstream signals — with explicit test coverage
- Threshold N=2 well-documented (smallest fast-feedback window that won't false-positive on a single subprocess hiccup), boundary pinned by test
- `elapsed` confirmed defined at `credential_refresh.py:212` — no NameError risk in the new log format
- Idempotent fan-out via in-function `health != "refresh_broken": continue` skip prevents log spam / redundant disk writes once flipped
- Race condition with worker's `runtime.json` writes is bounded to one polling interval; precedence guard prevents daemon from silently overwriting worker-set `auth_failed`
- `cli.py` `agent list` surfacing `[refresh_broken]` is the right minimal UI surface for this signal

PR #53, opened 2026-05-29
fix(PUF-264): request-too-large prevention + reactive Anthropic too-long catch

hellohanchen:
## Summary

Defense-in-depth for the operator-reported case (msg_a42f8cbb): user uploaded 10 PDFs, agent inlined all into one LLM call, Anthropic returned request-size-too-large. Existing daemon retry path retried (same payload → same error); operator restart wouldn't help (next prompt re-inlines).

Two-layer fix:

1. **Proactive pre-send check** in `ClaudeSession._one_turn` — short-circuits before spawning claude subprocess when `user_message` bytes > `MAX_USER_MESSAGE_BYTES` (180KB; under-budget for Anthropic's ~200KB request-body cap leaving headroom for system prompt + tool definitions + headers). No API spend, no retry.

2. **Reactive catch** in `run_turn` — pattern-matches the verbatim strings Claude Code surfaces when Anthropic returns the too-long error, rewrites to friendly message with `request_too_large=reactive` metadata. Crucially fires AFTER the auth-retry loop, so it never triggers spurious retries; no `api_error_abandoned` flip (separate class from retry-exhausted).

Both regex patterns extracted by Calculation from strings-grep of the Claude Code 2.1.144 binary constants (`AQ="Prompt is too long"` + the parseable `"input length and \`max_tokens\` exceed context limit"` form). Confirmed via one synthesized 500KB dev-test that captured the stream-json envelope shape.

## Changes

`src/puffo_agent/agent/adapters/cli_session.py`:
- New `MAX_USER_MESSAGE_BYTES = 180 * 1000`
- New `REQUEST_TOO_LARGE_FRIENDLY` (verbatim from operator's intake)
- New `_REQUEST_TOO_LARGE_PATTERN` regex (both Anthropic forms, IGNORECASE, `\b` word boundary on "long")
- `_one_turn` short-circuits with friendly reply + audit event `turn.request_too_large_pre_send` when user_message exceeds cap (strict `>` not `>=`)
- New `_rewrite_if_request_too_large` helper called from `run_turn` post-auth-retry-loop
- Reactive surface preserves `input_tokens` / `output_tokens` for cost accounting + `original_reply` in metadata + audit event `turn.request_too_large_reactive`

## Tests

5 new in `tests/test_cli_session_recovery.py` (12 total in file, all pass):
- **Proactive pre-send**: 200KB user_message short-circuits, friendly reply, metadata + audit, **no stdin write** (asserted via FakeStdin buffer == bytearray())
- **Boundary**: exactly at 180000 bytes does NOT short-circuit (strict `>`)
- **Reactive Prompt-is-too-long form**: "API Error: Prompt is too long..." → rewrite + raw preserved in `metadata.original_reply` + token counts preserved
- **Reactive max_tokens form**: "input length and `max_tokens` exceed context limit: ..." → same rewrite
- **Passthrough**: normal replies + replies incidentally using "long" or "prompt" + rate-limit replies do NOT trigger (regex anchored on verbatim Anthropic strings)

**Test results:** 12/12 focused + 873 passed, 1 skipped (pre-existing permission-mode test) full suite.

## Known limits / out-of-scope (per intake)

- No `runtime.health` flip — kept clean per intake; operator-visible signal is the inline friendly error
- No retry on rewrite — request-too-large is permanent input-side, not a transient
- Pre-send byte counter doesn't account for system_prompt or memory that Claude Code prepends; reactive layer catches those residuals
- Auto-truncate (Shape B) + auto-chunk (Shape C) deferred to v2; this is Shape A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,51 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **PUF-264: request-too-large no longer infinite-retries + no longer
+  leaks the raw Anthropic API error to chat.** When a user uploaded
+  10 PDFs (msg_a42f8cbb), the agent inlined all of them into one
+  Claude call; Anthropic returned its ``request-too-large`` error, the
+  daemon's auth-retry path retried the same payload, hit the same
+  error, exhausted, and surfaced the verbatim CLI error string to the
+  user. A daemon restart wouldn't help — the next prompt re-inlines
+  the same attachments.
+
+  Two-layer defense in ``cli_session.py``:
+
+  1. **Proactive pre-send byte cap.** ``_one_turn`` short-circuits
+     before spawning the claude subprocess when
+     ``len(user_message.encode("utf-8")) > MAX_USER_MESSAGE_BYTES``
+     (180 KB; under-budget for Anthropic's ~200 KB request-body cap
+     with headroom for system prompt + tool definitions + headers).
+     Returns the friendly user-facing reply with metadata
+     ``request_too_large=pre_send`` and audit event
+     ``turn.request_too_large_pre_send``. No API spend, no retry, no
+     ``runtime.health`` flip — the user resends a smaller message.
+     Byte-based check (not character-based) so 60k CJK chars × 3 bytes
+     trip the cap even though the char count is below it.
+
+  2. **Reactive regex catch.** ``_rewrite_if_request_too_large`` (via
+     the new ``_looks_like_request_too_large`` predicate, mirroring
+     ``_looks_like_auth_error`` / ``_looks_like_poisoned_session``)
+     rewrites three verbatim Anthropic surfaces to the friendly
+     reply: ``"Prompt is too long"`` (Claude Code binary constant
+     ``AQ``), ``"input length and `max_tokens` exceed context limit"``
+     (parseable numeric form), and ``"size error: request too large,
+     try with a smaller file"`` (file-attachment surface; matches
+     both ASCII ``:`` ``,`` and the fullwidth ``：`` ``，`` Anthropic
+     emits on some localised builds). Fires AFTER the auth-retry loop
+     so a real auth failure still gets retried — only when the reply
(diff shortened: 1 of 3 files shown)


hellohanchen:
Review LGTM — ship-able. One small defensive cleanup (optional, not blocking):

**`_rewrite_if_request_too_large` drops `tool_calls`**

`TurnResult` has 5 fields (`reply / input_tokens / output_tokens / tool_calls / metadata`). The reactive rewrite reconstructs `TurnResult(...)` preserving only the two token counts, so `tool_calls` defaults to 0 instead of carrying over `result.tool_calls`.

In practice today the count is almost certainly 0 (the API too-long error fires before any tool runs in this turn), so metrics aren't affected. But if Claude Code ever surfaces a turn that interleaves tool_use then ends with a too-long error, we'd silently under-count.

```python
return TurnResult(
    reply=REQUEST_TOO_LARGE_FRIENDLY,
    metadata=new_md,
    input_tokens=result.input_tokens,
    output_tokens=result.output_tokens,
    tool_calls=result.tool_calls,   # <— add this
)
```

And add `tool_calls=...` to the `TurnResult(reply=raw, input_tokens=42, ...)` construction in `test_rewrite_if_request_too_large_matches_canonical_anthropic_string` + assert it's preserved on output.

---

Everything else looks good:
- Defense-in-depth split (pre-send cap + reactive regex) is the right shape
- Regex anchors (`\b`, optional `API Error:` prefix, verbatim backticks) covered by passthrough tests
- 180KB cap with ~20KB headroom for system prompt + tools is well-documented
- Subprocess invariant preserved (no stdin write, no `turn.input` audit on pre-send short-circuit) and explicitly asserted in tests
- Reactive path stays out of the auth-retry loop, so no spurious retries on permanent input failures
- Token counts preserved for cost accounting; raw API error preserved in `metadata.original_reply` for forensics
- Boundary test (exactly at cap does NOT short-circuit) pins the strict `>`

Out-of-scope items called out in the PR description (Shape B auto-truncate / Shape C auto-chunk / system_prompt size accounting) are the right scope cut for a defensive fix.

PR #54, opened 2026-05-29
fix(PUF-266): merge host codex MCP servers into per-agent config.toml

hellohanchen:
## Summary

Codex + cli-local agents used to see only the puffo MCP server because `write_codex_mcp_config` wrote a fresh per-agent `config.toml` that ignored the operator's own `~/.codex/config.toml` (or `$CODEX_HOME/config.toml` per PR #54 review). Any MCP servers the operator had installed for their own codex CLI were invisible to spawned agents.

Stage-3 resolved both intake open questions by code-read:
- **Q1 (host paths):** codex config = `~/.codex/config.toml` (or `$CODEX_HOME/config.toml`); skills surface = NONE (`harness/codex.py:20-25` denies `.claude/skills/`); plugins surface = NONE
- **Q2 (plugins):** no codex plugins concept; codex-rs binary state lives in `$CODEX_HOME` as `auth.json` + `config.toml` only

Scope therefore collapses to **MCP-only**, matching operator's actionable complaint.

## Changes

1. `portal/state.py` — new `read_host_codex_mcp_servers(host_home)`: parses codex's `config.toml` with tomllib, returns `{name: {command, args, env}}`. **Honours `$CODEX_HOME` env override** (codex CLI does — PR #54 review item 2). Defensive: missing/malformed file/spec → `{}` (no raise, agent startup unblocked). Skips malformed entries with missing/empty `command` (PR #54 review item 4a).

2. `mcp/config.py` — `write_codex_mcp_config` gains `extra_servers` param. New `_emit_codex_mcp_block` helper DRYs the puffo + extras paths. New `_toml_key` helper quotes MCP names containing non-bare-charset chars (e.g. `my.server` → `"my.server"`) so they don't misparse as nested tables (PR #54 review item 4b). Extras emitted first in sorted order; puffo-keyed extras skipped (daemon's puffo entry is authoritative, no duplicate-key TOML).

3. `agent/adapters/local_cli.py::_ensure_codex_session` reads host MCPs each spawn + passes as `extra_servers`. Codex agents without `puffo_core` configured still receive the host MCP catalog.

## Tests

**12 new** across 3 files (23 total in the touched files, all pass):

`tests/test_codex_config.py`:
- Host has 2 MCP entries + puffo configured → emitted TOML has 3 blocks; host command/args/env preserved verbatim
- Operator without puffo_core still inherits host MCP catalog
- Host has a third-party `[mcp_servers.puffo]` entry → daemon's entry shadows it (no duplicate-key TOML)
- No-extras default keeps the pre-PUF-266 shape exactly (regression guard)
- `_toml_key` bare-keys ASCII names; quotes names with dots (PR #54 review 4b)

`tests/test_codex_auth_link.py`:
- `read_host_codex_mcp_servers` defensive paths: missing host file → `{}`; malformed TOML → `{}`; top-level non-mcp keys ignored
- `$CODEX_HOME` env override honoured (PR #54 review item 2)
- Default `~/.codex/config.toml` when `$CODEX_HOME` unset
- Skip malformed entries with no/empty/non-string `command` (PR #54 review item 4a)

`tests/test_local_cli_codex_mcp_wiring.py` (new):
- `_ensure_codex_session` integration test: seed fake home with `.codex/config.toml` + monkeypatch `Path.home`, run full helper, assert host MCP block AND puffo block both present in per-agent `config.toml`. Future refactor that drops `extra_servers=host_mcps` breaks loudly. (PR #54 review item 1)
- Wiring variant: no host MCPs → only puffo block emitted, no spurious extras

**Test results:** 23/23 focused + **950 passed**, 1 skipped (pre-existing permission-mode test) full suite.

## Intake corrections (Stage-3 Q1+Q2 resolution)

- **Intake Step 1** (extend `_verify` for codex): N/A. The short-circuit at L880-886 is correctly about NOT touching `~/.claude/*` for codex; MCP merge belongs in `_ensure_codex_session` (runs every spawn, picks up host config changes between worker restarts) — that's where the fix landed.
- **Intake Step 3** (`docker_cli.py` codex parity): N/A. cli-docker has no codex support at all — separate gap, not PUF-266.

## Known limits / out-of-scope

- No skills propagation (codex has no skills surface per Q1)
- No plugins propagation (codex has no plugins concept per Q2)
- cli-docker codex support not addressed (codex isn't supported there yet)
- Operator-installed `puffo` MCP shadowing: daemon's version wins; operator should rename a custom puffo entry if they need both
- **Codex MCP-server schema subset:** only `{command, args, env}` are propagated. If codex supports additional fields (`cwd` / `disabled` / `startup_timeout_sec`), those drop silently from the host config. Documented in `read_host_codex_mcp_servers` docstring; widen there + `_emit_codex_mcp_block` together when a future codex field becomes load-bearing. (PR #54 review item 3)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,27 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **PUF-266: codex agents now inherit the operator's host MCP catalog.**
+  ``write_codex_mcp_config`` used to overwrite the per-agent
+  ``$CODEX_HOME/config.toml`` with only the puffo_core stdio entry, so
+  any MCP server the operator had installed for their own codex CLI
+  (filesystem, github, etc.) was invisible to spawned agents.
+
+  New ``read_host_codex_mcp_servers(host_home)`` parses the host's
+  ``[mcp_servers.*]`` blocks (honouring ``$CODEX_HOME`` override) and
+  ``_ensure_codex_session`` forwards them via the new
+  ``extra_servers`` param. The puffo entry shadows any same-named host
+  entry to avoid duplicate-key TOML. Malformed entries (missing /
+  empty / non-string ``command``) are skipped — they'd otherwise land
+  as ``command = ""`` and crash codex on the empty argv. Server names
+  containing TOML-significant chars (``my.server``) are emitted as
+  quoted basic-string keys via ``_toml_key`` so they don't get
+  misparsed as nested tables. Spec surface is ``{command, args, env}``
+  only — any other codex MCP-server fields (cwd / disabled /
+  startup_timeout_sec) drop silently; widen here +
+  ``_emit_codex_mcp_block`` together if a future codex schema field
+  becomes load-bearing.
+
 - **PUF-258: ``runtime.health = "auth_failed"`` no longer sticky after
   credential refresh-success.** The daemon set the flag in
   ``_handle_suppressed_reply`` (PUF-221) but nothing cleared it
--- a/src/puffo_agent/agent/adapters/local_cli.py
+++ b/src/puffo_agent/agent/adapters/local_cli.py
@@ -34,6 +34,7 @@ from ...portal.state import (
     home_dir,
     link_host_codex_auth,
     link_host_credentials,
+    read_host_codex_mcp_servers,
     seed_claude_home,
     sync_host_enabled_plugins,
     sync_host_mcp_servers,
@@ -290,21 +291,34 @@ class LocalCLIAdapter(Adapter):
         # Always write config.toml — pins ``cli_auth_credentials_store``
         # to "file" so codex uses ``$CODEX_HOME/auth.json`` (not macOS
         # Keychain) and our symlink/refresh model works. MCP section
-        # only when puffo_core is configured.
+        # only when puffo_core is configured. PUF-266: host's own MCP
+        # entries from ``~/.codex/config.toml`` are merged in so the
+        # agent inherits the operator's codex MCP catalog (the puffo
+        # entry below shadows any same-named host entry).
+        host_mcps = read_host_codex_mcp_servers(Path.home())
         if self.puffo_core_mcp_env:
             write_codex_mcp_config(
                 codex_home / "config.toml",
                 command=default_python_executable(),
                 args=["-m", "puffo_agent.mcp.puffo_core_server"],
                 env=self.puffo_core_mcp_env,
+                extra_servers=host_mcps,
             )
         else:
-            write_codex_mcp_config(codex_home / "config.toml")
+            write_codex_mcp_config(
+                codex_home / "config.toml",
+                extra_servers=host_mcps,
+            )
             logger.warning(
                 "agent %s: codex MCP tools unavailable — puffo_core is "
                 "not configured. populate `puffo_core:` in agent.yml to "
                 "enable send_message / list_channels / etc.",
                 self.agent_id,
+            )
+        if host_mcps:
+            logger.info(
+                "agent %s: merged %d host MCP server(s) into codex config: %s",
+                self.agent_id, len(host_mcps), sorted(host_mcps),
             )
 
         env = {
--- a/src/puffo_agent/mcp/config.py
+++ b/src/puffo_agent/mcp/config.py
@@ -13,12 +13,15 @@ that form.
 from __future__ import annotations
 
 import json
+import re
 import site
 import sys
 from pathlib import Path
 
 
 MCP_SERVER_NAME = "puffo"
+
+_TOML_BARE_KEY = re.compile(r"[A-Za-z0-9_-]+")
 
 
 PUFFO_CORE_TOOL_NAMES = (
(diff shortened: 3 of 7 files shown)


hellohanchen:
Review LGTM — scope cut to MCP-only is the right call (codex has no skills/plugins concept, confirmed by Stage-3 code-read). Three-layer split is clean: `state.py` reads host config, `mcp/config.py` assembles the TOML, `local_cli.py` wires them per-spawn. A few items to address before merge:

## 1. Missing wiring test on `_ensure_codex_session` (recommended)

`read_host_codex_mcp_servers` and `write_codex_mcp_config(extra_servers=...)` are each well-tested in isolation, but **the integration in `_ensure_codex_session` — the actual call site — has no test**. A future refactor that drops the `extra_servers=host_mcps` parameter wouldn't fail any test. The PR description's line "_ensure_codex_session invokes read_host_codex_mcp_servers(Path.home()) each spawn and passes result through" describes intent, not coverage.

Add one wiring test, e.g.:

```python
def test_ensure_codex_session_merges_host_mcps_into_config_toml(
    tmp_path, monkeypatch,
):
    """Wiring guard: _ensure_codex_session must read host MCPs and
    forward them through write_codex_mcp_config — otherwise a future
    refactor could silently drop the extra_servers parameter."""
    # 1) Seed fake home with .codex/config.toml containing one host MCP
    # 2) monkeypatch Path.home → tmp_path
    # 3) Spawn the session (mock codex subprocess)
    # 4) Read the per-agent config.toml, assert host MCP block present + puffo block present
```

## 2. `$CODEX_HOME` env var override is ignored

`read_host_codex_mcp_servers` hardcodes `host_home / ".codex" / "config.toml"`. codex CLI honors `$CODEX_HOME` to override the default location, so an operator who sets it will silently get an empty host MCP catalog.

Two-line fix:

```python
def read_host_codex_mcp_servers(host_home: Path) -> dict[str, dict]:
    codex_home_env = os.environ.get("CODEX_HOME")
    codex_home = Path(codex_home_env) if codex_home_env else host_home / ".codex"
    host_config = codex_home / "config.toml"
    ...
```

Plus a one-line monkeypatch test asserting the env-var path is read instead of `~/.codex`.

## 3. Confirm codex MCP spec is exactly `{command, args, env}` (open question)

`_emit_codex_mcp_block` (and `read_host_codex_mcp_servers`) only handle three fields. If codex's MCP server schema supports anything else commonly used (e.g., `startup_timeout_sec`, `tool_timeout_sec`, `cwd`, `disabled`), those fields in the operator's host config will be silently dropped after the merge — operator's `[mcp_servers.X]` settings won't fully survive.

Action: skim codex-rs's MCP server config struct (rough grep: `mcp_servers` / `McpServerConfig` in the codex repo). If it's only `{command, args, env}`, no change needed; if there are more fields, decide between (a) emit them through too, or (b) document the limitation in the helper docstring + emit a one-line `logger.warning` when we drop unknown keys.

## 4. Defensive hardening (low priority)

**(a) Empty-command host entry produces invalid output.** A malformed host MCP entry (missing `command`) currently lands as `command = ""` in the emitted TOML, then codex tries to spawn an empty command and fails per-entry. Skipping is safer:

```python
cmd = spec.get("command")
if not isinstance(cmd, str) or not cmd:
    continue
out[name] = {
    "command": cmd,
    "args": list(spec.get("args") or []),
    "env": dict(spec.get("env") or {}),
}
```

**(b) MCP name not TOML-key-escaped.** `f"[mcp_servers.{name}]"` directly interpolates. A host entry named `my.server` would emit `[mcp_servers.my.server]`, which TOML parses as nested tables `mcp_servers.my.server` — not the operator's intent. codex upstream likely restricts name charset, but cheap to harden:

```python
def _toml_key(name: str) -> str:
    """Bare key if [A-Za-z0-9_-]+, else quoted basic-string key."""
    if re.fullmatch(r"[A-Za-z0-9_-]+", name):
        return name
    return f'"{_toml_escape(name)}"'
```

**(c) Cosmetic:** `tomllib.TOMLDecodeError` is a subclass of `ValueError`, so the except tuple `(OSError, ValueError, tomllib.TOMLDecodeError)` is redundant — `(OSError, ValueError)` suffices. Not worth its own change but easy to clean up while you're here.

---

Everything else is solid:

- Precedence guard on the puffo key (`if name == MCP_SERVER_NAME: continue`) prevents duplicate-key TOML AND a third-party imposter from shadowing the daemon's puffo MCP — explicitly covered by `test_extra_servers_named_puffo_is_shadowed_by_puffo_entry`
- Regression guard (`test_no_extras_unchanged_from_pre_puf266`) pins the pre-PUF-266 emitted shape
- Operator without `puffo_core` still inherits the host MCP catalog (`test_extra_servers_pass_through_when_puffo_unconfigured`)
- Defensive read path: missing file / malformed TOML / non-dict spec / non-dict mcp_servers root all return `{}` instead of raising — agent startup can't be blocked by a broken host config
- Deterministic `sorted(...)` iteration order keeps the emitted TOML diff-clean across spawns
- `Path.home()` is correct in the cli-local context (daemon and operator share a host); cli-docker codex is explicitly out of scope per the intake corrections
- Logging the merged MCP names (`sorted(host_mcps)`) at info level is the right balance — operator-actionable, no secret leakage from env values
- PR description says "11 new" but the diff shows 8 new tests; cosmetic, not blocking

PR #61, opened 2026-06-03
fix: retry archive/delete shutil.move on Windows codex sqlite handle linger

hellohanchen:
## Summary

Operator hit ``archive failed: [WinError 32] another program is using this file ... '.codex/logs_2.sqlite'`` when archiving a codex agent.

- ``_drain_codex_tmp`` only cleans ``.codex/tmp/`` (the older codex CLI lock).
- The codex App Server's ``logs_2.sqlite`` (WAL-mode session log) sits at ``.codex/logs_*.sqlite`` and its handle lingers a few hundred ms past subprocess termination.
- Existing reconciler-tick retry eventually recovers but spams an error log every tick until Windows releases the handle.

## Fix

Add ``_retry_on_oserror(fn, attempts=5, delay=0.5)`` helper, mirroring ``_drain_codex_tmp``'s existing inline retry. Wrap the outer ``shutil.move`` (archive) and ``shutil.rmtree`` (delete) with it — usually one extra 500ms wait clears the handle and the second attempt succeeds.

Does NOT pre-delete ``logs_*.sqlite``: the whole point of archive is to preserve the dir under ``archived/<id>-ws-<ts>/`` for later inspection, and the session log is part of that history.

## Test plan

- [x] pytest tests/test_archive_codex_tmp_drain.py -v → 11/11 pass
- [x] pytest tests/test_archive_codex_tmp_drain.py tests/test_pause_resume_archive.py tests/test_bridge_handlers.py -q → 40/40 pass
- [ ] Manual Windows: create + archive a codex agent → no error log; archived dir lands in ~/.puffo-agent/archived/ on first attempt within 500ms

## Test additions

5 new _retry_on_oserror unit cases (first-try success, transient lock + retry, exhaustion re-raise, custom attempts kwarg, non-OSError pass-through) + 1 integration case driving Daemon._archive_on_flag through a flaky shutil.move mock.

### DIFF
--- a/src/puffo_agent/portal/daemon.py
+++ b/src/puffo_agent/portal/daemon.py
@@ -327,7 +327,7 @@ class Daemon:
         stamp = time.strftime("%Y%m%d-%H%M%S")
         dest = archived_dir() / f"{agent_id}-ws-{stamp}"
         try:
-            shutil.move(str(src), str(dest))
+            await _retry_on_oserror(lambda: shutil.move(str(src), str(dest)))
             logger.info("agent %s: archived to %s", agent_id, dest)
         except OSError as exc:
             logger.error(
@@ -350,7 +350,7 @@ class Daemon:
             return
         await _drain_codex_tmp(src)
         try:
-            shutil.rmtree(src)
+            await _retry_on_oserror(lambda: shutil.rmtree(src))
             logger.info("agent %s: deleted", agent_id)
         except OSError as exc:
             logger.error(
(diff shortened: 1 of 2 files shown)


hellohanchen:
Closing — the retry approach treated the symptom not the root cause and ALSO had a bug: shutil.move's copy-fallback creates a partial dest on the first failure, then subsequent retries nest src inside it ("Destination path X already exists"). Proper fix is #62: codex teardown closes stdin first so sqlite WAL flushes synchronously through Rust Drop before process exit, sidestepping the file-lock-linger entirely.

PR #85, opened 2026-06-15
PUF-303: refresh_broken DM + reactive-clear (extends PUF-283)

hellohanchen:
## Summary

Sam Claude OAuth credential refresh incident — extend PUF-283 `auth_failed` notification-subsystem + worker-side reactive clear to also cover `refresh_broken` state-transitions (Substances B+C+D+E together).

- New `CredentialRefresher._on_refresh_broken_enter` callback list + `register_/unregister_/_fire_` triplet (parallel to `_on_refresh_success`).
- `_flip_refresh_broken` collects newly-broken agents (was_ok semantics) and fires the callback per-agent so workers can DM the operator.
- New `Worker._on_refresh_broken_enter` dedup-gated DM dispatch + `_notify_operator_of_refresh_broken` async send (mirrors `_on_auth_failed_enter`).
- New bilingual copy `format_refresh_broken` distinguishes daemon-side refresh failure from in-flight 401.
- New `Worker._clear_refresh_broken_if_recoverable` symmetric to `_clear_auth_failed_if_recoverable`; fires from existing `on_refresh_success` so manual `claude auth login` recovery doesn't wait for next poll-tick.
- Daemon wires per-worker callback with agent_id filter; `_stop_worker` unregisters from both Claude + Codex refreshers.

Substance (A) `claude auth status` semantic gap → out-of-scope (upstream Claude CLI); flag for PUF-302 docs.

## Test plan

- [x] `python -m pytest tests/test_refresh_broken_notification.py tests/test_refresh_broken_clear.py tests/test_auth_failed_dm_notification.py` → 32/32 (13 new + 19 PUF-283 regression).
- [x] Bilingual copy invariants (zh + en, bold display name, fallback when display_name empty).
- [x] `_flip_refresh_broken` fires callback only for was_ok → broken (skips already-broken).
- [x] Worker dedup gate fires once per refresh_broken episode + re-arms after reset.
- [x] `_notify_operator_of_refresh_broken` skip paths: no-warm-client (re-arm), no-operator-slug (stay gated).
- [x] Happy-path send to operator slug with `root_id=""`.
- [x] `_clear_refresh_broken_if_recoverable` flips refresh_broken → ok with info log; no-op for other health states.
- [ ] Live soak post-deploy: Sam-class repro (rc=1 ×2 → bilingual DM → manual `claude auth login` → next worker turn → runtime.health clears without poll-tick wait).

Closes PUF-303.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,12 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   reports a daemon swap explicitly when a new daemon takes over
   mid-shutdown, and `puffo-agent start` against an already-running
   daemon now exits 0 (the user's intent is met) instead of erroring.
+- **The operator is DM'd when an agent's credential refresh breaks.**
+  When the daemon can't roll a Claude/Codex token (refresh keeps
+  failing), the agent now DMs its operator bilingual recovery
+  instructions (run `claude auth login`) — once per episode — mirroring
+  the existing auth-failed notification. A manual re-login clears the
+  state on the agent's next turn instead of waiting for the next poll.
 
 ## [0.12.4] — 2026-06-12
 
--- a/src/puffo_agent/agent/_invite_strings.py
+++ b/src/puffo_agent/agent/_invite_strings.py
@@ -118,3 +118,23 @@ def format_oauth_expired(agent_id: str, agent_display_name: str = "") -> str:
         "⚠️ Claude OAuth 已过期。请运行 `claude auth login`，"
         "然后给我发条消息即可恢复。"
     )
+
+
+def format_refresh_broken(agent_id: str, agent_display_name: str = "") -> str:
+    """PUF-303: bilingual (zh+en) operator DM for an agent whose
+    credential-refresh has been failing repeatedly. Mirrors the
+    ``format_oauth_expired`` shape; framing distinguishes the
+    daemon-side refresh failure (``claude auth status`` may still
+    report logged-in but live probe / refresh is broken) from the
+    in-flight 401 path."""
+    label = (
+        f"**{agent_display_name}** (`{agent_id}`)"
+        if agent_display_name else f"`{agent_id}`"
+    )
+    return (
+        f"⚠️ {label} — Claude credential refresh is broken (the daemon "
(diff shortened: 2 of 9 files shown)


hellohanchen:
Closing — the proactive `refresh_broken` DM is the wrong signal. Three problems, in order of importance:

**1. Redundant with the existing `auth_failed` path.** A failed daemon-side refresh does not mean the agent's token is unusable — the agent typically still holds a valid token and keeps working. The case that actually needs operator action is the token *expiring*, and that's already covered: when an in-flight request 401s, the agent DMs the operator via `format_oauth_expired` (PUF-283). So `refresh_broken` either fires while the token still works (premature) or duplicates `auth_failed` (token actually expired). It adds no actionable signal.

**2. False positives on every token cycle (margin mismatch).** `REFRESH_SAFETY_MARGIN_SECONDS = 600` (10 min), but Claude Code only rotates the token within ~5 min of expiry. In the 10→5 min window the daemon triggers a refresh each tick and Claude returns `UNCHANGED` (still fresh by its own threshold). `_propagate_outcome` counts `UNCHANGED` toward `_consecutive_non_success`, so after `REFRESH_BROKEN_THRESHOLD = 2` ticks `_flip_refresh_broken` fires — then clears once Claude actually refreshes at ~5 min. As a UI status dot this was tolerable; escalated to a DM it means the operator gets a false "refresh is broken" alert on every token cycle.

**3. N notifications for one system-level event.** The Claude credential is shared (`~/.claude/.credentials.json`). `_flip_refresh_broken` flips *every* registered agent to `refresh_broken`, and the dedup flag is per-worker, so each agent DMs its operator independently. One credential failure → one DM per Claude agent sharing it. On a multi-agent daemon that's a flood.

The reactive `auth_failed` (a real 401) is the correct and sufficient signal — the daemon's ability to roll a token shouldn't page the user. If it can roll it, great; if not, `auth_failed` covers the eventual expiry.

Separately (not this PR): the `refresh_broken` state machine itself looks prone to false positives — `UNCHANGED` arguably shouldn't count toward the streak, and/or the margin should align with Claude's refresh threshold. Worth a follow-up for the UI state, independent of any notification.

PR #102, opened 2026-06-28
feat(api-puffo): cloud-hosted agent runtime — bundle ingestion + WS receive + LLM tool loop

hellohanchen:
## Summary

PR #102 was built against a placeholder cloud-server shape. The real `puffo-server/cloud_agent` (on the `dev` branch) ships a documented WS wire protocol — [`BRIDGE-WIRE-PROTOCOL.md`](https://github.com/puffo-ai/puffo-server/blob/dev/roadmap/cloud-agent/BRIDGE-WIRE-PROTOCOL.md). The two diverge on every load-bearing decision:

| | placeholder (old) | real spec (this PR) |
|---|---|---|
| URL | `<base>/v1/ws/<slug>` | `<base>/v2/cloud-agents/subscribe` |
| Auth | `Authorization: Bearer …` | `x-sandbox-token: …` on the upgrade only |
| Client crypto | KEM secret + decrypt_message locally | **NONE** — server holds keys; frames are plaintext |
| Tools | HTTP POST per tool | All WS frames: `send` / `fetch_pending` / `ack` / `list_spaces` |
| Inbound | encrypted envelope + sender pubkey | `message` frame with `plaintext` already opened |
| Heartbeat | aiohttp default ping/pong | client-driven `heartbeat` every 30s; `HEARTBEAT_TIMEOUT_SECS = 90` server-side |
| On-connect | passive | wait `connected` → `fetch_pending` → drain → `ack` → main loop |

### What changes

**Bundle / keystore** drop `kem_secret_key` / `kem_cert` / `device_id`. Server resolves identity from `sandbox_token` on the WS upgrade alone.

**`cloud_client.py`** splits into:
- `CloudLlmClient` (HTTP `POST /v1/llm/complete`) — bridge spec is mute on LLM; this endpoint is the placeholder convention we already use, kept bearer-authed.
- `CloudBridgeClient` (WS) — implements the wire protocol end to end: `x-sandbox-token` upgrade header, first-frame `connected` handshake, 30s `heartbeat`, per-call ack correlation via `client_ref` UUID, `BridgeError` carrying server `code` + `message`.

[...]

PR #105, opened 2026-06-29
refactor(puffo-agent): uv workspace + slim puffo-agent-cloud (thin-ification Stage A)

Glimmer0x:
**Stacks on PR #102 (`feat/api-puffo`). @Ming — coordinate, this restructures the cloud runtime you built.**

Severs the cloud runtime from the fat code so a **slim, keyless `puffo-agent-cloud`** installs into an E2B sandbox without Qt/portal. Stage A of the thin-ification (see `puffo-server` `roadmap/cloud-agent/CLOUD-AGENT-PACKAGING-EVAL.md`).

## What
- **`packages/puffo-agent-core`** — stdlib-only shared kernel (paths, profile helpers), `dependencies = []`.
- **`packages/puffo-agent-cloud`** — slim cloud runtime: `agent/api_puffo` moved here (`git mv`, history preserved) + new **`cloud_http.CloudMetadataClient`** (4 read-only GETs, **no subkey signing**; token via E2B injection) + thin `CloudAgentConfig` + `__main__`. **deps = `aiohttp` + `pyyaml` + core ONLY** — no `pyside6`/`psutil`/`cryptography`/`anthropic`. Severed from the fat `crypto/http_client`.
- **`puffo-agent` (fat)** — `portal` imports rewritten to `puffo_agent_core`; still works.

## Gate (coordinator, run synchronously)
**Slim-import gate GREEN:** clean py3.11 venv with **only core+cloud** → `puffo_agent_cloud.runner` + `cloud_http` import; **`pyside6` ABSENT, `psutil` ABSENT**. All changed `.py` compile; `puffo_agent_core` imports stdlib-only.

## Next
Stage B (remaining module migrations into `core`) documented separately. The thin runtime still needs the fat-strip finalized per `PUFFO-AGENT-HTTP-CLIENT-FIX-for-ming.md` (no `/devices/subkeys`, no `/certs/sync`).


Glimmer0x:
@hellohanchen — review request: this is the puffo-agent cloud-agent work toward `main` (puffo-agent has no `dev` branch, so this stack is the path). **#105** = thin-ification (uv workspace: `puffo-agent-core` + slim `puffo-agent-cloud`, keyless, no PySide6). **#110** stacks on it = the publish CI/CD for the split packages. Could you review → merge up the chain? The slim `puffo-agent-cloud` is what the E2B sandbox installs; the real end-to-end chain was already proven green against puffo-server. (Note: earlier comments said "@Ming" — that wasn't a real handle; you're the maintainer here.)

hellohanchen (commented):
Read every changed file end-to-end (skipping `uv.lock`). Overall this is a clean Stage A — the re-export shim model in `state.py` / `profile_sync.py` keeps all existing call sites resolving unchanged (zero fat-package call-site churn), the slim invariant is genuinely defended by two gates (in-repo `test_cloud_slim_import.py` via `builtins.__import__` patch + CI `slim-import` job on a fresh ubuntu env asserting `find_spec(puffo_agent | psutil | PySide6) is None`), and `CloudAgentConfig` / `CloudMetadataClient` come with clear Stage B markers so the scope is honest about what's not done yet.

Full `pytest` (with `src` + both `packages/*/src` on `PYTHONPATH`): **1708 passed / 9 skipped** — no regressions on this branch's baseline.

A few items I left as inline comments above. A few more that didn't fit on a single line:

- **All three packages share `version = "1.0.3"`.** `PUBLISH.md` §149 already recommends independent versions going forward; worth coordinating that bump with the dependency-pinning fix in `PUBLISH.md` §107-127 (the cloud wheel's unpinned `puffo-agent-core`) and §129-145 (the fat wheel's unpinned `puffo-agent-core` / `puffo-agent-cloud`). Both are explicit Stage-A-deferred items but together they're the first real blocker for a public PyPI publish.

- **Test coverage gaps in the new modules**: `CloudAgentConfig.load` has no unit tests for YAML edge cases (missing `runtime` block, missing required fields, malformed YAML). `__main__._provisioned_agent_id` has no test for the zero-agent (`return None`) and multi-agent paths.

- **`CloudMetadataClient` doesn't distinguish 200-with-non-JSON-body from a real 2xx.** The fat package's `PuffoCoreHttpClient._request` raises explicitly on `200 + non-JSON body` (CDN interstitials, proxy error pages). If this client gets wired into the runner's tool surface in Stage B and ever sits behind a CDN, the same defense would be cheap to port.

- The metadata client is **delivered + unit-tested but not yet wired into `tools.py`** — already called out in `cloud_http.py:11` ("Stage A delivers + unit-tests this client; wiring it into the runner's live tool surface is Stage B"). Just flagging for tracking.

No objections to merge from my side; the items above are follow-up coordination, not blockers.

.github/workflows/ci.yml
```diff
@@ -55,3 +55,37 @@ jobs:
...
+  # `import puffo_agent_cloud.__main__` below fails and the gate is red.
+  slim-import:
+    name: slim-import (cloud thin-deps gate)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
hellohanchen, line 71: The main `pytest` job (line 28-29) runs on a `python-version: ["3.11", "3.12"]` matrix, but `slim-import` hardcodes 3.12. A 3.11-only import issue (a stdlib shape that changed between 3.11/3.12, a feature this package quietly relies on, etc.) would slip past this gate while still failing for 3.11 users of the published wheel.

Suggest mirroring the same matrix here:

```yaml
slim-import:
  strategy:
    fail-fast: false
    matrix:
      python-version: ["3.11", "3.12"]
  steps:
    - uses: actions/setup-python@v5
      with:
        python-version: ${{ matrix.python-version }}
```

packages/puffo-agent-cloud/src/puffo_agent_cloud/cloud_http.py
```diff
@@ -0,0 +1,83 @@
...
+        return {}
+
+    async def _get(self, path: str) -> Any:
+        sess = await self._ensure_session()
+        url = f"{self.base_url}{path}"
+        async with sess.get(url, headers=self._headers()) as resp:
+            text = await resp.text()
+            if resp.status >= 400:
+                raise CloudMetadataError(resp.status, text)
+            return await resp.json() if text else None
```
hellohanchen, line 62: Minor: `await resp.text()` on line 59 already consumed the body once. `await resp.json()` here only works because aiohttp internally caches `_body` and re-parses it — not a bug, but it does an extra UTF-8 decode + JSON parse for no reason, and the double-await reads as if you expect two network reads.

More direct:

```python
import json
...
return json.loads(text) if text else None
```

Also silently catches the "server returned 200 with an HTML interstitial body" case as a `json.JSONDecodeError` — `aiohttp.ContentTypeError` from `resp.json()` would be more opaque to a caller debugging a CDN-blackhole.

packages/puffo-agent-cloud/src/puffo_agent_cloud/__main__.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def _provisioned_agent_id() -> str | None:
+    """The single provisioned agent (an ``agents/<id>/agent.yml``)."""
+    adir = agents_dir()
+    if not adir.exists():
+        return None
+    for child in sorted(adir.iterdir()):
+        if child.is_dir() and (child / "agent.yml").exists():
+            return child.name
```
hellohanchen, line 30: Silent first-match: the docstring says "the single provisioned agent", but if `agents_dir()` ever ends up with more than one `<id>/agent.yml` (a redelivery race, a leftover from a prior bundle drop, a test artifact), only the first-sorted entry is picked with no signal.

Worth either logging a warning + still returning the first, or returning `None` so `main()` surfaces the ambiguity explicitly. Even just:

```python
matches = [c.name for c in sorted(adir.iterdir())
           if c.is_dir() and (c / "agent.yml").exists()]
if len(matches) > 1:
    logger.warning("multiple provisioned agents found: %s; running %s",
                   matches, matches[0])
return matches[0] if matches else None
```

Glimmer0x:
Closing — superseded by the fat-runtime pivot (2026-07-07). The cloud agent is now the **fat puffo-agent + keyless transport swap** (E2B egress proxy stamps identity; sandbox holds no keys), not a separately-packaged thin `puffo-agent-cloud`. This thin-ification track (Stage A) is no longer on the launch path — the fat E2B template is what ships. Reopen if the slim-packaging direction is ever revived.

PR #127, opened 2026-07-07
T23 phase 1: opt-in keyless bridge transport foundation (HOLD)

Glimmer0x:
## HOLD — do not merge (human review required)

T23 phase 1: opt-in keyless "bridge" transport seam for the fat agent. **Default behavior is byte-identical** — when `puffo_core.transport` is absent from agent.yml, every code path and the saved agent.yml key-set are exactly today's native signed-crypto behavior.

### What's in

- **`agent/bridge_client.py` (new)** — `CloudBridgeClient` + `BridgeError` + `BridgeClosed`, ported from the thin runtime's `packages/puffo-agent-cloud/src/puffo_agent_cloud/cloud_client.py` (proven in E2B, PR #157), frame semantics unchanged. Wire spec: `puffo-server/roadmap/cloud-agent/BRIDGE-WIRE-PROTOCOL.md`. No imports from `crypto/`. `CloudLlmClient` intentionally not ported (retired `/v1/llm/complete` endpoint).
- **`portal/state.py`** — `PuffoCoreConfig.transport` (`"native"` default | `"bridge"`) + `sandbox_token`, validated fail-fast in `AgentConfig.load`; `save()` pops both keys for native agents so existing agent.yml files stay byte-identical.
- **`agent/puffo_core_client.py`** — keyword-only `bridge_client=` injection seam; `listen()` branches to a bridge lifecycle loop (connect / heartbeat / reconnect with the native backoff schedule). Frames are logged and dropped; consumer/invite/warm tasks are skipped in bridge mode (they drive signed HTTP endpoints that can't work keyless yet).
- **`portal/worker.py`** — builds + injects `CloudBridgeClient` when `transport == "bridge"`, skipping the identity-file import. Native path: zero diff.
- **`mcp/puffo_core_server.py`** (19 changed lines; kept surgical for the PR #126 overlap) — inert keystore under `transport="bridge"` raising a clear `RuntimeError` instead of `FileNotFoundError`; `_cfg_from_env` reads `PUFFO_CORE_TRANSPORT`.
- **`mcp/config.py`** — additive `transport=""` param on `puffo_core_mcp_env`; no callers updated, default output byte-identical.
[...]

PR #136, opened 2026-07-07
T23 phase 2: fat-cloud message flow over the keyless bridge (drop encrypt/decrypt at seam)

Glimmer0x:
## What (stacks on #127)

**Phase 2** of the fat-cloud transport swap. Under `puffo_core.transport: "bridge"`, the message path now carries **plaintext** (server holds all crypto). The `native` (signed-crypto, client-held keys) transport is **byte-for-byte unchanged** — this is transport dispatch, not a rewrite. Crypto module deletion is phase 4.

## The two seam points

- **IN** (`agent/puffo_core_client.py`): `_listen_bridge` drives **one** `send_fetch_pending()` per connect (cold-sandbox backfill), and routes inbound plaintext `message` frames via a new `_payload_from_bridge_frame` → `_handle_plaintext_payload` — the latter **extracted behavior-preservingly** from the native post-`decrypt_message` body, so native still decrypts and calls the same handler. `pending_delivered` terminates backfill.
- **OUT** (`mcp/puffo_core_tools.py` + `send_fallback_message`): send plaintext via `CloudBridgeClient.send_send` instead of `encrypt_message*`, mapping the ack to the existing return shape. Dispatch threaded **only** at the in-process ws-local site (`portal/ws_local/route.py`); the subprocess/RPC MCP path stays native (`bridge_client=None`) — a subprocess can't own the daemon's single per-agent bridge WS.

## Scope (design §2 = DM + channel text)
Threads (`root_id`) sent top-level, attachments **hard-refused** on bridge, name-enrichment degrades offline (helpers return empty rather than crash) — all **phase 3**. No `crypto/` module deleted (**phase 4**).

## Verify
LLM-free + E2B-free pytest (`tests/test_cloud_bridge_msgflow.py`): inbound plaintext lands in `message_store` identically to native (no `decrypt_message`); `send_message` uses bridge plaintext + no `encrypt_message*`; connect drives exactly one `send_fetch_pending`; **native transport still encrypts/decrypts** (asserted). Full suite green.

**Stacked on `fleet/fat-cloud-phase1` (#127). HOLD — do not self-merge.** Rebases onto the merged result when #127 lands.

### DIFF
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -19,6 +19,7 @@ from ..crypto.http_client import HttpError, PuffoCoreHttpClient
 from ..crypto.keystore import KeyStore, decode_secret
 from ..crypto.message import (
     EncryptInput,
+    MessagePayload,
     RecipientDevice,
     decrypt_message,
     encrypt_message,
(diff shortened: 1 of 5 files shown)


Glimmer0x:
Superseded by #142 (the complete consolidated fat cloud agent). Closing to reduce review surface; branch retained.

PR #160, opened 2026-07-10
cloud-agent: make chat-local tool-capable (send_message via tool_use, not text) — HELD

Glimmer0x:

Fixes the fat **cloud agent** leaking `send_message(...)` as raw text (with space-ids/debug) into channels instead of actually sending — Shan's #159 Part 2 / "Problem 3". Root cause: the `chat-local` runtime is a text-only completion path — `AnthropicProvider.complete()` sent no `tools=` and never parsed `tool_use`, and the `worker.py` chat-local branch wired no MCP dispatch, so the model's `send_message(...)` call arrived as prose and the core turn-router posted it via the raw-text fallback. This makes `chat-local` genuinely tool-capable (the `SDKAdapter` contract), so `send_message` is invoked as a real `tool_use`.

### Change (3 parts, mirrors `SDKAdapter`; native/desktop untouched)
1. **`agent/providers/anthropic_provider.py`** — `complete()` takes optional `tools=` + a `dispatch(name,input)` callback, runs an agentic loop while `stop_reason == "tool_use"` (executes each block, appends `tool_result`, `is_error` on handler failure, capped at 8 iterations), iterates `content` by block type (no blind `[0].text`), sums usage, returns a `CompletionResult` (reply, tokens, executed `tool_calls`, `assistant_text_parts`). OpenAI's legacy 3-tuple contract is preserved.
2. **`agent/adapters/chat_only.py`** — `ChatOnlyAdapter` gains a worker-injected `tool_dispatch`; when set + provider `supports_tools`, it advertises `send_message`/`send_message_with_attachments`, bridges the async puffo_core handlers via `run_coroutine_threadsafe` (120s timeout; `complete()` in `asyncio.to_thread`), and populates `TurnResult.metadata` (`send_message_targets`/`assistant_text_parts`/`tool_names`) exactly like `SDKAdapter` so `core.py` skips the raw-text fallback. Failed sends aren't counted as targets (fallback stays available).
3. **`portal/worker.py`** — `_build_chat_local_tool_dispatch(client)` (mirrors sdk-local's `mcp_servers_override`, reuses `build_dispatch`/`PuffoCoreToolsConfig`), injected in `Worker._run()` after the live puffo_core client exists.

### Verified
- New `tests/test_chat_local_tool_use.py` (6 tests): tool_use parsed+dispatched+`metadata` populated + wire shape (`tools=` on call 1, `tool_result` w/ matching id on call 2); **end-to-end through `handle_message` asserting the fallback is NOT taken**; dispatch-error → `is_error` tool_result + no target; plain-text still works; legacy no-dispatch + OpenAI tuple paths unchanged.
- `pytest --ignore=tests/test_host_credentials.py` → **2006 passed, 4 skipped** (exit 0).
- `git diff --stat`: only the 3 modules + new test; `sdk.py`/`local_cli.py`/`docker_cli.py`/`core.py` byte-identical.

### Notes / follow-ups
- Still depends on the **LiteLLM gateway passing through Anthropic `tool_use`** (same as any tool path) — a Shan coordination point to confirm live.
- The separate **reply-once-then-silent** bug (T27) is NOT in this PR (under investigation; will be its own fix).
- Chosen over baking the claude-CLI (`cli-local`): this fixes the runtime the agent actually runs today, no real-cloud template rebuild needed.

### DIFF
--- a/src/puffo_agent/agent/adapters/chat_only.py
+++ b/src/puffo_agent/agent/adapters/chat_only.py
@@ -1,30 +1,227 @@
 """Chat-only adapter.
 
 Wraps the message-completion providers (Anthropic/OpenAI) so existing
-agents keep working without the SDK or CLI runtimes. Does not run
-tools, does not touch the filesystem, ignores workspace/claude dirs.
+agents keep working without the SDK or CLI runtimes. Does not touch
+the filesystem and ignores workspace/claude dirs, but can run the
+puffo_core send tools when the worker injects ``tool_dispatch`` and
+the provider supports structured tool use (AnthropicProvider) — the
+model's ``send_message(...)`` then posts for real instead of leaking
+as plain text through the core turn-router's fallback.
 """
 
 from __future__ import annotations
 
 import asyncio
+import logging
 
 from .base import Adapter, TurnContext, TurnResult
+
+logger = logging.getLogger(__name__)
+
+# Ceiling on one in-turn tool execution; a hung network call surfaces
+# as an is_error tool_result instead of wedging the turn thread.
+_TOOL_CALL_TIMEOUT_SECONDS = 120.0
+
+# Tool names whose successful call means "the reply was posted via
+# puffo_core" — the core turn-router skips its raw-text fallback when
+# metadata carries a target for one of these.
+_SEND_TOOL_NAMES = ("send_message", "send_message_with_attachments")
+
+# Anthropic tool schemas for the puffo_core send tools, mirroring the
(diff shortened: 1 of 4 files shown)


shanjing:
Superseded — the chat-local structured `send_message` tool_use this PR adds already shipped via #176 and is present in the cloud-agent baseline (`chat_only.py`: `_SEND_TOOL_NAMES`/`_make_dispatch`, covered by `test_chat_local_tool_use.py`). This branch also targets a stale base, so merging it would revert the cloud-slim split work. Closing as done.

PR #161, opened 2026-07-10
cloud-agent: fix reply-once (thread-dedup cursor drops same-ms envelopes) — HELD

Glimmer0x:

Fixes the cloud agent **replying once to a DM/thread then going silent** (Shan, 07-10 / "Problem 2" / T27).

### Root cause
`puffo_core_client.py` `handle_envelope` rejected any inbound with `payload.sent_at <= last_processed_sent_at` (a per-`root_id` persisted cursor, meant only to drop cross-restart *redelivered* messages). `sent_at` is epoch-**milliseconds** and **not unique**: the server stamps consecutive bridge `Send` frames with its own `now_ms`, so a programmatic burst (e.g. agent-to-agent traffic, or two quick messages) can put two *distinct* envelopes on one thread root in the **same millisecond**. A DM conversation / thread has a stable `root_id`, so after the first message is processed (`mark_thread_processed(root, msg.sent_at)`), a genuinely-new second message sharing that ms hit `<=` and was dropped as a "duplicate" → no second turn.

### Fix
Key the thread-dedup cursor on the **`envelope_id` at the `sent_at` watermark**, not the bare timestamp:
- **`message_store.py`** — `thread_processing_state` now also remembers the envelope_id(s) processed *at* the high-water `sent_at`; `get_last_processed_sent_at`/`mark_thread_processed` extended (backward-compatible; migration/default). Reject only a message whose `sent_at < watermark`, OR `sent_at == watermark` **and** its `envelope_id` was already processed.
- **`puffo_core_client.py`** — the cursor check uses that boundary set, so a NEW envelope at the same ms is admitted while a genuinely-redelivered one is still dropped. (Not a bare `<=`→`<`, which would regress cross-restart dedup.)

### Verified
- New `tests/test_reply_once_cursor.py` (6 tests): a distinct second envelope at the SAME `sent_at` on one root IS admitted (was dropped); a true redelivery (same envelope_id + sent_at) is STILL dropped; DM + thread cases.
- Full gate `pytest --ignore=tests/test_host_credentials.py` → **exit 0** (independently re-run by the coordinator; only the 4 standard PySide6/platform skips).

Coordinator note: I traced the root cause directly (`handle_envelope`→cursor); a focused agent implemented + tested; I re-ran the gate before opening this.

### DIFF
--- a/src/puffo_agent/agent/message_store.py
+++ b/src/puffo_agent/agent/message_store.py
@@ -41,13 +41,20 @@ CREATE INDEX IF NOT EXISTS idx_messages_thread_root
 -- Per-thread cursor used by the thread-batched priority queue. After
 -- ``on_message_batch`` finishes successfully, the consumer advances
 -- this to the ``sent_at`` of the last message in the dispatched
--- batch. The listen handler then drops any inbound message whose
--- ``sent_at`` is <= the stored cursor, so server-side pending-message
--- redeliveries after a daemon restart don't re-trigger the agent on
--- already-processed threads.
+-- batch, plus the envelope_ids of the batch messages that share that
+-- exact ``sent_at`` (JSON array). The listen handler then drops any
+-- inbound message whose ``sent_at`` is strictly below the cursor, or
+-- equal to it with an envelope_id already recorded at the watermark —
+-- so server-side pending-message redeliveries after a daemon restart
+-- don't re-trigger the agent, while a genuinely-NEW message that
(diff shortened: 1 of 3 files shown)


Glimmer0x:
Closing as superseded by the Agent Foundation Global Inbox architecture. Current main no longer uses the per-thread sent_at cursor or the old reply-once admission path changed here; carrying this conflicting dev-targeted patch forward would reintroduce an obsolete state model.

PR #162, opened 2026-07-10
cloud-agent: handle added_to_space push → eager spaces refresh (pairs w/ server #202) — HELD

Glimmer0x:

The **puffo-agent side** of the space-join notify (pairs with puffo-server **#202**). The server now pushes `{"type":"added_to_space","space_id":…}` (`AgentServerMsg::AddedToSpace`) over the bridge control WS the moment a cloud agent is auto-added to a Space; this makes the agent react instead of only finding out lazily on its next `list_spaces`.

### Change
- **`puffo_core_client.py`** — new `added_to_space` branch in `_dispatch_bridge_frame` (the inbound-bridge-frame router, alongside `message`/`pending_delivered`). It schedules `_refresh_bridge_spaces(space_id)` fire-and-forget (same pattern as the ack tasks — awaiting inline would deadlock `frames()`, which must keep receiving to resolve the `spaces` reply).
- **`_refresh_bridge_spaces`** — re-issues `list_spaces` via the existing keyless `bridge.send_list_spaces()` primitive, then warms `_space_name_cache` + persists via `disk_cache.persist_space`. Idempotent (duplicate push = harmless re-list), fail-soft (a failed refresh logs a warning, never crashes the listen loop), defensive on entry keys (`space_id`/`id`).

### Verified
- 4 new tests in `test_cloud_bridge_msgflow.py` (frame → exactly one `list_spaces` + caches warmed; duplicate harmless; missing-id/failed-refresh survive; end-to-end over the real `_listen_bridge`). Module: 35 passed.
- Full gate `pytest --ignore=tests/test_host_credentials.py` → **2004 passed, 4 skipped** (exit 0).

Completes the T26 loop end-to-end (server #202 emits → agent refreshes). Bridge transport only; native/desktop signed-HTTP path untouched.

### DIFF
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -494,10 +494,11 @@ class PuffoCoreMessageClient:
         # non-None, listen() runs the bridge lifecycle loop instead of
         # the signed WS client — no identity file is ever loaded.
         self._bridge = bridge_client
-        # In-flight ack tasks scheduled off the bridge dispatch loop
-        # (F1). Held so the tasks aren't GC'd mid-flight; each removes
+        # In-flight fire-and-forget tasks scheduled off the bridge
+        # dispatch loop (F1 acks + added_to_space spaces refreshes).
+        # Held so the tasks aren't GC'd mid-flight; each removes
         # itself via a done-callback. Native transport never populates
-        # this — the ack lives only in the bridge dispatcher.
+        # this — both producers live only in the bridge dispatcher.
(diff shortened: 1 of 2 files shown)


shanjing:
Superseded by #187 — same change rebased onto `fleet/cloud-agent-agent-all` (this PR's base `dev` is ~13 commits stale, and the cloud E2B templates build from `fleet`, so as-was it would never reach running cloud agents). Rebase was clean (no conflicts with the merged split/DM work), tests green. Closing in favor of #187.

PR #172, opened 2026-07-14
feat(codex): route codex harness to the LiteLLM gateway via per-agent VK

shanjing:
Lets a **codex** agent authenticate to the **LiteLLM gateway** with its per-agent **virtual key** instead of ChatGPT OAuth — so cloud/E2B agents can run codex without a real provider key ever entering the sandbox (**Posture B**).

Pairs with **cloud-infra #78** (gateway serves codex; AIM derives the runtime and mints the VK).

**Live and proven:** agent `option-iii` on E2B answers real DMs through the gateway.

## How it works — no fork of codex

We use codex's **own custom-provider config**. `mcp/config.py` now emits, into the per-agent `$CODEX_HOME/config.toml`:

```toml
model_provider = "litellm"

[model_providers.litellm]
base_url = "https://litellm-staging.puffo.ai/v1"
env_key  = "OPENAI_API_KEY"     # holds the VK
wire_api = "responses"
```

`local_cli.py` builds that block from `runtime.llm_base_url` + `runtime.api_key` (both set by AIM in `agent.yml` at create), injects the VK into the codex subprocess env, and **bypasses the ChatGPT-OAuth gate** in gateway mode — that gate would otherwise raise *"codex needs auth — run `codex login`"*, which is impossible headless.

`wire_api = "responses"` is **required**, not a preference: codex speaks only the Responses API since OpenAI removed Chat Completions (Feb 2026).

## The credential-refresh fix (the ~40s/turn bug)

`portal/credential_refresh.py` is **correct** — for OAuth agents. The gap was that the new VK/gateway mode never opted out of it. A gateway agent has **no OAuth token to refresh**, but the daemon was still:
1. registering it with the periodic refresher, and
2. handing the worker a pre-delivery `ensure_fresh` gate.

Both probed `api.openai.com` directly, which **401s under VK auth** → the batch was deferred (~40s + a transient "API Error" flash in the UI) and `refresh_broken` was set. Every turn.

`daemon.py` now no-ops both when `runtime.llm_base_url` is set. **Native-auth codex agents keep the refresh behaviour unchanged.**

**Measured live on E2B: turn latency 48s → 20s**, no error flash. The remaining 20s is model time.

## Tests

- `test_codex_gateway_skips_refresh.py` (new, 7 tests) — gateway mode skips both the refresher registration and the `ensure_fresh` gate; OAuth codex still registers with the codex refresher; non-codex still uses the Claude refresher; a runtime **missing** `llm_base_url` degrades to OAuth mode rather than crashing worker startup.
- `test_codex_config.py` (+3) — the emitted provider block, its absence when unconfigured, and coexistence with the `puffo` MCP server entry.

**Full suite: 2029 passed / 7 failed — identical 7 failures on the clean base** (`test_host_credentials.py` reads the developer's real `~/.claude/.credentials.json` instead of a tmp dir). **Zero regressions.** That test-isolation bug is pre-existing and worth its own ticket.

## Reviewer note

`worker.py` in the vendored FAT tree also carries `_build_chat_local_tool_dispatch` / `ChatOnlyAdapter` wiring that is **not on this base and not mine** — deliberately **excluded** from this PR. If that belongs on this branch it needs its own PR from its author.

### DIFF
--- a/src/puffo_agent/agent/adapters/local_cli.py
+++ b/src/puffo_agent/agent/adapters/local_cli.py
@@ -372,6 +372,23 @@ class LocalCLIAdapter(Adapter):
         # sync_host_mcp_servers.
         merged_extras: dict[str, dict] = dict(self._desired_codex_extras)
         merged_extras.update(host_mcps)
+        # LiteLLM gateway path: when the agent carries a virtual key + gateway
+        # URL (set by AIM in agent.yml), codex authenticates to the gateway with
+        # the VK via the Responses API instead of ChatGPT OAuth. base_url must
+        # end in /v1 (codex POSTs to {base_url}/responses).
+        gateway_provider = None
+        if self.llm_base_url and self.llm_api_key:
+            base = self.llm_base_url.rstrip("/")
+            if not base.endswith("/v1"):
+                base = base + "/v1"
+            gateway_provider = {
+                "name": "litellm",
+                "display_name": "LiteLLM gateway",
+                "base_url": base,
+                "env_key": "OPENAI_API_KEY",
+                "model": self.model or "codex",
+                "wire_api": "responses",
+            }
         if self.puffo_core_mcp_env:
             write_codex_mcp_config(
                 codex_home / "config.toml",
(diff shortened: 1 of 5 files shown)


shanjing:
Tested in beta9 live — bundled into the beta9 fat template and validated end-to-end on staging (agents boot, reply to humans, and now hold agent↔agent chat with zero decrypt failures).

shanjing:
**Problem solved:** meters codex LLM calls via per-agent gateway key (off unmetered subscription).

PR #174, opened 2026-07-14
feat(cloud-agent): route keyless read tools to /v2/cloud-agents/* (agent half of dropping SubkeyOrSandboxTokenAuth)

shanjing:
The **agent half** of the `SubkeyOrSandboxTokenAuth` consolidation. Pairs with the merged puffo-server side (**#222** + the existing `/v2/cloud-agents/*` read routes).

## Problem

A keyless bridge (cloud) agent has **no subkey to sign with**, so the signed read routes never worked for it:

`/spaces` · `/spaces/{id}/channels` · `/spaces/{id}/members` · `/spaces/{id}/channels/{id}/members` · `/identities/profiles`

Result today: sender names render as raw slugs, and `list_channel_members` / `get_user_info` / `list_channels` / `list_spaces` are effectively dead for cloud agents. The code says as much — *"signed HTTP endpoints that can't work keyless; names render as ids until those are rewired."*

## Fix — one routing shim, not 15 call-site edits

`PuffoCoreHttpClient.get()` gains a keyless branch: a read route that has a `/v2/cloud-agents/*` twin is transparently rewritten to that twin and sent **unsigned** — the E2B egress proxy injects `x-sandbox-token` (same mechanism the existing keyless send/blob paths use). Every `self.http.get(...)` call site stays byte-for-byte unchanged, and **native (signed) agents leave `keyless=False` and never hit the branch.**

```python
async def get(self, path):
    if self.keyless:
        twin = cloud_agent_read_twin(path)   # None unless the path has a twin
        if twin is not None:
            return await self.get_unsigned(twin)
    ...signed path...
```

`keyless=(transport == "bridge")` is already wired at both hot construction sites (`worker.py`, `mcp/puffo_core_server.py`), so the shim fires exactly for cloud agents.

## The one subtlety — the events routes

`/spaces/events` and `/spaces/{id}/events` have **no** `/v2/cloud-agents` twin. The matcher's patterns are anchored so those deliberately **don't** rewrite — they stay signed and remain unavailable keyless, exactly as before. There's an explicit regression test for this (`test_events_routes_are_not_swept_up_by_the_spaces_matcher`).

## Tests

- `test_cloud_agent_read_twin.py` (new, 14 cases) — every twinned route rewrites (query preserved); events routes and non-read paths return `None`; no double-prefixing.
- `test_http_client.py` (+2) — a keyless `get("/spaces")` lands on `/v2/cloud-agents/spaces` **unsigned** with the egress token; a native `get("/spaces")` stays on `/spaces` **signed**.
- Related sweep: **414 passed, 0 failed.**

## Scope / follow-up

This migrates the **transport**. The on-demand read tools work immediately (they call `self.http.get` directly). A **follow-up** can un-gate the background `_warm_member_caches` / `_invite_poll_loop` tasks for bridge agents — they were skipped only because reads didn't work keyless, which is no longer true.

## Sequencing

This unblocks dropping `SubkeyOrSandboxTokenAuth` server-side. Order (do not reorder): **merge this → re-vendor + rebuild E2B template (beta7) → verify live → then drop the union extractor.** Dropping server-side before this ships would strand every cloud agent's reads.

### DIFF
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 from typing import Any
 
 import aiohttp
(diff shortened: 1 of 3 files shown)


shanjing:
Closing — no longer needed.

PR #175, opened 2026-07-15
feat(cloud-agent): keyless-identity fix — skip signed load_identity for bridge agents

shanjing:
**Draft for review** — not for merge until approved. Independent of the chat-local tools change (that's a separate PR); no shared files.

## Problem — `identity not found` for keyless bridge agents
Keyless (bridge-transport) cloud agents authenticate with an egress-injected `x-sandbox-token` and hold **no local signing identity** by design. But two subsystems made **signed** HTTP calls, and the signed path chains `post()` → `_ensure_subkey()` → `_rotate_subkey()` → `keystore.load_identity()`, which raises `identity not found: <slug>` for a keyless agent:

1. **`status_reporter.py`** — heartbeat / `begin_turn` / `end_turn` on **every turn**. Non-fatal for messaging (the bridge WS is a separate path), but it produced constant errors and made LLM agents **decline cross-agent operations** ("the Puffo identity lookup is failing").
2. **`profile_sync.py`** — `sync_agent_profile()` issued a signed `PATCH /identities/self` on every warm/startup sync.

## Fix
Guard both on the **existing** keyless signal (`http.keyless`, derived from `transport == "bridge"`) and skip the signed call — bridge agents have no route for it anyway:
- `status_reporter.py` — read `self._keyless` once; short-circuit all wire calls (heartbeat / begin_turn / end_turn / end_turn_batch / report_error / _send_heartbeat), keeping local status bookkeeping. Liveness for bridge agents comes from the bridge WS, so the skip is a clean no-op.
- `profile_sync.py` — `if pc.transport == "bridge": return` at the single choke point every sync caller reaches.

**Hard constraints held:** `crypto/keystore.py` untouched (its error is correct for native agents); native agents byte-for-byte unchanged (`getattr(..., False)` + `transport == "bridge"` guards; the 22 pre-existing `test_status_reporter.py` tests pass unchanged).

## Why other `load_identity` sites are safe
`puffo_core_client.py:619/708` are inside the native `listen()` decrypt path, which `_listen_bridge` returns before reaching. `import_agents.py` / `cli.py` / `daemon.py` are native-agent management paths, not in a bridge agent's runtime loop. Inbound-enrichment reads already degrade to ids on failure.

## Review gate
Built with an adversarial review pass (find issues → independently verify each → fix only confirmed). The `profile_sync` path was **found by review, verified real, and fixed** (it was a second keyless path missed on the first pass); a low-value dead-code finding was verified as a non-issue and correctly not "fixed."

## Tests
`status_report / keyless / bridge / identity` → **198 passed, 1 skipped (pre-existing)**. New tests added in `test_status_reporter.py` and `test_agent_sync.py` (incl. an exploding-keyless-http fake that fails loudly if any signed POST leaks).

### DIFF
--- a/src/puffo_agent/agent/status_reporter.py
+++ b/src/puffo_agent/agent/status_reporter.py
@@ -45,6 +45,19 @@ class StatusReporter:
         runtime_provider: Optional[Callable[[], dict[str, Any]]] = None,
     ) -> None:
         self._http = http
+        # Keyless (T23 bridge) transport: the agent authenticates with an
+        # egress-injected ``x-sandbox-token`` and holds NO local signing
+        # identity. Every status wire call here goes through the *signed*
+        # ``PuffoCoreHttpClient.post`` path (``_ensure_subkey`` →
+        # ``load_session`` → ``_rotate_subkey`` → ``load_identity``), which
+        # raises "identity not found: <slug>" for a keyless agent — on every
+        # heartbeat, ``begin_turn`` and ``end_turn``. Reading the flag off the
+        # http client (the SAME signal the tools + worker use;
+        # ``getattr(..., False)`` keeps test fakes and native agents at False)
+        # lets us short-circuit those calls so no ``load_identity`` is ever
+        # attempted. Server liveness for bridge agents is tracked via the
+        # bridge WS, not this heartbeat, so the skip is a clean no-op.
+        self._keyless = bool(getattr(http, "keyless", False))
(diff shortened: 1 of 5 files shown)


shanjing:
Tested in beta9 live — bundled into the beta9 fat template and validated end-to-end on staging (agents boot, reply to humans, and now hold agent↔agent chat with zero decrypt failures).

shanjing:
**Problem solved:** lets keyless cloud agents boot without signing keys.

PR #186, opened 2026-07-17
fix(reply): complete DM top-level fallback across all reply paths (supersedes #159)

shanjing:
Supersedes #159 — completes the DM-top-level fix so it covers **every** reply path, including the keyless/bridge path our cloud agents actually use.

## Problem
An agent's reply in a **DM** was threaded (nested under the incoming message), so clients collapsed it behind an "N replies" badge — the agent looked silent even though it replied. Channels *should* thread; DMs are linear and must render inline/top-level.

## What #159 did, and the gap
#159 fixed two of the reply paths (keyless `send_message`, native `send_fallback_message`) but left the symmetric siblings still threading DMs — most importantly the **keyless/bridge `send_fallback_message`** path, which is the cloud/E2B agent fallback (the exact scenario the fix targets). It also broke an existing test that encoded the old threaded-DM behavior.

## This PR
Gates **all five** reply paths on DM-vs-channel — DM → top-level (no `thread_root_id`), channel → threaded (unchanged, byte-for-byte):
- keyless `send_message` *(from #159)*
- native `send_fallback_message` *(from #159)*
- keyless/bridge `send_fallback_message` — **the cloud-agent fix**
- native `send_message`
- native `send_message_with_attachments`

Every change is a one-line `None if envelope_kind == "dm" else <root>` guard (or the bridge DM branch set to `None`); the channel arm is the original expression untouched. No refactors.

## Tests
- Updated the two stale `*_threads_on_bridge` DM assertions to expect top-level (channel assertions unchanged).
- New `test_dm_fallback_toplevel_verify.py`: 5 paths × (DM top-level + channel threaded) = 10 tests, all passing, exercising the real functions with the repo's own mocks.
- Relevant suite: 417 tests, 0 failures. Neighbor suite: 219, 0 failures. (The 7 full-suite failures are pre-existing `test_host_credentials.py` host-state failures, identical on the clean base — 0 new.)

Scope: both cloud (keyless) and local (native) agents. Channel threading is unaffected.

### DIFF
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -4056,8 +4056,16 @@ class PuffoCoreMessageClient:
             ack = await self._bridge.send_send(
                 plaintext=text,
                 recipient_slug=recipient,
-                thread_root_id=root_id or None,
-                reply_to_id=root_id or None,
+                # DM fallback replies go top-level so they render inline in
+                # the linear DM conversation. Threading a DM reply under the
+                # incoming message buries it behind a "N replies" thread badge
+                # in the main view — the human never sees the answer inline and
+                # it reads as the agent going silent. This is the keyless/bridge
+                # path the cloud/E2B agents use; it mirrors the native DM
+                # fallback fix below and the keyless send_message DM tool.
+                # Channels keep threading (root_id) as before.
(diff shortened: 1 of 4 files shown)


shanjing:
Closing — no longer needed.

PR #188, opened 2026-07-21
fix(http): recover PuffoCoreHttpClient session after a dead connection

shanjing:
## Bug
`PuffoCoreHttpClient._get_session` reuses a cached `aiohttp.ClientSession` and rebuilds it only when `.closed`. If the pooled socket dies while the session object stays open (e.g. a long idle gap, or a paused/resumed process leaving a stale keep-alive), every subsequent request on that session fails and never recovers — nothing resets the session on a connection error, and the retry path only covers `401`.

In practice this made an agent's periodic status POSTs stop landing after the first connection drop.

## Fix
On a connection error (`aiohttp.ClientConnectionError` / `asyncio.TimeoutError`), drop the stale session and retry once on a fresh connection — both the JSON (`_do_request`) and bytes (`_do_request_bytes`) paths, via a new `_reset_session()` plus extracted `_issue_json` / `_issue_bytes` helpers.

Retries are safe for the affected endpoints (idempotent heartbeat upsert; processing-start is idempotent on its client-issued run id).

## Tests
`tests/test_http_client_resume.py` (4 new):
- stale connection (`ServerDisconnectedError`) → reset + retry succeeds, dead session torn down
- read timeout → reset + retry
- success path → session reused, not reset
- both attempts dead → error propagates after exactly one retry (no infinite loop)

22/22 pass (18 existing `test_http_client.py` + 4 new); `py_compile` clean.

### DIFF
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 from typing import Any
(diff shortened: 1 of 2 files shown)


shanjing:
Closing — this is a valid robustness fix (recover a dead aiohttp session after a connection drop) but it is NOT the cause of the empty cloud-agent Log. Root cause turned out to be an auth gap: keyless bridge agents can't sign the status write endpoints (`/agents/me/heartbeat`, `/messages/{id}/processing/*`), which have no keyless `/v2/cloud-agents/*` twin, so every status post fails with `identity not found`. Real fix tracked in the keyless-status-twin work (puffo-server + puffo-agent). This branch stays available if we want the session-recovery hardening as a standalone change later.

PR #201, opened 2026-07-27
Add durable global Python agent message runtime

Glimmer0x:
## Summary

- Replace the Python agent's per-thread dispatch path with a durable SQLite receipt inbox and one coalesced global turn scheduler.
- Add provider context admission/compaction controls and a daemon-owned `SendCoordinator` for freshness-aware channel sends, held-send recovery, and structured semantic MCP results.
- Integrate the runtime with the `dev` CloudBridge/keyless transport and proactive reload path. Bridge messages without a trusted server sequence use an explicit local-ordinal lane rather than fabricating freshness.
- Preserve native encrypted sends, bridge keyless sends, attachment handling, and provider session continuity.

## Why

The old runtime could ACK before model admission, process accumulated targets independently, and send against stale channel context. This change makes receipt ownership durable, batches pending work into coherent turns, and coordinates sends against server conversation heads.

## Integration note

This feature was implemented and verified on the current `main` lineage, while the requested `dev` base has independently diverged. The integration commit merges `dev` into the tested feature snapshot, so this PR also reconciles the required `main` and `dev` prerequisites. The merge specifically resolves CloudBridge, keyless send, credential projection, and proactive reload behavior.

## Validation

- Full Python test suite passed; two existing platform/permission tests were skipped by their own markers.
- Focused global inbox, send coordinator, held reconsideration, worker, CloudBridge, proactive reload, and MCP tool suites passed.
- `python -m compileall -q src`
- `git diff --check`


Glimmer0x:
Superseded by #203, rebuilt from the synchronized dev baseline as one clean feature commit. The old branch is retained temporarily for traceability.

PR #210, opened 2026-08-03
feat(runtime): support Codex in cli-docker

hellohanchen:
## Summary

- Support Codex in `cli-docker`, with isolated auth, skills, MCP configuration, sessions, and `/workspace` execution.
- Limit `cli-docker` to Claude Code and Codex; remove the unfinished Gemini and Hermes Docker paths.
- Resolve Docker through `PUFFO_DOCKER_BIN`, active and persistent user `PATH`, cached paths, and known Docker Desktop locations.
- Replace the desktop Hermes placeholder with Docker installation status and its resolved executable path.
- Harden Docker lifecycle handling with tri-state probes, bounded subprocesses, and child-process kill/reap.
- Sync Codex host MCP registrations and the selected portable OAuth credential for both `cli-local` and `cli-docker`.

## Codex MCP OAuth

Codex stores MCP OAuth separately from `config.toml`. Agent homes use the file-backed store, and `sync_host_mcp(name)` copies only the selected server credential into the isolated `CODEX_HOME`.

Host OAuth must also use the portable store:

```bash
codex -c 'mcp_oauth_credentials_store="file"' mcp login <name>
```

If an existing credential is encrypted in the OS keyring, sync reports the required one-time host re-login instead of claiming success. Agents do not need a separate login.

## Docker safety

- Transient probe failures no longer cause container creation, recreation, or forced removal.
- Existing compatible containers and Codex sessions are preserved.
- Docker control commands have timeouts; timed-out and cancelled children are killed and reaped.
- Host-only MCP commands are skipped for container runtimes, while portable registrations remain available.
- Literal `env_overrides` values may appear in process listings or errors; secrets are forwarded by environment-variable name only.

## Validation

- Full suite: `2243 passed, 11 skipped`
- Focused shared-primer, resolver, harness, and host-MCP tests: `106 passed`
- CLI source smoke: `version` and `agent refresh --help`
- GitHub Actions: Python 3.11, Python 3.12, Actions analysis, Python analysis, and CodeQL passed for `20d388e`
- `git diff --check` passed

## Smoke test

- Docker Desktop 29.6.2 was resolved from its per-user install outside the daemon's `PATH`.
- The PR worktree is running in background mode with the operator's normal Puffo Agent home.
- The existing Codex Docker container and saved thread resumed successfully.
- Synchronized MCP registrations, including Linear, are visible to the Codex worker.

Remaining manual checks: final desktop-card layout and a Linear tool call after the one-time file-store OAuth migration.

## Follow-ups

- Live Codex configuration refresh remains deferred; workers apply it after restart.
- Invalid `agent.yml` handling remains a separate TODO.


src/puffo_agent/portal/runtime_matrix.py
```diff
@@ -73,6 +72,18 @@
     HARNESS_CODEX:       frozenset({PROVIDER_OPENAI}),
 }
 
+HARNESSES_FOR_RUNTIME: dict[str, frozenset[str]] = {
```
hellohanchen, line 75: **This is the blocker: the table changes what's legal, but nothing migrates agents that are already on disk.**

Newly rejected combinations: `cli-docker` + `hermes`, `cli-docker` + `gemini-cli`, and — see the README comment — `cli-local` + `gemini-cli`. The first two were documented as supported.

The failure chain for an operator who upgrades with one of those on disk:

1. `validate_triple` rejects → `AgentConfig.load` raises `RuntimeError`
2. `daemon.py` catches *every* load failure with `logger.warning(...)` + `continue`
3. The agent never starts. No `RuntimeState` is written, so the portal keeps showing whatever status it last wrote — there is no operator-visible signal that anything is wrong.
4. `_load_agent_cfg_cached` only populates its cache *after* a successful load, so the parse and the warning repeat on **every reconcile tick**. `reconcile_interval_seconds` defaults to `2.0` → ~43k identical WARNING lines/day per stranded agent.

And the reason I'd call it a blocker rather than a rough edge: **the obvious repair doesn't work.** `puffo-agent agent runtime <id> --harness claude-code` calls `AgentConfig.load(agent_id)` at `cli.py:1020` *before* it applies the flag, so it exits with the same unhandled traceback. `agent show` (`cli.py:691`) too. Only `agent list` catches and degrades. The sole recovery is hand-editing `agent.yml`.

Secondary, worth fixing either way: the `continue` in the reconcile loop happens before `_worker_needs_restart`, so a worker already running when its config is edited into invalidity is **never stopped** — it keeps serving the stale config indefinitely.

This file already has the right precedent in `migrate_legacy_kind` / `_LEGACY_KIND_MIGRATIONS`. Options, cheapest first:

- Make `cmd_agent_runtime` load leniently (parse without validating) so operators can fix their own agents with the command that already exists.
- Add a harness migration mapping `cli-docker` + `hermes`/`gemini-cli` → `claude-code` with a one-time WARNING.
- At minimum, have `_reconcile_once` write an errored `RuntimeState` and log once per (agent, mtime) instead of per tick, so the failure is visible and doesn't flood.

The CHANGELOG also carries no migration note for this.

README.md
```diff
@@ -384,21 +384,30 @@ The `runtime.kind` in an agent's `agent.yml` decides where its brain runs:
...
 ```
 
 Pass `--help` for the full flag list (provider, harness, allowed_tools,
 docker_image, permission_mode, max_turns).
 
-> **codex** (`runtime.harness=codex`, `cli-local` only) spawns OpenAI's `codex
+> **codex** (`runtime.harness=codex`, `cli-local` or `cli-docker`) spawns OpenAI's `codex
 > app-server` — `codex login` once (ChatGPT-account OAuth, no API key path).
 >
+> The Docker runtime supports only `claude-code` and `codex`; Gemini and
```
hellohanchen: This sentence is wrong for Gemini, and following it walks the operator straight into the stranded state described on `runtime_matrix.py`.

> The Docker runtime supports only `claude-code` and `codex`; Gemini and Hermes remain available through their supported non-Docker runtimes.

Hermes is fine — it's still in `HARNESSES_FOR_RUNTIME[cli-local]`. **`gemini-cli` is in neither entry**, is removed from both `--harness` choice lists, and `local_cli.py:199` raises on it. So it isn't "available through non-Docker runtimes"; it's gone everywhere.

An operator reading this writes `kind: cli-local, harness: gemini-cli`, and because there's no migration path they get an agent that silently never starts plus a warning every 2 seconds.

The CHANGELOG entry gets this right — it mentions only Hermes. Worth making the README match, and deciding explicitly whether dropping `gemini-cli` from `cli-local` was intended at all: `GeminiCLIHarness`, `sync_host_gemini_skills`, `sync_host_gemini_mcp_servers` and the gemini `model_catalog` entries are all dead code now, which suggests it was — but the README says the opposite, and `resolve_effective_harness` still maps `google → gemini-cli`, so `create --runtime cli-local --provider google` fails while naming a harness the operator never typed.

src/puffo_agent/agent/cli_bin.py
```diff
@@ -240,6 +244,27 @@ def _claude_bundle_paths() -> list[Path]:
     )
 
 
+def _docker_bundle_paths() -> list[Path]:
```
hellohanchen, line 251: The bundle paths themselves are reasonable, and Windows quoting is fine — these flow as list elements into `create_subprocess_exec`, which applies `list2cmdline`, so `C:\Program Files\...` is quoted correctly. `%LOCALAPPDATA%` expansion via `os.path.expandvars` is correctly confined to the `win32` branch.

The issue is the **order** they sit in. `_resolve` consults, in sequence:

1. `$PUFFO_DOCKER_BIN` — accepted on `is_file()` alone
2. in-memory cache
3. **on-disk `~/.puffo-agent/resolved_clis.json`**
4. `shutil.which(name)` — process `PATH`
5. `shutil.which(name, path=_real_path())` — reconstructed login-shell PATH
6. these bundle paths

Step 3 precedes `$PATH`, and the only validation anywhere is `Path(p).is_file()` — nothing checks the target is executable, is owned by root, or lives in a trusted directory.

So an ordinary user-writable JSON file is the authority for which binary the daemon executes as `docker`, `claude`, and `codex`, persisted across restarts. The resolver predates this PR, but this PR is what routes **docker** through it — and docker-group access is root-equivalent on most hosts, which makes it the highest-value entry in that file.

Why it's not merely theoretical here: `cli-local` agents run as the operator's UID with full filesystem access — that's the documented model. Model output that writes `{"docker": "/tmp/x"}` to `~/.puffo-agent/resolved_clis.json` redirects every subsequent docker spawn, for every agent on the host, to an attacker-chosen binary with the daemon's privileges. (`cli-docker` containers can't reach the file — only `agent_dir(agent_id)` and `shared/` are mounted — so the vector is cli-local agents and anything else running as that user.)

Two cheap mitigations, either of which closes it: move the disk cache *after* `shutil.which`, so a real PATH entry always wins over a cached one; and add `os.access(p, os.X_OK)` to the revalidation.

Related, same file: `_login_shell_path` runs `$SHELL -ilc` from the daemon — arbitrary code execution driven by an env var — with `_run_capture` swallowing everything in a bare `except Exception`. Pre-existing and probably intentional, but worth a comment saying so.

src/puffo_agent/agent/adapters/docker_cli.py
```diff
@@ -958,52 +668,57 @@ async def _ensure_started(self) -> None:
...
+                )
+            except OSError:
+                layout_current = False
+            package_current = (
+                await self._puffo_pkg_mount_is_current() if existed else True
+            )
+            harness_current = (
+                await self._container_harness_is_current() if existed else True
+            )
+            if existed and not (layout_current and package_current and harness_current):
```
hellohanchen, line 726: `docker rm -f` on a heuristic composed of three independently-computed booleans, any of which can be false because Docker was briefly unreachable rather than because anything is actually stale. See the comment on the probes below for why.

**Damage assessment — and I want to be accurate rather than alarming here.** I traced `_start_container` and everything durable is a host bind-mount: `workspace_dir:/workspace`, `claude_home_src:/home/agent/.claude`, `codex_home:/home/agent/.codex` (which is where `codex_session.json` lives), `agent_claude_json`, `shared_fs_dir`, and the whole `agent_home_dir:/home/agent/.puffo-agent-state`. Session ids, credentials, MCP config, keystore and message DB are all host-side and survive `rm -f`. **This is not data loss.**

What is destroyed is the container-local filesystem outside those mounts — anything the agent installed at runtime (`npm i -g`, `pip install`, `apt`), `~/.cache`, `~/.config`, `~/.bashrc`, plus a multi-minute image rebuild if `_ensure_image` also has to run. Recoverable, but a real cost paid for an infrastructure hiccup.

The safety property is the important part and it's asserted nowhere. A single test — write a sentinel into `workspace_dir` and `codex_home`, force the stale branch, assert both survive and that `rm -f` was issued — locks in the invariant that makes this design acceptable. Right now a future mount refactor breaks it silently.

Separately, one staleness check can't detect what it exists for: `_puffo_pkg_mount_is_current` runs `test -f /opt/puffoagent-pkg/puffo_agent/__init__.py` inside the container, deliberately avoiding a `Mount.Source` comparison because Docker Desktop rewrites Windows paths. But that makes it blind to the *likeliest* staleness — the daemon reinstalled into a different venv. The old container still mounts the old directory, that directory still exists, `test -f` succeeds, and the container keeps importing the previous release's `puffo_agent` against a newer daemon, indefinitely and silently. Writing `puffo_agent.__version__` (or a build id) into the layout marker and comparing it via `python3 -c 'import puffo_agent; ...'` would actually cover it.

hellohanchen (commented):
## Review: Codex in cli-docker

Reviewed against merge-base `08fbc86`. The additive half of this PR is in good shape; the subtractive half — removing the Gemini/Hermes Docker paths — has no migration story, and that's the one thing I'd fix before merging.

### What holds up well

**The secret-handling claim is true, and I checked every subprocess site.** `_ensure_codex_session` emits `-e NAME` with **no value**, so Docker pulls it from the daemon's own environment and nothing lands on argv or in `ps`. `_prepare_codex_config` collects only variable *names*; the unset-var warning logs the name, never a value. `write_codex_mcp_config` writes `bearer_token_env_var = "<name>"`. Auth sync returns closed-vocabulary mode strings, and the refresh token is scrubbed to `""`. No `shell=True` anywhere in `src/puffo_agent/agent/`.

**The tests are better than most.** `test_docker_codex.py` asserts literal argv lists, real generated `config.toml` / `auth.json` contents, and real mount strings — including negative assertions (`assert "host_only" not in config`, `assert "/.gemini" not in ...`, `assert "super-secret" not in argv`). `test_codex_session.py` spawns a real subprocess running a scripted JSON-RPC fake that asserts on inbound params. CI runs everything with no `-k`, no markers, and no skip guards in the new files — so the coverage is real coverage.

**The 213 deleted lines in `test_harness.py` are a legitimate deletion.** All 16 removed tests target `_gemini_model_id` / `_parse_gemini_reply` / `_build_gemini_argv`, none of which exist any more. No surviving behavior lost coverage. Worth noting the one invariant they carried — "a provider secret reaches the container via `docker exec -e`, never the host env" — is preserved by `test_codex_session_forwards_only_available_bearer_env` with stronger semantics.

Also: `cmd_agent_create`'s new harness resolution fixes a real pre-existing bug where `harness=claude-code` + `provider=openai` was written for `cli-local` and then rejected on the very next start.

### 1. Existing Gemini/Hermes Docker agents are stranded, and the obvious repair command doesn't work

This is the blocker. `validate_triple` now rejects the combination, `AgentConfig.load` raises, and `daemon.py` catches every load failure with `logger.warning(...)` + `continue`. So the agent never starts, no `RuntimeState` is written (the portal keeps showing stale status), and the warning re-emits on every reconcile tick — `reconcile_interval_seconds` defaults to `2.0`, so roughly 43k identical WARNING lines a day per stranded agent.

What makes it a blocker rather than a rough edge: **`puffo-agent agent runtime <id> --harness claude-code`, the natural fix, calls `AgentConfig.load` at `cli.py:1020` before it applies the flag**, so it dies with the same unhandled `RuntimeError`. `agent show` too. Only `agent list` degrades gracefully. The sole recovery is hand-editing `agent.yml`.

Secondary: the `continue` happens before `_worker_needs_restart`, so a worker already running when its config becomes invalid is never stopped — it keeps serving the old config indefinitely.

The file already has the right precedent in `migrate_legacy_kind` / `_LEGACY_KIND_MIGRATIONS`. At minimum, make `cmd_agent_runtime` load leniently so operators can fix their own agents.

### 2. `gemini-cli` is rejected on **every** runtime, but the README says otherwise

`README.md:408` says *"Gemini and Hermes remain available through their supported non-Docker runtimes."* `gemini-cli` appears in **neither** entry of `HARNESSES_FOR_RUNTIME`, is gone from both `--harness` choice lists, and `local_cli.py:199` raises on it. An operator following that sentence writes `cli-local` + `gemini-cli` and lands straight in finding 1. The CHANGELOG is correct (mentions only Hermes); the README isn't.

### 3. Docker resolution: the disk cache is consulted before `$PATH`

Order in `_resolve` is: `$PUFFO_DOCKER_BIN` → in-memory cache → **`~/.puffo-agent/resolved_clis.json`** → `shutil.which` → reconstructed PATH → bundle paths. The only validation at any step is `Path(p).is_file()` — no exec check, no ownership or trusted-directory check.

So a plain user-writable JSON file decides which binary the daemon runs as `docker`. The resolver is pre-existing, but this PR is what routes `docker` through it, and docker-group access is root-equivalent on most hosts. Since `cli-local` agents run as the operator's UID with full filesystem access by design, "model output writes that file → next docker spawn executes an attacker-chosen binary" is a chain inside this product's own threat model. Moving the disk cache after `shutil.which` and adding `os.access(p, os.X_OK)` costs nothing.

### 4. A transient Docker failure is read as "container is stale" and triggers `docker rm -f`

Both staleness probes return `rc == 0` and run with `check=False`, so *any* non-zero exit means "stale" — including a daemon that's unreachable, a container that's restarting, or Docker Desktop mid-reconnect. Docker Desktop auto-updating during a reconcile is enough to hard-remove the container.

**The damage is bounded, and I want to be accurate about that**: everything durable is a host bind-mount — `/workspace`, `.claude`, `.codex` (which is where `codex_session.json` lives), `.claude.json`, `.shared`, and the whole agent state dir. Sessions, credentials, MCP config and the message DB all survive. What's lost is the container-local filesystem: anything installed at runtime, `~/.cache`, `~/.config`, plus a multi-minute image rebuild. So not data loss — but a destructive action on an infrastructure hiccup, and the invariant that makes it safe is asserted nowhere.

### 5. Smaller things

- **`refresh(host_sync=True)` is a silent no-op for Codex agents.** `worker.py` skips host sync for codex and logs *"delegated to Codex adapter"* — but on cli-docker the skill sync only runs inside `_ensure_started()`, which returns immediately once started, and `reload()` re-runs only the MCP config. On cli-local `_verify()` returns early for codex. Operator edits skills, hits refresh, sees a success-shaped log line, nothing happens until restart.
- **The reverse cwd bug is unguarded.** The rotation is gated on `thread_cwd is not None`, and `local_cli.py` never passes `thread_cwd` — while both runtimes share the same session file path. Switching an agent from cli-docker back to cli-local resumes a thread whose cwd is `/workspace`, which doesn't exist on the host. Comparing the *effective* cwd (`thread_cwd or cwd`) unconditionally would close it.
- **`resolve_effective_harness` can return a harness `validate_triple` rejects** — `google → gemini-cli` still maps, so `create --runtime cli-local --provider google` fails naming a harness the operator never typed and that `--harness` won't accept.
- **`--sandbox` help is now wrong for the runtime it was just extended to** — it still warns that `workspace-write` is downgraded on Windows, which is false inside a Linux container and pushes operators to `danger-full-access` for no reason.
- **Mojibake** in the new `codex_session.py` log line (`鈫?` — a UTF-8 arrow round-tripped through cp936). The sibling message two lines up has a correct `→`.

### 6. Test gaps that map to the two riskiest paths

Both of the above blockers are untested:

- **No test loads an existing `cli-docker` + `gemini-cli`/`hermes` `agent.yml`.** What's tested is `validate_triple` as a pure function and the `DockerCLIAdapter` constructor gate — never the middle. (`test_cli_agent_create.py:24` is also misnamed: it uses `--harness claude-code --provider openai`, so it trips the pre-existing *provider* gate, not the new runtime/harness one.)
- **No test covers the destructive rebuild.** Grepping for `CONTAINER_LAYOUT_VERSION`, `_container_state`, `.docker-layout`, `recreating stale` across `tests/` finds only two comment lines. Nothing asserts the three staleness predicates, that `rm -f` is issued, or — most valuable — that workspace and `codex_session.json` survive it. That last one is a single test with a sentinel file, and it's exactly the invariant a future mount refactor would break silently.

### Verdict

The Codex integration itself is solid work with unusually good tests. What I'd want before merge: a migration or a lenient-load path so existing Gemini/Hermes Docker agents can be repaired without editing YAML by hand, the README line corrected, and the two tests above. The resolver ordering and the probe/rebuild coupling are worth fixing but don't have to gate this PR.

(Disclosure: I run on this daemon — cli-local + claude-code, so not in the path this PR changes.)

src/puffo_agent/agent/adapters/docker_cli.py
```diff
@@ -783,22 +438,37 @@ async def _puffo_pkg_mount_is_current(self) -> bool:
...
-                "docker", "exec", self.container_name,
+                self._docker_bin, "exec", self.container_name,
                 "test", "-f",
                 "/opt/puffoagent-pkg/puffo_agent/__init__.py",
             ],
             check=False,
         )
         return rc == 0
 
+    async def _container_harness_is_current(self) -> bool:
```
hellohanchen: This probe — and its sibling above — return `rc == 0` and run with `check=False`, so there's no distinction between *"the probe answered no"* and *"the probe could not run"*.

`docker exec` exits non-zero when the container genuinely lacks the harness, but also when the daemon is unreachable, when the container is restarting, when the exec times out, or when Docker Desktop's file-sharing layer is mid-reconnect. All of those become `harness_current = False`, which the block below turns into `docker rm -f`.

Concrete: Docker Desktop auto-updates on a Windows or macOS host while the daemon is reconciling. Both probes return `rc=1`, and the container is hard-removed and rebuilt for no reason.

The fix is to make the probes tri-state — `True` / `False` / `unknown` — and treat `unknown` as "abort, don't recreate". A container that can't be probed is a container you know nothing about, which is the one case where destroying it is least justified.

`_container_state()` has the same shape: it returns `""` for both "no such container" and "docker inspect failed", and the caller acts on `""` by building an image and running `docker run --name ...`. If Docker comes back and the container still exists, the operator sees `Conflict. The container name "/puffo-<id>" is already in use` rather than "Docker is not running" — a confusing error for a simple cause. A distinct sentinel for probe failure fixes both.

Also worth a look while here: `_run_cmd` and `_build_image` both `await proc.communicate()` with no `asyncio.wait_for`, and `_ensure_started` holds `self._started_lock` across all of them. Docker Desktop wedging (blocking rather than erroring is a common failure mode there) means `docker inspect` never returns, the lock is never released, and every subsequent `run_turn` on that agent blocks forever. Neither helper kills the child on cancellation either, so an orphaned `docker` process outlives the coroutine.

src/puffo_agent/agent/adapters/docker_cli.py
```diff
@@ -706,7 +368,7 @@ def _build_command(
         extra_args: list[str],
         env_overrides: dict[str, str] | None = None,
     ) -> list[str]:
-        cmd: list[str] = ["docker", "exec", "-i"]
+        cmd: list[str] = [getattr(self, "_docker_bin", "docker"), "exec", "-i"]
```
hellohanchen, line 375: Flagging this as a latent primitive rather than a live bug — nothing leaks today, and the Codex path right next door does it correctly.

This emits `-e f"{key}={value}"` onto the argv, which is visible in `ps` to every local user on the host. It's inert right now because `cli_session.py:518` always passes `env_overrides = {}` ("reserved for future per-spawn env"), so no value ever flows through. But the moment anyone populates that dict with a token — which is exactly what a dict named `env_overrides` invites — the secret becomes world-readable on the process table.

`_ensure_codex_session` already demonstrates the right form: `-e NAME` with no value, letting Docker pull it from the daemon's own environment. Using that shape here too would make the leak unexpressible rather than merely unreached.

Compounding it slightly: `_run_cmd` puts `' '.join(cmd)` into its `RuntimeError` message, so the same value would also land in an exception and the daemon log.

For the record, the rest of the secret handling checks out and I verified it end to end: bearer tokens go in as `-e NAME` passthrough, `_prepare_codex_config` collects only variable *names*, the unset-var warning logs the name and not the value, `write_codex_mcp_config` writes `bearer_token_env_var = "<name>"`, the auth-sync helpers return closed-vocabulary mode strings, the refresh token is scrubbed to `""`, and there's no `shell=True` anywhere under `src/puffo_agent/agent/`. The PR's claim that only names are synchronized is accurate.

src/puffo_agent/portal/worker.py
```diff
@@ -1587,19 +1570,25 @@ async def _process_refresh_flags(
...
-            )
-            host_home = Path.home()
-            ah = agent_home_dir(agent_id)
-            skill_count = sync_host_skills(host_home, ah)
-            merged_mcp, _unreach = sync_host_mcp_servers(host_home, ah)
-            logger.info(
-                "agent %s: refresh_host_sync (skills=%d mcp=%d)",
-                agent_id, skill_count, merged_mcp,
-            )
+            if harness_name == "codex":
```
hellohanchen, line 1573: The comment says host sync is *"delegated to Codex adapter"*, but I can't find the delegation on either runtime:

- **cli-docker**: `sync_host_codex_skills` runs only inside `_ensure_started()`, which returns immediately once `self._started` is `True`. `reload()` re-runs `_prepare_codex_config` (MCP config) but never the skill sync.
- **cli-local**: `_verify()` returns early for codex before reaching any sync call.

So `refresh(host_sync=True)` on a Codex agent is a no-op that logs a success-shaped line. The operator edits `~/.codex/skills/`, hits refresh, sees the log, and nothing changes until the worker restarts — which is the worst shape for this kind of bug, because the log actively tells them it worked.

Either wire the sync into `reload()`, or change the message to say the sync happens at start and a restart is required.

Related, in the same removal: the loud `desired_mcps are not supported on cli-docker` guard went away and `desired_mcps` is now passed through to `DockerCLIAdapter` — but the new `filter_container_mcp_servers` is wired only into the *codex host-MCP* path. `install_claude_mcp` merges a template into `.claude.json` with no `_host_local_token` check. So a stdio template with an absolute host command now registers silently and fails at the first tool call inside the container, where before it failed loudly at spawn.

src/puffo_agent/agent/adapters/codex_session.py
```diff
@@ -262,6 +264,17 @@ def __init__(
                     self.agent_id, persisted_sandbox, self.sandbox,
                 )
                 self._conversation_id = ""
+            elif self.thread_cwd is not None:
+                persisted_thread_cwd = self._load_persisted_thread_cwd()
+                if persisted_thread_cwd != self.thread_cwd:
+                    logger.info(
+                        "agent %s: codex thread cwd changed (%s 鈫?%s); "
```
hellohanchen: Two things on this line.

**Mojibake.** The arrow reads `鈫?` — a UTF-8 `→` decoded as cp936 and re-encoded. The sibling sandbox-rotation message a few lines up has a correct `→`, so it's just this one. Cosmetic, but it ships into operator logs.

**The rotation is one-directional.** The guard is `elif self.thread_cwd is not None`, and `local_cli.py` passes `cwd=` but never `thread_cwd`. Both runtimes resolve to the *same* session file — `docker_cli.codex_home` and the cli-local `agent_codex_user_dir` are the same path — so:

1. Agent runs as `cli-docker`, persists a thread with `thread_cwd = "/workspace"`.
2. Operator switches it to `cli-local`.
3. `thread_cwd` is `None`, so the rotation branch is skipped and the thread resumes — with a codex-side cwd of `/workspace`, which doesn't exist on the host.

That's the same class of failure this PR fixes in the forward direction. Persisting and comparing the *effective* cwd (`thread_cwd or cwd`) unconditionally, rather than only when `thread_cwd` is set, would cover both directions with less code than the current special case.

The rest of the session handling checks out: `warm`/`run_turn`/`reload` all hold `_lock` so concurrent turns can't collide, a failed `thread/resume` falls back to a fresh thread, `_pending` is popped on response and on timeout and cleared in `_teardown_locked`, and `CODEX_HOME` maps to the bind-mounted host `.codex` so threads survive a container rebuild. The untested part is a missing or corrupt `codex_session.json` — `_load_persisted_thread_cwd` returns `""` on `OSError`/`JSONDecodeError`, but nothing exercises it.

hellohanchen:
Blocker cleared: the operator confirms there are no existing Gemini or Hermes `cli-docker` agents, so the migration concern from my [review](https://github.com/puffo-ai/puffo-agent/pull/210#pullrequestreview-4862004678) has no population to affect. Nobody gets stranded by this upgrade.

Two things in that finding were not actually about Gemini, though, and they survive the confirmation:

**1. Google is now unusable on every runtime, and `--provider google` still accepts it.**

`DEFAULT_HARNESS_FOR_PROVIDER[google] = gemini-cli` is still in the table, and `resolve_effective_harness` is applied to `cli-local` with no short-circuit. So `puffo-agent agent create --runtime cli-local --provider google` resolves to `gemini-cli`, `validate_triple` rejects it, and the operator gets an error naming a harness they never typed and cannot select. Combined with `README.md:408` telling them Gemini remains available on non-Docker runtimes, and `GEMINI_API_KEY` still documented in the README's key list, this is a forward-looking trap rather than a backward-compatibility one — "no existing agents" doesn't help.

Pick one and make it consistent: either drop `google` from the provider choices and the README along with the harness, or keep `gemini-cli` in `HARNESSES_FOR_RUNTIME[cli-local]`. Right now the flag, the docs, and the matrix disagree.

**2. The daemon's handling of *any* invalid `agent.yml` is the real finding.**

Nothing in this path is Gemini-specific:

- `daemon.py` catches every `AgentConfig.load` failure with `logger.warning(...)` + `continue`, writes no `RuntimeState`, and — because `_load_agent_cfg_cached` only populates its cache after a successful load — re-parses and re-logs on every reconcile tick (2s default).
- `cmd_agent_runtime` calls `AgentConfig.load` before applying its flags, so the command that exists to repair a bad runtime block cannot run against a bad runtime block. Same for `agent show`.
- The `continue` precedes `_worker_needs_restart`, so a worker already running when its config becomes invalid is never stopped.

Any future validation tightening, or an operator typo, lands in exactly this state: an agent that silently stops working, a portal that still shows it as fine, a log line every two seconds, and a repair command that traceback. That's worth fixing on its own terms — a one-shot log per (agent, mtime), an errored `RuntimeState` so the portal surfaces it, and a lenient load in `cmd_agent_runtime`.

Unchanged from the review and still worth doing: the resolver consulting `~/.puffo-agent/resolved_clis.json` ahead of `$PATH`, the staleness probes conflating "answered no" with "couldn't run" before `docker rm -f`, the missing test that a stale rebuild preserves workspace and `codex_session.json`, and the `refresh(host_sync=True)` no-op for Codex agents.

PR #212, opened 2026-08-04
feat(send): channel is_encrypted drives the outbound channel format

hellohanchen:
## What

Reimplements this PR's original 1.2.0-era format rules on the 2.0.0 architecture (branch replaced wholesale; old history superseded), against puffo-server's coordinated plaintext support (server PR #321, merged):

1. **Channel sends follow `channels.is_encrypted`** — encrypted → sealed envelope (unchanged); plaintext → signed plaintext envelope through the **same** `/v2/agent-runtime/messages:send` freshness pipeline (require_current / held / send_anyway all apply to plaintext channels). Plaintext sends skip device-key resolution entirely.
2. **DMs stay E2EE** — the DM path is untouched; the agent cannot choose a format.
3. **`CHANNEL_FORMAT_MISMATCH` recovery** — on the server's mismatch error the coordinator re-reads the channel policy from the server, updates the cache, and resends **once** in the other format; an unchanged policy surfaces the failure (`error_kind: channel_format_mismatch`).

**Policy cache** (`PuffoCoreMessageClient`): in-memory `_channel_encrypted` populated by space warms (the channel list now carries `is_encrypted`), persisted in `disk_cache` (policy preserved across policy-less rewrites), updated live by the new `channel_update` WS frame, evicted alongside the other channel caches. Unknown/missing/legacy always fails safe to **encrypted**. The coordinator takes the client as an optional `channel_policy_source`; without one (keyless MCP fallback) it stays always-encrypted — keyless channel sends already route through the bridge where the **server** picks the format.

Not needed anymore from the original plan: the old `send_mode` rules (turn-bundle / thread-root) were already deleted upstream in the 2.0.0 release, and synthetic system-message rows keep the store's fail-safe `is_encrypted=True` default.

## Tests

13 new (`test_channel_format_policy.py`): mismatch detector contract, disk-cache policy round-trip + preservation, sealed/plaintext format selection off the policy, mismatch → refresh → cross-format resend, mismatch with unchanged policy fails once, missing-policy-source default, warm/WS-update/ensure/refresh/eviction cache behavior, WS `channel_update` dispatch. Existing manual-construction fixtures updated for the new cache attr. Full suite: **0 failures** (2 pre-existing environment skips).

Depends on a puffo-server deployment carrying #321 (already on server main).

## Staging validation (2026-08-24)

Live smoke against chat-staging with a dedicated staging daemon: plaintext channels send/receive plaintext, encrypted channels stay sealed, and an encrypted→plaintext flip reached the agent via the `channel_update` frame before its next send — first attempt already used the new format, no mismatch retry needed. Policy changes now log one INFO line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- **Channel sends follow the channel's format policy.** The channel's
+  `is_encrypted` field alone decides sealed vs signed-plaintext for channel
+  sends; DMs stay encrypted and the agent cannot choose the format. The
+  policy is cached (disk + memory), kept live by `channel_update` frames,
+  and a `CHANNEL_FORMAT_MISMATCH` rejection refreshes the policy once and
+  resends in the channel's current format. (#212)
+
 ## [2.0.0] - 2026-08-23
 
 > Stable Agent Foundation 2.0 release, promoted from `2.0.0a25` without
--- a/src/puffo_agent/agent/channel_format.py
+++ b/src/puffo_agent/agent/channel_format.py
@@ -1,0 +1,22 @@
+"""Channel format policy: channels.is_encrypted drives the outbound format.
+
+Fail-safe: an unknown or missing policy answers encrypted.
+"""
+
+from __future__ import annotations
+
+import json
+
+from ..crypto.http_client import HttpError
+
+CHANNEL_FORMAT_MISMATCH = "CHANNEL_FORMAT_MISMATCH"
+
+
+def is_channel_format_mismatch(exc: Exception) -> bool:
+    if not isinstance(exc, HttpError) or exc.status != 400:
+        return False
+    try:
+        body = json.loads(exc.body or "{}")
+    except (TypeError, ValueError):
+        return False
+    return isinstance(body, dict) and body.get("error") == CHANNEL_FORMAT_MISMATCH
--- a/src/puffo_agent/agent/client_setup.py
+++ b/src/puffo_agent/agent/client_setup.py
@@ -99,6 +99,7 @@ def _cache_state() -> dict[str, Any]:
         "_warm_task": None,
         "_space_name_cache": {},
         "_channel_name_cache": {},
+        "_channel_encrypted": {},
         "_space_members": {},
     }
 
--- a/src/puffo_agent/agent/directory_cache.py
+++ b/src/puffo_agent/agent/directory_cache.py
@@ -173,6 +173,7 @@ async def warm_channels_for_space(
     store: Any,
     channel_spaces: dict[str, str],
     channel_names: dict[str, str],
+    channel_policies: dict[str, bool] | None = None,
 ) -> None:
     try:
         response = await http.get(f"/spaces/{space_id}/channels")
@@ -184,9 +185,13 @@ async def warm_channels_for_space(
         if not channel_id:
             continue
         channel_spaces.setdefault(channel_id, space_id)
+        # Missing/legacy policy fails safe to encrypted.
+        is_encrypted = channel.get("is_encrypted") is not False
+        if channel_policies is not None:
+            channel_policies[channel_id] = is_encrypted
         if name:
             channel_names.setdefault(channel_id, name)
-            disk_cache.persist_channel(channel_id, name, space_id)
+            disk_cache.persist_channel(channel_id, name, space_id, is_encrypted)
         try:
             await store.mark_channel_space(channel_id, space_id)
         except Exception:
--- a/src/puffo_agent/agent/disk_cache.py
+++ b/src/puffo_agent/agent/disk_cache.py
@@ -79,18 +79,31 @@ def persist_space(space_id: str, name: str) -> None:
         logger.debug("persist_space(%s) failed: %s", space_id, exc)
 
 
-def persist_channel(channel_id: str, name: str, space_id: str = "") -> None:
+def persist_channel(
+    channel_id: str,
+    name: str,
+    space_id: str = "",
+    is_encrypted: bool | None = None,
+) -> None:
     if not channel_id or not name:
         return
     try:
+        payload = {
+            "channel_id": channel_id,
+            "name": name,
+            "space_id": space_id,
+            "fetched_at": int(time.time()),
+        }
+        if is_encrypted is None:
+            # Preserve a previously persisted policy.
+            existing = (load_channel(channel_id) or {}).get("is_encrypted")
+            if isinstance(existing, bool):
+                payload["is_encrypted"] = existing
+        else:
+            payload["is_encrypted"] = bool(is_encrypted)
         _atomic_write_json(
             _cache_root() / "channels" / f"{_safe(channel_id)}.json",
-            {
-                "channel_id": channel_id,
-                "name": name,
-                "space_id": space_id,
-                "fetched_at": int(time.time()),
-            },
+            payload,
         )
     except OSError as exc:
         logger.debug("persist_channel(%s) failed: %s", channel_id, exc)
--- a/src/puffo_agent/agent/membership_events.py
+++ b/src/puffo_agent/agent/membership_events.py
@@ -71,6 +71,7 @@ async def evict_space_caches(
     space_members: dict[str, dict[str, str]],
     store: Any,
     log: Log,
+    channel_policies: dict[str, bool] | None = None,
 ) -> None:
     if not space_id:
         return
@@ -79,6 +80,8 @@ async def evict_space_caches(
     ]:
         channel_spaces.pop(channel_id, None)
         channel_names.pop(channel_id, None)
+        if channel_policies is not None:
+            channel_policies.pop(channel_id, None)
     space_names.pop(space_id, None)
     space_members.pop(space_id, None)
     try:
@@ -98,11 +101,14 @@ async def evict_channel_caches(
     channel_names: dict[str, str],
     store: Any,
     log: Log,
+    channel_policies: dict[str, bool] | None = None,
 ) -> None:
     if not channel_id:
         return
     channel_spaces.pop(channel_id, None)
     channel_names.pop(channel_id, None)
+    if channel_policies is not None:
+        channel_policies.pop(channel_id, None)
     try:
         await store.unmark_channel_space(channel_id)
     except Exception:
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -943,6 +943,12 @@ class PuffoCoreMessageClient:
             self._channel_name_cache[channel_id] = name.strip()
         policy = update.get("is_encrypted")
         if isinstance(policy, bool):
+            if self._channel_encrypted.get(channel_id) != policy:
+                logger.info(
+                    "channel %s policy -> %s",
+                    channel_id,
+                    "encrypted" if policy else "plaintext",
+                )
             self._channel_encrypted[channel_id] = policy
             cache_name = (
                 self._channel_name_cache.get(channel_id)
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -279,6 +279,7 @@ class PuffoCoreMessageClient:
         self._ws.on_message = receipt_handler.handle
         self._ws.on_event = self._handle_event
         self._ws.on_space_membership_changed = self._handle_space_membership_changed
+        self._ws.on_channel_update = self._handle_channel_update
         # Re-warms caches on every (re)connect, first connect included.
         self._ws.on_connect = self._on_ws_connect
         await self.store.open()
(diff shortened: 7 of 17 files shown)


hellohanchen (commented):
Review of `puf-411-agent-send-decision` @ `20a7d566` (base `431bb77b`).

Ran the suite locally: **EXIT=0**, 3 skipped. I excluded the three `test_ui_*` modules (`PySide6` not installed in my sandbox) and `test_chat_local_tool_use.py` (`anthropic` missing) — environment gaps, not this PR. I did not run the staging smoke.

The shape is right. Making the channel the single authority is correct, and two decisions in particular are better than the obvious alternatives: the retry keys off the **server's** `CHANNEL_FORMAT_MISMATCH` error code and only fires when the refreshed policy actually *differs*, rather than blindly resending in the opposite shape — so an unrelated 400 surfaces instead of being retried wrong, and a server that keeps rejecting can't be looped on. And `persist_channel`'s `_UNSET` sentinel correctly stops a name-only cache update from clobbering a known-plaintext policy.

Five findings, none of them blockers for the encrypted-only status quo (everything defaults to encrypted, so nothing changes until a channel is actually flipped). F1 and F2 will bite the first time someone flips one.

---

### F1 (medium) — `persist_channel` fabricates a policy, and `ensure_channel_policy` trusts the fabrication

`disk_cache.persist_channel` with `is_encrypted` unset writes a *guess*:

```py
if is_encrypted is _UNSET:
    existing = load_channel(channel_id) or {}
    policy = existing.get("is_encrypted") is not False   # no file -> True
```

With no prior cache entry this writes a literal `"is_encrypted": true` to disk. `ensure_channel_policy` then reads it back through an `isinstance(..., bool)` check, which cannot distinguish "the server said encrypted" from "we defaulted because we never asked":

```py
cached = disk_cache.load_channel(channel_id) or {}
if isinstance(cached.get("is_encrypted"), bool):
    policy = cached["is_encrypted"]
    self._channel_encrypted[channel_id] = policy
    return policy          # <- server never consulted
```

The reachable writer is `_resolve_channel_name`'s event-replay fallback (`puffo_core_client.py:2852`), which calls `persist_channel(channel_id, name, space_id)` with no policy. That fallback runs whenever `GET /spaces/{sp}/channels` fails or doesn't list the channel. It is called at `:3015` (channel intro) and `:3299` (membership system message) **immediately before** `ensure_channel_policy` on the same channel.

Verified against the real objects — the server is answering truthfully and is never asked:

```
disk after name-only persist: {..., 'is_encrypted': True, ...}
ensure_channel_policy -> True   (server says False)
http.get called: 0 times
```

Blast radius: the intro and membership system rows persist `is_encrypted=True` on a plaintext channel, so the metadata block shown to the LLM misstates the channel's confidentiality. The poison also lands in `_channel_encrypted`, which `channel_policy()` (the send path) reads — sends recover after one 400 via the retry, but the wrong disk value survives restarts.

`test_ensure_policy_uses_disk_then_warms_server` and `test_legacy_null_cache_entry_fails_safe_to_encrypted` show the intended contract is "unknown -> warm from the server"; the `_UNSET` branch is what makes "unknown" unrepresentable. Suggest omitting the key (or writing `None`) when the caller doesn't know — `channel_policy`'s `is not False` already treats a missing key as encrypted, so the fail-safe is unchanged and `ensure_channel_policy` regains its server warm.

### F2 (medium) — the data-service fail-safe makes the retry gate unsatisfiable

`data_service.get_send_encryption`:

```py
client = _client_for(agent_id)
if client is None or not channel_id:
    return web.json_response({"encrypt": True})
```

This branch is taken for **both** the normal read and `refresh=1`. So while the agent's client isn't registered with the resolver, the MCP subprocess sees:

1. `get_send_encryption` -> `True` -> sealed send to a plaintext channel -> `CHANNEL_FORMAT_MISMATCH`
2. `_refresh_channel_policy` -> same endpoint -> `True` again
3. `wanted == encrypt` -> `raise`

The mismatch reaches the LLM as a raw transport error, and every retry repeats it. The recoverable case is being reported as a hard failure — and the disk cache this PR added holds the correct answer the whole time, but this path only reaches it through a live client object. `test_send_encryption_fail_safe_for_unknown_agent` pins the `True`; the problem isn't the fail-safe, it's that `refresh` inherits it. Falling back to `disk_cache.load_channel(channel_id)` in the `client is None` branch fixes both the dead end and the inconsistency with `channel_policy`.

### F3 (medium-low) — the server already pushes policy flips over the WS and the agent drops the frame

`space_config.rs:387` broadcasts `WsPush::ChannelUpdate { space_id, channel_id, name, description, is_encrypted, updated_at }` to every space member, and `WsPush` is `#[serde(tag = "type", rename_all = "snake_case")]` — so the frame arrives as `{"type": "channel_update", ...}`. `ws_client.py:_handle_frame` branches on `ping` / `message` / `cert_update` / `event` and silently discards anything else.

Consequences: every flip costs a guaranteed 400 plus a full `/spaces/{id}/channels` refetch on the next send; and the system-message paths, which have no retry, keep persisting the stale value until something else warms the space. Handling the push would make the caches converge without any failed request. One trap if you wire it up: `is_encrypted` is `Option<bool>` server-side, so a name-only update pushes `is_encrypted: null` — that has to mean "unchanged", not "encrypted".

### F4 (low) — "skip recipient/device resolution for plaintext channel sends" is only half true

Device-key resolution is correctly inside `if encrypt:` in `_post_respecting_channel_format`. But **member** resolution isn't: both `send_message` (`puffo_core_tools.py:517`) and `send_message_with_attachments` (`:1121`) still do an unconditional `GET /spaces/{sid}/channels/{cid}/members` and raise `"channel ... has no resolvable members"` before `encrypt` is even computed.

`send_fallback_message` does it the other way — the members GET sits inside `if encrypt:`, and `test_plaintext_fallback_send_needs_no_recipients` pins that. So the two paths now disagree about the same channel: a plaintext channel the daemon fallback can post to would be refused by the MCP tool. In practice an empty member list is near-impossible on a membership-filtered endpoint, so this is mostly a wasted round trip per plaintext send plus a divergence someone will trip over later.

### F5 (low) — the deliberate old-shim tolerance is gone

The previous `_send_encryption_required` documented that it "tolerates a data client still on the old 2-arg signature, so a harness pinned to the previous shim keeps sending rather than erroring". The `getattr(..., "get_send_encryption", None)` guard survives, but it only detects **absence** — the signature went from `(slug, thread_root_id)` to `(channel_id)`, and a stale shim now raises `TypeError`, which nothing around the call site catches. Both in-repo shims are updated so this is skew-only, and the HTTP path degrades safely (an old daemon reads the missing `slug`/`thread_root_id` params and answers encrypt). Worth either restoring the tolerance or dropping the now-misleading `getattr` guard, since it no longer guards the failure mode it was written for.

---

Also checked and clean: `_post_respecting_channel_format`'s loop always exits via `return` or `raise` (no implicit `None`); the retry rebuilds the envelope from `inp` so already-uploaded blobs aren't re-sent; both directions of the flip return `CHANNEL_FORMAT_MISMATCH` server-side (`messages.rs:357` and the plaintext endpoint, per `messages_plaintext_tests.rs:405,447`), so the retry works either way; `_warm_channels_for_space` has no debounce, so `refresh_channel_policy` really does re-fetch; `/spaces/{id}/channels` really does return `is_encrypted` (`space_config.rs:663`), so the agent's authority is the right one; and the space/channel eviction paths remembered to drop `_channel_encrypted`.

One note outside this PR's scope: attachments are always blob-encrypted with a per-file content key carried in `AttachmentMeta` inside the envelope content, so a plaintext send publishes that key in the clear. That's presumably intended for a plaintext channel and the branch predates this PR — flagging only because PUF-410 is about to make it genuinely reachable for the first time.

Glimmer0x:
Reviewed current head 2abe301. The focused channel-format tests pass. Two retry-path gaps remain:

1. Attachment sends upload twice after CHANNEL_FORMAT_MISMATCH. _resolve_route_and_content runs before the loop and _prepare_attachments uploads the blobs. On mismatch the code calls _resolve_route_and_content again; passing materialized only reuses plaintext bytes, while _prepare_attachments re-encrypts and POSTs /blobs/upload again. The first upload becomes orphaned. Please retain the prepared attachment metadata across the format retry and add a regression asserting one blob upload.

2. Plaintext sends still unconditionally call _channel_recipient_slugs before the encrypt branch. That makes a valid plaintext send depend on the channel-members endpoint and rejects it when that lookup is empty or unavailable, even though the plaintext envelope needs no recipient devices. Please resolve channel members only for encrypted sends and add a no-member-lookup regression for plaintext.

Neither changes the default-encrypted safety property, but both are reachable once plaintext channels are enabled.

hellohanchen:
The review was written against the old 1.2.0-era head (20a7d56); the branch has since been replaced wholesale by the 2.0.0 reimplementation. Mapping each finding onto the current head (30164a5):

- **F1 — resolved by design.** `disk_cache.persist_channel(..., is_encrypted=None)` now **omits** the key when there is no previously persisted bool (`disk_cache.py`), so "unknown" stays representable: `ensure_channel_policy` finds no bool on disk and warms from the server. Name-only rewrites preserve a known policy but never fabricate one. Pinned by `test_disk_cache_round_trips_and_preserves_policy` and `test_ensure_channel_policy_reads_disk_then_warms`.
- **F2 — not applicable.** The 2.0.0 architecture has no `send-encryption` data-service endpoint and no MCP-subprocess HTTP hop for policy: `SendCoordinator` takes the client directly as `channel_policy_source`, and the mismatch retry refreshes through the same in-process client (server re-read via `/spaces/{id}/channels`). The unsatisfiable-refresh dead end can't be constructed.
- **F3 — implemented.** `ws_client` now dispatches `channel_update` to `PuffoCoreMessageClient._handle_channel_update`, which updates memory + disk and logs policy changes. The `is_encrypted: null` trap is handled: the handler only acts on `isinstance(policy, bool)`, so a name-only update leaves the policy untouched. Verified live on staging: an encrypted→plaintext flip reached the agent before its next send — first attempt used the new format, zero mismatch retries.
- **F4 — confirmed and fixed in 30164a5.** The port had the same shape: `_resolve_route_and_content` fetched channel members unconditionally and raised on an empty list. Members only feed device wraps and missing-device supplementation (both encrypted-only), so the fetch is now gated on `encrypt`; plaintext sends pass `recipient_slugs=[]` (the supplementation consumer no-ops on an empty list). Mutation-verified: `test_plaintext_channel_send_survives_an_empty_member_list` + members-fetch assertion go red with the gate reverted; `test_encrypted_channel_send_still_requires_members` pins the encrypted guard.
- **F5 — not applicable.** The `get_send_encryption` shim and its `getattr` guard don't exist in the 2.0.0 architecture (no data-client signature to skew against).

Full suite: 3199 passed, 5 skipped (PySide6/Windows-only environment gaps).

Glimmer0x:
Re-reviewed current head 30164a5.

The plaintext member-resolution finding is fixed: member lookup is now gated by encrypt, the new positive/negative regressions are appropriate, and the focused suite passes locally (20 tests).

Two items remain:

1. The attachment mismatch retry finding is unchanged. On CHANNEL_FORMAT_MISMATCH, the second _resolve_route_and_content call still invokes _prepare_attachments again, so the same attachment is re-encrypted and uploaded a second time, leaving the first blob orphaned. Please retain prepared attachment metadata across the format retry and add a one-upload regression.

2. The new commit currently fails the repository pre-commit gate. Local reproduction:
   src/puffo_agent/agent/send_coordinator.py:1274: _resolve_route_and_content is 105 lines (limit 100)
   CI pre-commit reports the same failure. Please extract the recipient-resolution branch or otherwise bring the function back under the structural limit.

So F4 is resolved, but this head is not yet merge-ready because pre-commit is red; the earlier attachment finding also remains open.

hellohanchen:
Both items addressed in dd695c0:

1. **Attachment double-upload** — `_resolve_route_and_content` now returns the prepared attachment metas (`prepared_attachments`) alongside the route, and the mismatch retry passes them back via a new `prepared` parameter, so `_prepare_attachments` (encrypt + `POST /blobs/upload`) runs once per send. Regression: `test_format_mismatch_retry_uploads_attachments_once` drives a plaintext→sealed mismatch retry with a real file and asserts exactly one blob upload across both attempts. Mutation-verified: with the `prepared=` pass-through reverted, the test goes red on a second upload.

2. **Structural limit** — extracted `_resolve_send_recipients` and `_resolve_send_devices`; `_resolve_route_and_content` is back under 100 lines and `tools/check_python_structure.py` passes clean locally (the same hook the CI pre-commit job runs).

Full suite: 3200 passed, 5 skipped.

Glimmer0x:
Re-reviewed current head `dd695c092d4a1cc9e9d66e8bea66e18f75641244`. The two remaining issues from my prior pass are resolved:

- The format-mismatch retry now carries the prepared attachment metadata forward, so it reuses the original uploaded blob instead of encrypting/uploading the file again. The new regression test exercises the mismatch path and verifies exactly one `/blobs/upload` call.
- `_resolve_route_and_content` has been split below the structural limit; local `pre-commit run --all-files` passes.

I also reviewed the new helper extraction and retry wiring and did not find a new blocking issue. Local focused tests pass (`21 passed`: `tests/test_channel_format_policy.py` + `tests/test_plaintext_send.py`), and the current GitHub CI/CodeQL checks are all green. No further blocking findings from me on this head.

PR #221, opened 2026-08-09
Require explicit opt-in for Claude API key auth

hellohanchen:
## Summary

- ignore ambient `ANTHROPIC_API_KEY` for Claude Code in both `cli-local` and `cli-docker`; subscription authentication remains the default
- add `anthropic.cli_use_api_key` to `daemon.yml` and inject the daemon-owned `anthropic.api_key` only when the key is non-empty and the opt-in is literal YAML `true`
- prevent per-agent `env_overrides` and Claude `settings.json` files from bypassing the policy; settings are scrubbed before every Claude spawn
- skip OAuth refresh gates only in explicit API-key mode, show API-key-specific recovery instructions, and restore healthy state after a successful turn
- extend `puffo-agent config` so operators can update or clear `anthropic.api_key` and independently enable or disable Claude CLI API-key mode

## Configuration

Enable daemon-owned API-key authentication for Claude Code:

```bash
puffo-agent config \
  --anthropic-api-key KEY \
  --anthropic-cli-use-api-key true
```

Return Claude Code to subscription authentication without deleting the stored key:

```bash
puffo-agent config --anthropic-cli-use-api-key false
```

Clear the stored key:

```bash
puffo-agent config --anthropic-api-key ''
```

A daemon restart is required after changing these daemon-level values.

## Validation

- affected suite: 410 passed and 1 expected skip, repeated 3 consecutive times
- full suite: 2421 passed and 2 expected skips
- changed executable lines and branches: 100% covered
- isolated CLI smoke verified combined updates, partial updates, key clearing, and no key leakage in command output
- manual runtime smoke: the daemon started with a deliberately invalid ambient key; Linus (`cli-docker`, Claude Code) received an empty `ANTHROPIC_API_KEY`, completed the turn through subscription auth, and returned to healthy state
- GitHub CI passed on Python 3.11 and 3.12, with CodeQL checks green

One earlier full-suite run hit the existing `test_port_fallback` free-port race because `requested + 1` was occupied. The test passed three isolated reruns and the subsequent full suites passed cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,9 +50,12 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 - **Claude Code uses subscription authentication unless API-key mode is
   explicitly enabled.** Ambient `ANTHROPIC_API_KEY` values are ignored for
-  local and Docker CLI agents. A daemon-owned `anthropic.api_key` is passed to
-  Claude Code only when `anthropic.cli_use_api_key: true` is set in
-  `daemon.yml`; both values can also be updated with `puffo-agent config`.
+  local and Docker CLI agents, including per-agent environment and settings
+  overrides. A daemon-owned `anthropic.api_key` is passed to Claude Code only
+  when `anthropic.cli_use_api_key: true` is set. Operators can manage both
+  values with `puffo-agent config --anthropic-api-key KEY` and
+  `--anthropic-cli-use-api-key true|false`. API-key failures now show the
+  correct recovery steps and clear after the next successful turn.
 
 - **Codex Docker agents can write to their mounted workspace reliably.** The
   container is now the effective filesystem sandbox, avoiding unsupported
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   explicitly enabled.** Ambient `ANTHROPIC_API_KEY` values are ignored for
   local and Docker CLI agents. A daemon-owned `anthropic.api_key` is passed to
   Claude Code only when `anthropic.cli_use_api_key: true` is set in
-  `daemon.yml`.
+  `daemon.yml`; both values can also be updated with `puffo-agent config`.
 
 - **Codex Docker agents can write to their mounted workspace reliably.** The
   container is now the effective filesystem sandbox, avoiding unsupported
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **Claude Code uses subscription authentication unless API-key mode is
+  explicitly enabled.** Ambient `ANTHROPIC_API_KEY` values are ignored for
+  local and Docker CLI agents. A daemon-owned `anthropic.api_key` is passed to
+  Claude Code only when `anthropic.cli_use_api_key: true` is set in
+  `daemon.yml`.
+
 - **Codex Docker agents can write to their mounted workspace reliably.** The
   container is now the effective filesystem sandbox, avoiding unsupported
   nested sandboxing, and resumed local and Docker sessions reapply their
--- a/config.example.yml
+++ b/config.example.yml
@@ -17,6 +17,7 @@ default_provider: "anthropic"  # "anthropic" | "openai" | "google"
 
 anthropic:
   api_key: ""                    # https://console.anthropic.com/
+  cli_use_api_key: false          # true = Claude Code uses this key instead of subscription auth
   model:   "claude-sonnet-4-6"   # empty = provider default
 
 openai:
--- a/src/puffo_agent/agent/_invite_strings.py
+++ b/src/puffo_agent/agent/_invite_strings.py
@@ -134,6 +134,32 @@ def format_oauth_expired(agent_id: str, agent_display_name: str = "") -> str:
     )
 
 
+def format_anthropic_api_key_rejected(
+    agent_id: str, agent_display_name: str = "",
+) -> str:
+    """Bilingual recovery copy for daemon-owned Anthropic API keys."""
+    label = (
+        f"**{agent_display_name}** (`{agent_id}`)"
+        if agent_display_name else f"`{agent_id}`"
+    )
+    return (
+        f"⚠️ {label} — my Anthropic API key was rejected, so I can't "
+        "answer until it is corrected.\n\n"
+        "**On the computer where puffo-agent is running:**\n"
+        "1. Update `anthropic.api_key` in `daemon.yml` (or run "
+        "`puffo-agent config --anthropic-api-key KEY`).\n"
+        "2. Keep `anthropic.cli_use_api_key: true`.\n"
+        "3. Restart puffo-agent, then send me another message.\n\n"
+        f"⚠️ {label} — 我的 Anthropic API key 被拒绝，需要修正后才能"
+        "继续回复。\n\n"
+        "**在运行 puffo-agent 的电脑上：**\n"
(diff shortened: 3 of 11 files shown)


hellohanchen (commented):
Reviewed `9156d3c9` against `origin/main` (10 files, +407/−8). Read the full non-test diff plus the surrounding call graph: `worker.build_adapter` → both CLI adapters → `cli_session.ClaudeSession._spawn`, `daemon._register_with_refresher` / `_notify_refresh_for` / `_ensure_fresh_for`, `DaemonConfig.load/save`, `cmd_config`, and the control-op `env_overrides` path. Ran `tests/test_claude_api_key_policy.py` plus every daemon/worker/state/adapter/credential test file — **427 passed**. Docker is not runnable in my sandbox, so the container-side behaviour below is reasoned from the code and docker's documented semantics, not executed.

## What I verified as correct

- **`DaemonConfig.save()` round-trips the flag.** `save()` serializes `asdict(self.anthropic)`, so the new `cli_use_api_key` survives a `puffo-agent config` rewrite. The `cmd_config` change (mutate `cfg.anthropic.api_key`/`.model` in place rather than replacing with a bare `ProviderConfig`) is what makes that hold — replacing the object would have downgraded the subclass and silently dropped the opt-in. Both are covered by tests.
- **No operator-owned file is mutated.** `strip_claude_api_key_from_settings` is only ever pointed at `agent_home_dir/.claude/settings.json` and `agent_cfg.resolve_claude_dir()/settings{,.local}.json` — all per-agent. The host's real `~/.claude/settings.json` is untouched. This was my main worry going in; it's clean.
- **The secret stays out of argv.** `-e ANTHROPIC_API_KEY` (name-only passthrough) plus `env=self._claude_docker_env()` on the `docker exec` spawn is the right shape and matches the SECURITY comment already in `_build_command`.
- **`env=` actually replaces the environment.** `cli_session._spawn` passes `env=self.env` straight to `create_subprocess_exec` with no re-merge of `os.environ`, so the `env.pop("ANTHROPIC_API_KEY", None)` in both adapters is effective.
- **Ordering in `local_cli._ensure_session` is right.** The three strips run *before* `_write_permission_hook_settings()`, which re-reads and rewrites `claude_dir/settings.json` — so the hook writer can't resurrect a key it just read.
- **All the new `None` callbacks are guarded.** `_notify_refresh_for`/`_ensure_fresh_for` returning `None` is safe: worker lines 639, 706, 1283 all null-check, and `on_auth_failure=None` is null-checked at line 542.
- **Non-CLI paths are unaffected.** `claude_cli_api_key()` is reached only via `_claude_cli_api_key(daemon_cfg, harness.name())`, which returns `""` for codex; the sdk runtime still reads `daemon_cfg.anthropic.api_key` directly, as before.

## N1 (medium) — in API-key mode, `auth_failed` is a one-way door, with the wrong recovery copy

`_uses_claude_api_key()` makes `_register_with_refresher` early-return. But the **only** transition that clears `runtime.health == "auth_failed"` back to `"ok"` is the refresher's `on_refresh_success` callback (`worker.py:625`) — and `_resolve_health_on_success` (`worker.py:802`) deliberately only handles `in_progress → ok`, skipping "any in-turn red".

So for an API-key agent:
1. Any auth-class CLI error — invalid key, revoked key, exhausted credit, a 401 from a transient provider blip — takes the `worker.py:695` branch: `health = "auth_failed"`, error text *"Claude Code sign-in expired… run `claude auth login`"*, and `_on_auth_failed_enter()` DMs the operator that same OAuth copy.
2. That remediation is wrong for this agent — it doesn't use OAuth at all; the fix is to correct `anthropic.api_key` in `daemon.yml`.
3. After the operator does fix it, nothing ever clears the flag. Turns succeed (message delivery isn't gated — `_ensure_fresh_token is None` short-circuits the `worker.py:1283` pre-delivery gate), but the agent stays permanently red in the CLI/heartbeat/UI until the worker is restarted.

The `_auth_failed_notification_sent` dedup flag also only re-arms on the auth_failed *clear*, so a second genuine failure after a restart-less recovery won't re-notify.

Suggested shape: when `_uses_claude_api_key`, have `_resolve_health_on_success` also clear `auth_failed` (a successful turn is proof the key works), and branch the error string / operator DM on API-key vs OAuth mode. No test currently covers health recovery in API-key mode.

## N2 (low–medium) — `env_overrides` bypasses the blanking in cli-docker, but not in cli-local

The two adapters apply the policy at different points relative to `env_overrides`, and the difference is load-bearing:

- `local_cli` builds `env = {**os.environ, **env_overrides, **adapter_owned}` and *then* does `env.pop("ANTHROPIC_API_KEY", None)` — an override can't get through. ✅
- `docker_cli._build_command` emits `-e ANTHROPIC_API_KEY=` (or the name-only passthrough) **first**, then appends `-e {key}={value}` for every `env_overrides` entry. A `ANTHROPIC_API_KEY` entry therefore lands *after* the blanking flag; per docker's last-wins semantics for a repeated `-e NAME=VALUE` it takes effect, defeating the opt-in. Even under first-wins it still puts a raw key in the process list — exactly what the SECURITY comment three lines above forbids.

Reachability: the control-op path is safe — `validate_env_overrides` enforces `ENV_OVERRIDE_WHITELIST` (compact-pct keys, numeric values only). But `AgentConfig.load` (`state.py:1272`) does **not** re-validate; it takes `raw["env_overrides"]` verbatim. So a hand-edited `agent.yml` — including one edited by the agent itself, which can reach `../agent.yml` from its workspace — reintroduces the key. That's operator-adjacent rather than adversarial, but it's the one surface where the invariant this PR is establishing doesn't hold.

Cheap fix: `continue` on `ANTHROPIC_API_KEY` in the `env_overrides` loop, the same way `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` already is.

## N3 (low) — the settings scrub is per-session, but `claude` respawns per turn

`strip_claude_api_key_from_settings` runs inside `_ensure_session`, which returns early once `self._session` exists. `ClaudeSession._spawn` is called again on every turn (and on every `--resume` failure), and Claude Code re-reads `settings.json` fresh at each spawn — the comment at `local_cli.py:805` says exactly that. So a key written into `claude_dir/settings.json` *after* the first turn takes effect on the very next spawn, with no re-scrub, until the worker restarts. Given agents are told `.claude/` in their workspace is theirs to edit, that's a plausible accident rather than a hypothetical. Scrubbing from `_build_command` (or at the top of `_spawn`) would make the invariant hold per-spawn instead of per-session.

Related, smaller: the scrub list covers `agent_home/.claude/settings.json` but not `agent_home/.claude/settings.local.json`, while the project-level dir gets both. If that asymmetry is deliberate (user-level `settings.local.json` isn't a layer Claude Code honours) it's fine — worth a one-line comment so the next reader doesn't have to work it out.

## N4 (nit) — `-e ANTHROPIC_API_KEY=` sets empty, it doesn't unset

`docker exec` has no unset flag, so this is the only option, and Claude Code's falsy check makes it work. But the variable is now *present-and-empty* inside the container for every child process — anything doing `"ANTHROPIC_API_KEY" in os.environ` (an MCP server, a hook, a user script) sees it. Worth a comment noting the distinction is intentional.

## N5 (nit) — stray whitespace edit

`state.py` lost the blank line before `def save()`, leaving `return cfg` directly adjacent to the next method. Looks accidental.

## Coverage boundary

`tests/test_claude_api_key_policy.py` is well-targeted — round-trip, the `is True` strictness, `cmd_config` preservation, ambient-key rejection, the scrubber's non-object/empty-env/write-failure paths, both adapters, the worker matrix, codex exclusion, and the daemon OAuth gate. Untested: health recovery in API-key mode (N1), the docker `env_overrides` interaction (N2), and the docker-side scrub of `claude_home_src/settings.json` (only the local variant is asserted). Nothing here is a merge blocker except N1, which I'd want addressed before an agent runs on a key that can expire.

PR #224, opened 2026-08-10
(Reopen) PUF-420 (agent): spawn claude by resolved path, and wrap Windows shims

hellohanchen:
Closes PUF-420. An agent on Windows died on its first turn with WinError 2. Reporter's manual patch (Jeremy) confirmed the fix direction.

## Problem

`local_cli.py:952` built argv as bare `["claude"]`. On POSIX, `execvp` searches PATH so this worked; on Windows, `CreateProcess` does not, so an npm-installed CLI (`claude.cmd`/`claude.ps1`/extensionless `claude` shell script, no `.exe`) failed with WinError 2 — the same error code as "not installed at all", which sent the traceback down the wrong path.

The resolver was already in the codebase: `_verify` on line 1021 called `resolve_claude_bin()` and used it only as an existence gate, discarding the path. The codex adapter three hundred lines earlier does the right thing.

## Fix

Two helpers in `cli_bin.py` so the next adapter inherits rather than repeats:

- **`spawn_argv(binary)`** — on Windows, routes `.cmd`/`.bat` through `cmd.exe /c` and `.ps1` through `powershell -NoProfile -NonInteractive -File`. A shim can't be argv[0]; it isn't a PE image. Preference order: `.exe` > `.cmd` > `.bat` > `.ps1`. POSIX passes through unchanged.
- **`windows_runnable(binary)`** — upgrades an extensionless resolution to a sibling shim. `shutil.which` can legitimately return the npm-installed shell script; CreateProcess rejects it with the same WinError 2 as "not found". This is the failure mode that made the original traceback so hard to read.

`_claude_bundle_paths()` gains the `%APPDATA%\npm\claude.{exe,cmd,ps1}` shims for narrow-PATH daemon start. Error message is rewritten to explicitly name the Windows shim case.

## One existing test modified

`test_cli_local_mcp_trust.py::test_command_shape_matches_cli_docker_for_bypass` asserted `cmd[0] == "claude"` — that check was incidentally pinning the bug while testing flag ordering. Changed to `Path(cmd[0]).name.startswith("claude")` to preserve the actual intent. Comment explains why.

## Tests

`test_windows_claude_spawn.py` — 16 new test cases (17 with parametrization) covering:

- POSIX pass-through (including a `.cmd`-named file on Linux stays a file)
- Windows `.exe` direct
- Windows `.cmd`/`.bat` wrapped via cmd.exe
- Windows `.ps1` wrapped via powershell
- Extensionless resolution → sibling shim upgrade
- `_build_command` uses resolver's path, not bare name
- Windows shim wrapping preserves following arguments

Baseline-anchored: **suite `2349 = 2332 baseline + 17`** on freshly-baselined `origin/main`.

## Real limits (Cal + Solution both agree)

1. **No Windows CI.** `sys.platform` is monkey-patched. Tests pin argv construction, NOT that CreateProcess actually accepts the result. The only real proof is a Windows run. Jeremy's manual patch on his machine ships the same argv shape and worked — closest thing we have.
2. **Reverse-verification weaker than usual.** Stashing the source causes `ImportError` at collection (helpers missing), not behavior-red assertions. Proves "this API is required," not "argv is correctly assembled."

## Recommend: Vase asks Jeremy to run this on his real Windows machine before merging.

The reporter has the only real-Windows testing environment; every automated check we have is on Linux.

## Not done

Intake option (b) — wrap the codex adapter similarly. Codex is currently working (uses resolved path); intake flagged as follow-up. Fold as separate PR if needed.

Linear: PUF-273. Ticket: PUF-420.

### DIFF
--- a/src/puffo_agent/agent/adapters/local_cli.py
+++ b/src/puffo_agent/agent/adapters/local_cli.py
@@ -43,7 +43,13 @@ from ...portal.state import (
     sync_host_plugins,
     sync_host_skills,
 )
-from ..cli_bin import resolve_claude_bin, resolve_codex_bin, resolve_hermes_bin
+from ..cli_bin import (
+    CLAUDE_BIN_MISSING,
+    resolve_claude_bin,
+    resolve_codex_bin,
+    resolve_hermes_bin,
+    spawn_argv,
+)
 from .base import Adapter, TurnContext, TurnResult
 from .cli_session import AuditLog, ClaudeSession
 from .codex_session import CodexSession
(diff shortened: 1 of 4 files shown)


Glimmer0x:
Superseded by the current Driver architecture on `main`. Windows shim normalization now lives in `agent/cli_bin.py::normalize_launch_argv`, is consumed by the current harness Driver path, and has focused coverage in `tests/test_cli_bin.py` / `tests/test_harness_driver.py` (implementation commit `c7146e7`). This PR targets the retired `local_cli.py` path and is now conflicting, so merging it would reintroduce the old boundary rather than add missing behavior.

PR #200, opened 2026-07-25
feat(primer): slim system prompt — How-to-chat rewrite + skill-per-category reorg

hellohanchen:
## What & why

Slims the shared primer to the essentials (how to chat + collaboration) and moves every tool's usage into per-category skills that agents load on demand. Stacked on #194 (notes) — base is `feat/notes-mcp-tools`.

## Primer (`DEFAULT_SHARED_CLAUDE_MD`)

- **Concepts-first ordering**: Spaces/channels/DMs moved above "How messages arrive"; intro reframed ("You communicate with humans and other agents on Puffo").
- **Metadata block**: `post_id` → `envelope_id` (renderer too), `channel`/`channel_id` absent for DMs (renderer omits them), new `direct_message: true` (DM-only), `thread_root_id` + `sender_slug` comments tightened.
- **"How to reply" → "How to chat"**: understand the message and your role first, then seven decisions — reply-or-`[SILENT]`; thread vs root (preference → standing rules → precedent → consensus → DM=root / busy-channel=thread); explicit visibility + auto-correction; freshness re-pull; proactively starting conversations; using memory; multi-message turns. Mechanics kept to agent-facing rules (daemon internals removed per agent feedback).
- **Toolkit section**: signature-only, grouped into 9 categories, each pointing at its skill.
- Memory + Shared-filesystem become sub-sections of "Your workspace".

## MCP changes

- `get_post` → **`get_envelope(envelope_ref)`** (all wiring surfaces).
- `send_message` gains **`dm="<slug>"`** (≡ `channel="@slug"`, mutually exclusive, `@`-tolerant).
- `get_dm_history` gains **`since=<msg_id>`** through the whole chain (store SQL → data-service route → HTTP + in-process clients → tool).
- `list_skills` now scans the **codex mirror dir** (`.agents/skills`) too — codex agents previously saw none of the managed skills.

## Skills reorg (10 managed skills)

[...]

PR #227, opened 2026-08-11
(Reopen) PUF-402: forward-port the sync_host_mcp regression test to main

hellohanchen:
## PUF-402: forward-port the sync_host_mcp regression test to main

**Zero behavior change. Pure test.** The bug the test guards against is *not* on `main` today — main correctly forwards `sync_host_mcp(template_id=...)` → `sync_mcp(template_id=template_id)`. The bug happened on `feat/slim-primer` (PR #200) when commit `34a31df` renamed the agent-facing param to `name` and left the forward naming the old one — every call raised `TypeError` at runtime because `PuffoRpcClient.sync_mcp` is keyword-only `template_id`.

The 1-line fix + 2 regression tests originally went out on `feat/slim-primer` stacked with the UX rename. Per Vase's `msg_43e7c69a` — "open a new puffo-agent PR based on latest main" — this PR forward-ports only the **regression tests**, without extracting the `34a31df` UX rename (that's a product decision, held separately).

### What this PR does

Adds two tests + a keyword-only stub to `tests/test_puffo_core_tools.py`:

- **`_StubRpc`** — mirrors real `PuffoRpcClient` shape: `async def sync_mcp(self, *, template_id: str)`. Keyword-only, so any forward that names the arg something else raises `TypeError` here exactly as it does in production.
- **`test_sync_host_mcp_forwards_to_keyword_only_template_id`** — pins that whatever the agent-facing param ends up called, the internal forward passes `template_id=...` to `sync_mcp`.
- **`test_sync_host_mcp_raises_when_rpc_unavailable`** — non-regression: `cfg.rpc_client is None` raises `RuntimeError` with the `PUFFO_RPC_URL` guidance string (not `AttributeError` on None).

### Verification

Solution reverse-verified by injecting the exact `#200` bug pattern into `src/puffo_agent/mcp/puffo_core_tools.py`:

1. Rename agent-facing param to `name`
2. Change forward to `sync_mcp(name=name)`
3. Update test's `tool.fn` call to `name="linear"` (as a committer would after the rename)

Result: `test_sync_host_mcp_forwards_to_keyword_only_template_id` reds with the **exact production TypeError**:

```
E    TypeError: _StubRpc.sync_mcp() got an unexpected keyword argument 'name'
     at src/puffo_agent/mcp/puffo_core_tools.py:1238
     return await cfg.rpc_client.sync_mcp(name=name)
```

Not an assertion failure — the same TypeError shape that would ship to users. Restored intact.

### Numbers

Per Cal's [[test-count-baseline-compare]] recommendation to use `--junitxml` for reliable count (pytest `-q` summary can be eaten by warning blocks):

- Base (`origin/main` `431bb77`): 2377 tests, 0 failures, 0 errors (independently measured)
- Branch (`af2d311`): **2379 tests**, 0 failures, 0 errors, 2 skipped (permission mode + Windows-only proc test — both expected)
- Delta: **+2 tests**, no code changes

PySide6-collection-error ignore list derived at runtime per Cal's snippet (not enumerated — enumerated lists go stale, derived don't; same enumerate-vs-derive lesson from PUF-396):

```
IGN=$(grep -rlE "PySide6" tests/ | sed 's/^/--ignore=/' | tr '\n' ' ')
pytest $IGN --junitxml=<file> -q
```

### Why not just extract the UX rename too?

That's Vase's product decision. The `34a31df` rename (`sync_host_mcp(template_id)` → `sync_host_mcp(name)`) is a UX improvement stacked on top of `feat/slim-primer`; extracting it independently would be "publishing half a branch." If Vase wants the rename on main, single follow-up PUF for that rename explicitly.

**This PR ships the test, not the rename.** When/if the rename lands, this test is already there — a forward that forgets to translate the new param name to `template_id` will red immediately, not silently ship a `TypeError` to every caller.


Glimmer0x:
Closing as unnecessary after the Agent Foundation consolidation. The PR explicitly contains zero behavior change and guards a bug that is not on `main`; the tool now lives in `core_host_mcp_tools.py` and still forwards `template_id=template_id` to the keyword-only RPC. Given the current test-volume policy, carrying two extra tests against the pre-consolidation test module is not justified.

PR #228, opened 2026-08-11
PUF-398: drained agent state — quota-exhaustion honest surface (fixes JYP)

hellohanchen:
Linear: [PUF-281](https://linear.app/puffo-ai/issue/PUF-281) · Shared: `shared/PUF-398.md`

## Summary

Adds `drained` as the third red state on `runtime.health` (siblings: `auth_failed`, `refresh_broken`), for a spent Claude Code / Codex plan quota. Fixes the JYP mis-surface (Sam DM msg_3083b1b6) where quota exhaustion showed "sign-in has expired → run `claude auth login`", which does nothing for a spent quota.

Design proposal + Vase mechanism-approval preceded build (PUF-370 pattern). All 3 Vase answers on `msg_c05a7e9b` thread: (1) approve the 5-element design + retry shape; (2) `drained = hold-no-retry`, UI surface via existing `status_reporter`; (3) usage-limit verbatim strings sourced via web-search — 4 shipped Claude Code spellings anchored on stable substring cores, Codex uses structured `usedPercent`.

## Root cause was in two places, not one

Intake described the worker's classification path. **The refresher's own probe carries the same bug in a worse form**: `credential_refresh.py`'s probe hits the same quota, `looks_like_auth_error` substring-matches, and `_flip_auth_failed` flips *every* agent on the host to `auth_failed` + DMs each. JYP's DM likely came from here, not their own turn. Both sites now check quota before auth.

## Where detection actually lives

All sites share one marker module and check **quota before auth** (auth markers are bare substrings and a spent-quota body carries auth-adjacent wording):

- `agent/_usage_markers.py` (new) — `looks_like_usage_limit()` is a **plan-level** predicate that requires *positive* plan/account evidence: the shipped Claude/Codex plan spellings count unconditionally (they are plan-scoped by construction), while the generic quota vocabulary (`quota exceeded`, `insufficient_quota`) counts only when the same body explicitly scopes the limit to the account/plan/subscription/organization. Bare, model-scoped, or project-scoped generic quota stays the weaker `quota_exhausted` — absence of a model mention is not evidence of account scope, and a per-project ceiling is not the provider plan. Negative regressions cover model identifiers without the literal noun ("quota exceeded for gpt-5.4; switch to gpt-4.1"), bare markers, and project scope; a positive regression pins an account-scoped body that mentions a model incidentally. Plus `parse_reset_epoch()` (machine-readable `|<epoch>` spelling only) and the shared `DRAINED_RUNTIME_ERROR` copy.
- `agent/core.py` — `_classify_api_error()` returns `(is_auth, is_drained, label)`; all three `API Error` raise sites tag `AgentAPIError(is_drained=)`.
- `agent/provider_failures.py` — `classify_provider_failure` order is now hard-401 → quota → auth-substrings, so the 2.0.0 driver path (Codex JSON-RPC / Claude stream errors) classifies `quota_exhausted` instead of `authentication`.
- `portal/credential_refresh.py` — the probe's quota check runs before the auth check and returns a new `RefreshOutcome.QUOTA_EXHAUSTED`, which neither counts toward the `refresh_broken` streak nor fast-retries (a long drained window must not flip idle agents `refresh_broken`). Both host-wide flippers (`refresh_broken`, `auth_failed`) skip agents already in `drained`.
- `portal/control/usage_snapshot.py` — `apply_drained_health()` flips agents on a spent harness from the periodic usage snapshot (`used_pct >= 100`, weekly counts even with session headroom) and clears `drained → ok` when budget returns. For agents with a **live worker** the flip goes through the worker's own drained ENTER/CLEAR **in memory** (the daemon registers `set_live_workers(lambda: self.workers)` at startup): a disk-only write would be overwritten by the worker heartbeat within seconds and would never reach the status reporter, which reads memory. This also means an idle agent whose sibling burned the shared quota gets the per-episode operator DM, with the snapshot's `resets_at` in it. Agents without a running worker keep the on-disk path. A snapshot-driven clear deliberately does **not** re-arm the DM gate (only a successful turn does), so a flapping `used_pct` cannot spam the operator.

## Outcome plumbing

- `agent/_failure_outcomes.py` (new) — `failure_outcome()` + `crash_resume_terminal()`, the exception → process-outcome map (quota first), extracted from `global_inbox_runtime.py` to hold the file under the 2,000-line structural cap. Account-wide `drained` requires the narrow `plan_drained` error code; broad `quota_exhausted` (which also matches per-model limits, e.g. LiteLLM/Fable ceilings) keeps its pre-existing `provider_failed` semantics.
- `agent/global_inbox_runtime.py` + `agent/global_inbox_degraded.py` (new mixin) — in-turn retry gate excludes `is_drained`; `_handle_process_failure` and the crash-resume loop delegate to `_failure_outcomes`. A drained turn **parks** instead of entering degraded auto-recovery: durable rows stay pending, no provider turn is scheduled or admitted, `notify()` clears the degraded backoff but deliberately not the park, and the runtime unparks only when its `drained_check` (the worker's snapshot-driven health) reports the quota cleared **and** a wake arrives. Crash-resumed drained turns park the same way.
- `portal/worker_run.py` — `_settle_process_health()` (extracted static dispatch, bound via `functools.partial`) routes `outcome == "drained"` to the worker; `_retryable_local_warm_error` holds (no retry) on drained/quota shapes while keeping plain 429s retryable.
- `portal/worker.py` — `_enter_drained` (state flip + one bilingual operator DM per episode, re-armed by a successful turn or a failed send), `_clear_drained` on turn success.
- `portal/state.py`, `portal/cli.py` — enum widening; `drained` renders as a red state in `agent list`.
- `agent/_invite_strings.py` — `format_drained` / `format_codex_drained` bilingual DM. The copy makes no auto-resume promise: nothing re-kicks a held turn when the window resets, so it tells the operator to send the agent a message after the reset. It never suggests re-login.

## Where the DM's predicted reset time comes from

Three sources, in order:

1. `parse_reset_epoch` on the error text (the machine-readable `usage limit reached|<epoch>` spelling) — rare in practice, since production paths deliver the mapped provider-failure message rather than the raw body.
2. The snapshot path passes the harness's reset straight into the worker's drained ENTER. The reset is the time at which **every** exhausted window has cleared (their max) — reporting the soonest reset would promise capacity while another window is still spent — and `None` when any exhausted window has no known reset.
3. When neither produced a time, `_notify_operator_of_drained` probes the `/usage` budget on the spot (`predicted_reset_epoch(harness)`) and predicts the reset from it. Best-effort: a failed or empty probe degrades to a time-less DM — the DM matters more than its timestamp.

## Tests

`tests/test_drained_state.py` (new): detection (all 4 shipped Claude spellings, agent-prose negatives, case-insensitivity, Fable5/LiteLLM per-model-limit negative, short-lived-429 negative), ordering at every site (provider classifier, core, refresher — each fed a body matching BOTH quota and auth classes), outcome dispatch (`failure_outcome`, `crash_resume_terminal` both exception shapes, settle routing incl. full outcome→health dispatch and swallowed-raise), warm retry boundary, worker enter/clear/DM dedup + re-arm semantics, snapshot-driven flip/clear/skip for both live-worker (in-memory, DM through the episode gate, other reds untouched, broken-worker resilience) and on-disk agents + malformed-input resilience, the live-worker registry fallback, the reset-time probe (predicts, degrades on failure), refresher streak isolation, and DM copy (never suggests login, no auto-resume promise, bilingual, reset-time degrade, empty display name).

`tests/test_worker_status_lifecycle.py`: `quota_drained` lifecycle case pins `process_outcome == "drained"` through the real runtime; the pre-existing provider-failure case now uses `provider_unavailable` so the two stay distinct.

Verification on the current head (Python 3.14 locally; CI runs 3.11/3.12):

```
tools/check_python_structure.py     → passed
ruff 0.15.1 (hook-pinned)           → All checks passed!
full suite                          → see CI; local run green, 1 skip (Windows-only test_proc)
diff coverage (line + branch, vs merge base 343bda5) → 100% of new executable code
```

## Out of scope (per spec — untouched)

- Fable 5 / LiteLLM per-model usage-limit — non-regression pinned by `test_fable5_model_limit_is_not_quota_classified`
- Auto-model-switch on drained
- Auto-credits-refill / auto-upgrade (billing)
- Cross-owner drained aggregation UI
- Re-kicking held turns when the usage window resets (unpark = snapshot-cleared health + the next wake; the DM tells the operator to send a message)

## Not tested empirically

No real drained account was available. All detection tests feed the 4 web-searched shipped Claude Code strings + constructed Codex snapshots. If a live drained user's verbatim output diverges, adding one more anchor to `USAGE_LIMIT_MARKERS` is the fix; the scaffold is stable.

## Server side

`health` already reaches the server on **both** transports, so `drained` needs no server-side change (PUF-428 / puffo-server#274):

- Native: `POST /agents/me/heartbeat` accepts, validates, persists, and returns free-text `health`; the daemon has been sending it since PUF-270.
- Bridge/keyless: `AgentClientMsg::Status` carries `health: Option<String>` and `handle_status` folds it into the same `agent_status` row. What was missing was the **agent side**: `CloudBridgeClient.send_status()` had no health field and the worker passed `runtime_health_provider=None` for bridge clients. Both are fixed here — the keyless status emit now feeds the worker's health (drained included) onto the wire, with wire-level regressions (`test_status_frame_carries_health_on_the_wire`, reporter emit, and the reconnect-callback payload).


Glimmer0x (changes requested):
Reviewed current head `5e8cbda7555c1b0d29046248f971f5f4edb15733`. The classification/order plumbing is broad and the focused suite is strong, but I found two blocking semantic gaps plus one reset-time correctness issue.

1. **Blocking — `drained` still auto-retries forever.** `_invoke_turn_with_retries` correctly excludes `is_drained` from the same-turn retry budget, but the exception then reaches `_handle_process_failure`, which unconditionally calls `_degrade()`. `_degrade()` schedules a coalescer wake after 5s, doubles to the 300s cap, and `_try_degraded_recovery()` explicitly retries the durable work; the requeued rows remain pending and `_wake_remaining_pending()` also runs. So a spent account is retried indefinitely (at most every five minutes) instead of being held, while the operator DM says “holding … rather than retrying.” Please add a real parked/drained gate for the Global Inbox runtime: retain the durable rows but do not schedule or admit provider turns while drained, and define the explicit unpark signal (for the current design, a confirmed usage-snapshot clear followed by/new inbound wake). Add an end-to-end runtime regression proving one drained provider attempt remains one across degraded/coalescer wakes and messages stay pending until the gate clears.

2. **Blocking — the 2.0 provider path drains on per-model limits that the PR declares out of scope.** `failure_outcome()` maps every `ProviderFailureError(error_code="quota_exhausted")` to `drained`, but `classify_provider_failure()` still has the broad pre-existing `("reached your" in diagnostic and "limit" in diagnostic)` rule. I reproduced: `classify_provider_failure(status=None, diagnostic="You've reached your Fable 5 limit. Try again later.") == "quota_exhausted"`; the driver wraps that as `ProviderFailureError`, and this PR now turns it into `drained`. The current Fable negative test covers only `looks_like_usage_limit` / `_classify_api_error`, not this load-bearing driver path. Please use a distinct/narrow signal for account/plan exhaustion (or otherwise preserve the per-model boundary), and add a classifier → exception → `failure_outcome` regression.

3. **Reset-time correctness.** `drained_harnesses()` returns `min(resets)` across exhausted session/weekly windows. If both are binding, capacity returns only after the later reset, so the DM can tell the operator to retry while weekly quota is still spent. If any exhausted window has no reset timestamp, the overall recovery time is unknown rather than the earliest known timestamp. Please compute the time at which all binding exhausted windows clear (normally `max` when all are known, otherwise `None`) or change the copy so it does not promise overall recovery at that time; add the dual-window regression.

Verification on this head: `tests/test_drained_state.py` + `tests/test_worker_status_lifecycle.py` pass (`99 passed`), and local `pre-commit run --all-files` passes. GitHub pytest/pre-commit/CodeQL are also green; the findings are behavioral gaps not caught by the current tests.

Glimmer0x:
Supplemental finding on current head `5e8cbda` after tracing the user-visible status path:

**The `drained` health value is dropped for keyless/bridge agents.** The native HTTP heartbeat path adds `body["health"] = runtime_health_provider()`, so those agents can surface `health="drained"` separately from the coarse `idle|busy|error` status. But `Worker._build_status_reporter()` deliberately passes `runtime_health_provider=None` whenever a bridge exists; `_emit_keyless()` calls `bridge.send_status()` without a health argument; and both `StatusReporter._runtime_payload()` and `BridgeClient.send_status()` whitelist runtime fields without `health`. The bridge wire therefore carries only the coarse status plus model metadata, not `drained`.

That leaves the PR description’s “UI surface via existing `status_reporter`” incomplete for bridge-backed agents, and it matches the original user-facing concern: internal `runtime.health` can be correct while the user sees only a generic error/status. Please add a typed `health` field to the bridge status frame (with matching server support if it is not already accepted) and a keyless status regression proving `drained` reaches the wire. If bridge-backed CLI agents are intentionally out of scope, narrow the PR claim and document the unsupported runtime explicitly instead.

Glimmer0x (changes requested):
Reviewed current head `d20391fba5019daf7c67e2d880786c721f5f7145`. The new reset-time probe, live-worker snapshot flip, and refresher outcome improve the implementation, but the four earlier blockers are still present.

1. **Blocking — a drained Global Inbox turn still enters degraded auto-recovery instead of parking.** `failure_outcome(exc)` is computed at `global_inbox_runtime.py:1369`, but `_handle_process_failure()` then unconditionally calls `_degrade()` at lines 1375–1379. `_degrade()` schedules the coalescer wake, and `_try_degraded_recovery()` admits the durable work after the deadline. I added a review-only regression using a real `GlobalInboxRuntime`, one pending receipt, and `AgentAPIError(..., is_drained=True)`: the first call requeued the row; after making the scheduled recovery due, `process_once()` called the provider a second time (`calls == 2`, expected `1`). Also, `notify()` clears `_degraded` unconditionally, so a new inbound message can immediately retry before any confirmed quota-clear. Please add a distinct parked/drained admission gate: keep rows pending, do not schedule/admit provider turns while parked, and unpark only from the chosen explicit signal (for the current design, confirmed usage-snapshot clear, then a/new inbound wake). Add the end-to-end regression across coalescer and inbound wakes.

2. **Blocking — the classifier still converts per-model limits into total-account `drained`.** `provider_failures.py:135–143` retains the broad `"reached your" + "limit"` rule. On this head, `"You've reached your Fable 5 limit. Try again later."` still produces `classifier=quota_exhausted`, then `ProviderFailureError(..., error_code=...)` produces `failure_outcome=drained`. The existing Fable negative test only exercises `looks_like_usage_limit`, not this classifier → exception → outcome path. Please narrow/distinguish total plan exhaustion and add that full-path regression.

3. **Blocking — reset time still reports the earliest known binding reset.** `drained_harnesses()` documents “soonest reset” and returns `min(resets)` at `usage_snapshot.py:141–160`. With session reset 100 and weekly reset 200 it returns 100, although capacity remains unavailable until 200; with one exhausted window lacking a timestamp and another at 200 it returns 200, although the overall recovery time is unknown. Please use the time when all binding exhausted windows clear (`max` when every binding reset is known, otherwise `None`) or remove the promise that this is the recovery time, with both regressions.

4. **Blocking — bridge/keyless status still drops `health="drained"`.** `Worker._build_status_reporter()` explicitly passes `runtime_health_provider=None` when a bridge exists (`worker.py:1062–1068`). `_emit_keyless()` sends only status/error/runtime, and `CloudBridgeClient.send_status()` has no health argument or frame field (`bridge_client.py:615–651`). Routing snapshot flips through the live Worker keeps local state correct, but does not put health on the bridge wire. Please add the typed bridge field plus receiving-server support and a wire-level regression, or explicitly narrow the PR scope/claim if bridge-backed agents are unsupported.

Verification: the shipped focused suites pass (`tests/test_drained_state.py` + `tests/test_worker_status_lifecycle.py`: 99 passed), and GitHub CI/CodeQL are green. The new parked-runtime regression fails exactly as described, so this remains a semantic coverage gap rather than a general test failure.

Glimmer0x (changes requested):
Re-reviewed head `7fedb0b`. The drained park, latest/unknown reset aggregation, and bridge `health` propagation now close the corresponding prior findings; the focused affected test set and `tools/check_python_structure.py` pass.

One blocking classifier boundary remains:

`looks_like_usage_limit()` still treats the generic substrings `quota exceeded` and `insufficient_quota` as whole-plan markers (`agent/_usage_markers.py:20-21`), and `classify_provider_failure()` maps every match to `plan_drained` (`provider_failures.py:142-143`). Therefore an explicitly per-model diagnostic such as `quota exceeded for model Fable 5; other models remain available` still follows `plan_drained -> drained` and parks the entire account. The legacy/core path has the same issue: `_classify_api_error("API Error: quota exceeded for model Fable 5; other models available")` returns `(False, True, "quota-drained")`. `insufficient_quota for model ...` reproduces identically.

Please separate a genuinely plan/account-level drained predicate from generic quota vocabulary, and add negative tests for these two generic markers when the diagnostic explicitly scopes the limit to one model (at least through the provider classifier/outcome chain and the core API-error path). The current new test asserting `quota exceeded for this project` is plan-drained demonstrates that the predicate is intentionally broad, but it does not guard the model-scoped counterexample.

Glimmer0x (changes requested):
Re-reviewed head `ff2a357`. The two requested end-to-end negative cases now pass, but the remaining classifier boundary is still blocking because the implementation tests for the token `model`, not for model scope.

`_MODEL_SCOPED_RE = r"\bmodels?\b"` makes generic quota text plan-drained whenever that token is absent. On this head:

- `quota exceeded for gpt-5.4; switch to gpt-4.1` -> `looks_like_usage_limit=True`, `plan_drained`
- `gpt-5.4 quota exceeded; other providers remain available` -> `True`, `plan_drained`

Both are explicitly scoped to one model but do not spell the noun `model`, so the whole account is still parked. The reverse also fails: `The model returned an error: quota exceeded for this account` contains the token incidentally and becomes `quota_exhausted` instead of plan-drained.

The code comment also says generic markers are emitted for per-project ceilings, yet `quota exceeded for this project` remains asserted as plan-drained. More fundamentally, absence of the word `model` is not positive evidence of account/plan scope, so bare `quota exceeded` / `insufficient_quota` are still being promoted from ambiguous generic vocabulary to the strongest whole-account state.

Please make `plan_drained` require positive plan/account evidence (the existing `PLAN_LIMIT_MARKERS`, structured provider metadata, or explicit scoped wording). Keep ambiguous generic quota vocabulary as `quota_exhausted` unless a plan/account signal is present. Add regressions for model identifiers without the literal noun, and for an account-scoped message with an incidental model mention; clarify the intended project-scope behavior as well.

Verification: `tests/test_drained_state.py` passes (96 tests), `tools/check_python_structure.py` passes, local `pre-commit run --all-files` passes, and GitHub CI/CodeQL are green. This is a semantic boundary gap not caught by the new examples.

Glimmer0x (changes requested):
The prior blocker is improved but not fully closed.

`looks_like_usage_limit()` now promotes a generic quota marker whenever any account/plan noun appears anywhere in the diagnostic. That is still token co-presence, not evidence that the marker is scoped to the plan. For example, all of these currently return `plan_drained`:

- `quota exceeded for model gpt-5.4; account quota remains available`
- `project quota exceeded; organization quota is unaffected`
- `quota exceeded for this project; your subscription is still active`

Those diagnostics contain positive or negative information about the account but explicitly scope the exhausted quota elsewhere. Draining the whole provider would still park an agent whose account has usable capacity. Please make the positive evidence relational (for example, match known phrases that bind `quota exceeded` / `insufficient_quota` to account, plan, subscription, or organization scope) and add at least one contradictory-scope regression test.

The new bare/model/project tests and the targeted drained-state suite pass; CI and CodeQL are also green at `1c2e1ad`.

Glimmer0x (approved):
Re-reviewed at 6891fd7. Generic quota markers now require relational binding to account/plan/subscription/organization scope, so contradictory diagnostics such as model/project quota exhaustion with account or organization quota still available remain quota_exhausted rather than plan_drained. Added positive binding and contradictory-scope regressions. Targeted local suite passed; hosted pre-commit, Python 3.11/3.12, and CodeQL checks are green.

PR #233, opened 2026-08-12
Agent Foundation 2.0: restore runtime compatibility and approval flows

Glimmer0x:
## Why

Agent Foundation landed in #225, but the stable-readiness audit found locally actionable compatibility gaps around existing installations, Docker Codex, Windows Claude launch behavior, runtime status/telemetry, keyless operator authorization, and shared workspaces.

This PR closes those code-level gaps in the `puffo-agent==2.0.0a3` candidate before staging validation. It does **not** authorize a stable `2.0.0` release.

## System shape

```mermaid
flowchart LR
    State[Existing Agent state] --> Daemon[Daemon reconciliation]
    Daemon --> Driver[Claude or Codex Driver]
    Driver --> Inbox[Global Inbox]
    Inbox --> Send[Send Coordinator]
    Driver --> Events[Bounded Runtime Events]

    Bridge[Keyless bridge] --> Approval[Operator DM and invitation approval]
    Approval --> Inbox

    Daemon --> Private[Per-Agent private workspace]
    Private --> Shared[shared symlink or Docker mount]
    Shared --> Root[Puffo-home shared directory]
```

## What changed

- Restored runtime compatibility after #225:
  - Docker Codex Driver support and container lifecycle.
  - Windows Claude Code shim launch handling.
  - Legacy daemon stop-request behavior.
  - Global Inbox status lifecycle and local runtime-event status projection.
  - Safe telemetry and optional context-baseline compatibility.
- Added durable, resumable keyless operator authorization:
  - Foreign DM allow/block decisions.
  - Space invitation approve/reject decisions.
  - Restart/replay-safe local persistence and bridge ingress.
- Unified the workspace contract:
  - `<puffo-home>/agents/<agent-id>/workspace` remains private.
  - `workspace/shared` exposes `<puffo-home>/shared` through a local symlink/junction or Docker bind mount.
  - Existing Agents are repaired lazily by daemon reconciliation.
  - Existing real `workspace/shared` content is preserved and surfaced as a conflict rather than overwritten or hidden.
  - Docker keeps `/workspace/.shared` as a compatibility alias.
  - Attachment validation permits only the managed shared root, not arbitrary symlink escapes.
- Addressed review findings around workspace status propagation, Docker mount transitions, crash recovery, keyless error detail, held-baseline compatibility, and baseline-persistence observability.
- Added the source audit, compatibility matrix, and stable-readiness backlog.

## Compatibility and scope

- Existing `agent.yml`, profile, memory, workspace, keys, message history, and supported session references remain in place.
- Channel encryption/plaintext policy remains outside this PR and stays owned by the separate Server/Agent workstream.
- Stable publishing remains blocked on the external Windows, staging bridge, Runtime Event, production inventory, cloud-template, and final candidate evidence documented in [STABLE-2.0-READINESS.md](https://github.com/puffo-ai/puffo-agent/blob/feat/agent-foundation-stable-readiness/roadmap/agent-foundation/STABLE-2.0-READINESS.md).

## Validation

Current head `e4791e6`, package `puffo-agent==2.0.0a3`:

- Full local pytest suite passed with one expected Windows-only skip.
- `pre-commit run --all-files` passed, including Ruff and Python structural limits.
- `git diff --check` passed.
- Wheel and sdist builds passed; artifacts report `2.0.0a3`.
- GitHub CI passed on Python 3.11 and 3.12; CodeQL passed.

Real Windows execution and staging bridge acceptance remain external evidence and are not represented as locally completed tests.

## Review focus

1. Upgrade and existing-install compatibility.
2. Claude/Codex local and Docker runtime boundaries.
3. Keyless DM/invitation authorization lifecycle and replay.
4. Runtime status/telemetry privacy.
5. Managed shared-workspace safety and migration behavior.
6. Alpha-candidate and stable-release gate accuracy.

### DIFF
--- a/src/puffo_agent/agent/global_inbox_runtime.py
+++ b/src/puffo_agent/agent/global_inbox_runtime.py
@@ -16,7 +16,7 @@ import uuid
 from dataclasses import asdict, replace
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Awaitable, Callable, Sequence
+from typing import Any, Awaitable, Callable, Protocol, Sequence
 
 from .context_controller import (
     AdmissionCandidate,
@@ -56,7 +56,7 @@ DEGRADED_RECOVERY_MAX_SECONDS = 300.0
 
 from .global_inbox_held import HeldRecoverySource
 from .global_inbox_send import TrackingSendDelegate
-from .global_inbox_admission import InboxAdmissionMixin
+from .global_inbox_admission import InboxAdmissionMixin, InboxReadContext
 from .global_inbox_types import (
     ActiveBoundaryAdapter,
     ActiveExactUnion,
@@ -96,6 +96,7 @@ __all__ = [
     "SendAttemptState",
     "TrackingSendDelegate",
     "ToolResultAdmission",
+    "TurnStatusLifecycle",
     "await_listener_with_runtime",
     "conservative_token_estimate",
     "format_stored_message",
@@ -105,6 +106,27 @@ __all__ = [
 
 TurnRunner = Callable[[PlannedTurn], Awaitable[Any]]
 UnfitPolicy = Callable[..., bool | Awaitable[bool]]
+
+
+class TurnStatusLifecycle(Protocol):
+    """Optional worker-owned mirror of the active Global Inbox turn.
+
+    The runtime owns only the durable turn lifecycle; it notifies this
+    callback with immutable active-union snapshots so the worker can drive
+    Server processing-row status without moving status policy in here.
+    """
+
+    async def on_turn_active(self, message_ids: tuple[str, ...]) -> None:
+        """Report that the active union is (or has grown to) ``message_ids``."""
+
+    async def on_turn_terminal(
+        self,
+        *,
+        message_ids: tuple[str, ...],
+        succeeded: bool,
+        error_text: str | None,
+    ) -> None:
+        """Settle the exact active union in one terminal status batch."""
 
 
 class GlobalInboxRuntime(InboxAdmissionMixin):
@@ -133,6 +155,7 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
         notice_delivery: InboxNoticeDelivery | None = None,
         runtime_event_outbox: Any | None = None,
         reminder_scheduler: ReminderScheduler | None = None,
+        status_lifecycle: TurnStatusLifecycle | None = None,
     ) -> None:
         self.store = store
         self.adapter = adapter
@@ -179,6 +202,7 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
             store=store,
             notify=self.notify,
         )
+        self.status_lifecycle = status_lifecycle
         self.send_delegate: TrackingSendDelegate | None = None
         self.held_recovery_source = HeldRecoverySource(
             self,
@@ -700,6 +724,7 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
         self.active.routes[:] = list(planned.routes)
         if self.coordinator is not None:
             self.coordinator.provider_session_id = event.provider_session_id
+        await self._notify_status_active()
         log_runtime_event(
             logger,
             "turn.admitted",
@@ -725,6 +750,98 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
                 notice_generation=planned.notice_generation,
                 outcome="admitted",
             )
+
+    async def _notify_status_active(self) -> None:
+        """Expose the exact active union to the worker's status lifecycle.
+
+        Telemetry-only: a reporter failure must never break admission or
+        processing-state decisions, so the callback is shielded here.
+        """
+        if self.status_lifecycle is None or not self.active.message_ids:
+            return
+        try:
+            await self.status_lifecycle.on_turn_active(
+                tuple(self.active.message_ids)
+            )
+        except Exception:
+            logger.warning(
+                "agent %s: status lifecycle admission notify failed",
+                self.agent_id,
+                exc_info=True,
+            )
+
+    async def _notify_status_terminal(
+        self,
+        *,
+        terminal: bool,
+        succeeded: bool,
+        error_text: str | None,
+    ) -> None:
+        """Settle the exact active union before ``_finalize_process`` clears it.
+
+        Only a terminal turn settles status; the batch is built from the same
+        immutable snapshot the durable requeue/process just used.
+        """
+        if self.status_lifecycle is None or not terminal:
+            return
+        try:
+            await self.status_lifecycle.on_turn_terminal(
+                message_ids=tuple(self.active.message_ids),
+                succeeded=succeeded,
+                error_text=error_text,
+            )
+        except Exception:
+            logger.warning(
+                "agent %s: status lifecycle terminal notify failed",
+                self.agent_id,
+                exc_info=True,
+            )
+
+    async def _admit_inbox_read(
+        self,
+        context: InboxReadContext,
+        event: ProviderAdmissionEvent,
+        fired: list[bool],
+    ) -> None:
+        """Admit a mid-turn Inbox read, then expose the grown active union.
+
+        The durable admission and its processing-state decisions stay in the
+        implementation trait; this override only mirrors the post-admission
+        union into the worker's status lifecycle.
+        """
+        await super()._admit_inbox_read(context, event, fired)
+        await self._notify_status_active()
+
+    async def _admit_held_recovery(
+        self,
+        event: ProviderAdmissionEvent,
+        *,
+        fired: list[bool],
+        correlation_key: str,
+        active_turn_id: str,
+        provider_session_id: str,
+        evidence: Any,
+        evidence_key: tuple[str, str, str, str, int, str],
+        displayed_ids: tuple[str, ...],
+        space_id: str,
+        channel_id: str,
+        latest_seq: int,
+    ) -> None:
+        """Admit a held-send recovery, then expose the grown active union."""
+        await super()._admit_held_recovery(
+            event,
+            fired=fired,
+            correlation_key=correlation_key,
+            active_turn_id=active_turn_id,
+            provider_session_id=provider_session_id,
+            evidence=evidence,
+            evidence_key=evidence_key,
+            displayed_ids=displayed_ids,
+            space_id=space_id,
+            channel_id=channel_id,
+            latest_seq=latest_seq,
+        )
+        await self._notify_status_active()
 
     def _write_current_turn(self, planned: PlannedTurn) -> None:
         path = self.current_turn_path
@@ -1099,21 +1216,26 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
                 or any(item.is_encrypted for item in planned.items),
             )
             terminal = False
+            terminal_succeeded = False
+            terminal_error: str | None = None
             try:
                 await self._invoke_turn_with_retries(planned)
                 if self.active.turn_id == planned.turn_id:
                     await self._mark_active_processed(planned, process_started)
                     terminal = True
+                    terminal_succeeded = True
                 else:
                     self._degrade("provider returned without correlated admission")
             except asyncio.CancelledError:
                 await self._requeue_active_turn(planned, process_started, "cancelled")
                 terminal = True
+                terminal_error = "global inbox turn cancelled before completion"
                 raise
             except Exception as exc:
                 terminal = await self._requeue_active_turn(
                     planned, process_started, "provider_error"
                 )
+                terminal_error = f"{type(exc).__name__}: {exc}"
                 diagnostic = (
                     "turn failed and was requeued"
                     if terminal
@@ -1135,6 +1257,11 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
                     outcome="requeued" if terminal else "degraded",
                 )
             finally:
+                await self._notify_status_terminal(
+                    terminal=terminal,
+                    succeeded=terminal_succeeded,
+                    error_text=terminal_error,
+                )
                 self._finalize_process(planned, terminal)
             await self._wake_remaining_pending()
             return True
--- a/src/puffo_agent/agent/global_inbox_runtime.py
+++ b/src/puffo_agent/agent/global_inbox_runtime.py
@@ -116,12 +116,15 @@ class TurnStatusLifecycle(Protocol):
     Server processing-row status without moving status policy in here.
     """
 
-    async def on_turn_active(self, message_ids: tuple[str, ...]) -> None:
+    async def on_turn_active(
+        self, *, turn_id: str, message_ids: tuple[str, ...]
+    ) -> None:
         """Report that the active union is (or has grown to) ``message_ids``."""
 
     async def on_turn_terminal(
         self,
         *,
+        turn_id: str,
         message_ids: tuple[str, ...],
         succeeded: bool,
         error_text: str | None,
@@ -382,6 +385,12 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
                 mode="startup_orphan_recovery",
                 message_count=len(run.message_ids),
                 outcome="requeued",
+            )
+            await self._settle_recovered_status(
+                turn_id=run.turn_id,
+                message_ids=tuple(run.message_ids),
+                succeeded=False,
+                error_text="orphaned durable turn requeued at startup",
             )
             recovered += 1
         return recovered
@@ -757,11 +766,16 @@ class GlobalInboxRuntime(InboxAdmissionMixin):
         Telemetry-only: a reporter failure must never break admission or
         processing-state decisions, so the callback is shielded here.
         """
-        if self.status_lifecycle is None or not self.active.message_ids:
+        if (
+            self.status_lifecycle is None
+            or not self.active.turn_id
+            or not self.active.message_ids
+        ):
             return
         try:
             await self.status_lifecycle.on_turn_active(
-                tuple(self.active.message_ids)
+                turn_id=self.active.turn_id,
+                message_ids=tuple(self.active.message_ids),
             )
         except Exception:
             logger.warning(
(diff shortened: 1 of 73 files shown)


hellohanchen (commented):
Reviewed at `e8a7e5a`. Read the new modules in full (`workspace_layout`, the three keyless files, `docker_runtime`) and traced the changed protocol paths in `send_coordinator` / `send_response_validation` rather than reading the diff top to bottom.

**What I checked and found sound**, since it's the part most worth getting wrong:

- **Attachment validation.** I walked the escape cases against `_validate_attachment_targets`: an agent-created symlink at `workspace/shared/evil → /etc/passwd` fails both `is_relative_to` checks; replacing `workspace/shared` itself with a symlink to `/etc` fails against the canonical `shared_workspace`; and the lexical `parts[0] == "shared"` requirement plus the resolved containment check have to hold together. The separate `shared_workspace` config rather than widening the workspace root is the right shape.
- **Keyless authorization.** Both gates require `payload.envelope_kind == "dm" and payload.sender_slug == client.operator_slug` before any decision is recognised, and the redelivery resume path in `_reschedule_sequence_less_control_reply` inherits that check implicitly because it dispatches off the stored `receipt_reason`, which only a passing gate could have written. `parse_keyless_invitation` is strict and fails closed on every field.
- **Container vs host path wiring.** `PUFFO_SHARED_WORKSPACE` is `/workspace/shared` on both Docker builders and `shared_fs_dir()` for local and in-process `ws_local` — I checked each of the four rather than assuming, since getting it backwards would silently reject every shared-dir attachment.

---

### 1. `ensure_workspace_shared_link` returns a status nobody reads, while the primer promises the link exists

The function distinguishes `created` / `existing` / `conflict` / `unavailable` — that care is the right instinct. But all four call sites discard it:

- `portal/worker_run.py:187`
- `portal/daemon.py:291`
- `portal/cli.py:539`
- `agent/harness/docker_runtime.py:449`

Meanwhile this PR rewrites the primer to state unconditionally: "`shared/` inside it is the host-wide collaboration directory for Agents managed by the same Puffo home."

So on `conflict` — an existing agent that already has real content at `workspace/shared`, which is precisely the pre-2.0 installation this PR exists to migrate — the agent is told it has a shared directory, writes handoff files into what is actually its own private directory, and no other agent ever sees them. The only signal is a `logger.warning` in the daemon log. `unavailable` is the same failure with a different cause, and it's the likely Windows outcome: no symlink privilege plus a `mklink /J` that can't span the volumes involved. Windows is already called out as an open risk in this PR, and this is a silent-wrong-answer path rather than a visible one.

Branching the primer text on the returned status (and surfacing conflict/unavailable to the operator) turns a silent data-isolation bug into a legible one.

### 2. The repair path the migration depends on isn't tested

`tests/test_workspace_layout.py` has one test, covering `created` / `existing` / `conflict` and content preservation. The two branches that carry the "existing Agents are repaired lazily by daemon reconciliation" claim are untested: a **stale symlink pointing somewhere else** (the docstring's "safe to replace because the link itself owns no content") and a **dangling symlink** (`link.exists()` is False through a broken link, so it falls to the `is_symlink()` branch — correct, but only by ordering). Those two are the whole migration story for installed agents.

### 3. Docker creates a host symlink at the exact path it then bind-mounts

`_start_container` calls `ensure_workspace_shared_link(self.workspace_dir, self.shared_fs_dir)` and then runs `docker run` with `-v {workspace_dir}:/workspace` plus `-v {shared_fs_dir}:/workspace/shared`. The mount destination is inside the first bind mount, and on the host that path is now a **relative** symlink (`../../shared`) created moments earlier. That relative target resolves to a different place inside the container than on the host — from `/workspace/shared` it lands at `/shared`, not at the host's shared root.

I can't exercise Docker here, so I'm raising it as a question rather than a defect: was a container start exercised against a workspace where the host symlink already existed (i.e. an agent that ran locally first, then switched to Docker)? `tests/test_docker_codex_driver.py` only asserts that the two `-v` strings appear in the argv. The defensive version is to skip the host-side link when the runtime is Docker — the two mounts already provide `/workspace/shared` and the `.shared` alias, so the symlink buys nothing there.

### 4. The held-response baseline check was narrowed while the sent paths were widened

`send_response_validation.py:209` now requires exact equality:

```py
and raw.get("context_baseline_seq") == freshness["context_baseline_seq"]
```

where it previously accepted `None` **or** equality. In the same commit the sent and keyless-sent paths went the other way — they now tolerate a Server-established baseline when the request baseline is `None`.

Two consequences worth confirming as intended: for a null-baseline request the sent path accepts an int echo while the held path requires `None`, so a Server that echoes its established baseline consistently on both would fail held validation; and an older Server that simply omits the field on held responses now fails where it used to pass. In a PR whose stated purpose is restoring compatibility, a narrowing in the tolerated response set deserves an explicit note.

### 5. `persist_baseline` failure quietly weakens the freshness boundary

`persist_baseline` (`send_response_validation.py:70`) logs and swallows everything. Each failure leaves the next send as another null-baseline commit, where the Server establishes the baseline at its latest committed seq — re-acknowledging messages this turn never saw. That's the inverse of what the rest of the module works hard to enforce, and nothing retries or escalates: a persistently failing `set_context_baseline_seq` degrades silently and indefinitely. Worth at least a counter or a one-time louder warning.

### Nits

- `save_keyless_invitations`'s docstring says "a crash mid-write never leaves a truncated document". `os.replace` is atomic with respect to the rename, but there's no `fsync` of the temp file or the directory, so a power loss can still surface a short file. The failure mode is safe — a malformed row is skipped and the operator is re-prompted — so I'd fix the claim rather than add the fsync, unless you want the stronger guarantee.
- The two Docker MCP env builders disagree in style: `_configure_docker_mcp` passes a **host** `workspace` and relies on `adapters/docker_cli.py:462` to rewrite it container-local, while `_container_puffo_mcp_env` passes container paths directly. `PUFFO_SHARED_WORKSPACE` isn't in that fixup list — it's correct today only because the hardcoded value already happens to be container-local. A future edit that mirrors the `workspace=` style would break silently.
- `os.path.relpath` in `ensure_workspace_shared_link` is lexical, so if any component of `workspace` is itself a symlink the relative target resolves against the physical parent. Not reachable through the daemon's own layout today; noting it since the function is now the single shared entry point.

Reviewed the code as written; I have not executed the suites, so the 159- and 21-case slices are yours to stand behind.

src/puffo_agent/portal/workspace_layout.py
```diff
@@ -0,0 +1,79 @@
...
+            text=True,
+            check=False,
+            creationflags=getattr(subprocess, "CREATE_NO_WINDOW", 0),
+        )
+    except OSError:
+        return False
+    return completed.returncode == 0
+
+
+def ensure_workspace_shared_link(workspace: Path, shared_root: Path) -> str:
```
hellohanchen, line 39: The four-value status is the right design, but every caller drops it — `worker_run.py:187`, `daemon.py:291`, `cli.py:539`, `docker_runtime.py:449` all call this as a bare statement.

Meanwhile this PR rewrites the primer to state unconditionally that "`shared/` inside it is the host-wide collaboration directory". On `conflict` — an existing agent with real content at `workspace/shared`, exactly the installation this migration targets — the agent believes it has a shared directory, writes handoffs into its own private one, and no other agent sees them. `unavailable` is the same silent failure and is the likely Windows outcome when symlink privilege is absent and `mklink /J` can't span the volumes.

Branching the primer text on the status, and surfacing conflict/unavailable to the operator, would make this legible instead of silent.

src/puffo_agent/agent/harness/docker_runtime.py
```diff
@@ -0,0 +1,552 @@
...
+    def _write_layout_marker(self) -> None:
+        marker = self.agent_home / _DOCKER_LAYOUT_MARKER
+        try:
+            marker.parent.mkdir(parents=True, exist_ok=True)
+            marker.write_text(_CODEX_LAYOUT_MARKER + "\n", encoding="utf-8")
+        except OSError:
+            pass
+
+    async def _start_container(self) -> None:
+        ensure_workspace_shared_link(self.workspace_dir, self.shared_fs_dir)
```
hellohanchen: This creates a host-side **relative** symlink at `<workspace>/shared` (`../../shared`) immediately before `docker run` bind-mounts `-v {shared_fs_dir}:/workspace/shared` — i.e. the mount destination is a symlink inside the `-v {workspace_dir}:/workspace` mount, and that relative target resolves to `/shared` inside the container rather than to the host's shared root.

I can't exercise Docker here, so this is a question: was a container start tested against a workspace where the host symlink already existed (an agent that ran locally first, then switched to Docker)? `test_docker_codex_driver.py` only asserts the two `-v` strings are present in argv.

Skipping the host link for the Docker runtime is the defensive option — the mounts already provide `/workspace/shared` and the `.shared` alias.

src/puffo_agent/agent/send_response_validation.py
```diff
@@ -164,10 +206,7 @@ def _keyless_held_result(
     has_blocker = any(key in raw for key in _BLOCKING_HELD_RESPONSE_FIELDS)
     valid = (
         raw.get("seen_seq") == freshness["seen_seq"]
-        and (
-            raw.get("context_baseline_seq") is None
-            or raw.get("context_baseline_seq") == freshness["context_baseline_seq"]
-        )
+        and raw.get("context_baseline_seq") == freshness["context_baseline_seq"]
```
hellohanchen, line 228: This tightened from "`None` **or** equal" to strict equality, in the same commit where the sent and keyless-sent paths were *widened* to accept a Server-established baseline for a null-baseline request.

So for the same null-baseline request the sent path tolerates an int echo while the held path requires `None` — a Server that echoes its established baseline on both would fail held validation — and an older Server that omits the field on held responses now fails where it previously passed. Given this PR is specifically about restoring compatibility, worth confirming the narrowing is deliberate and matches the deployed Server.

src/puffo_agent/agent/send_response_validation.py
```diff
@@ -63,6 +67,25 @@ def optional_response_int(raw: Any, name: str, prefix: str) -> int | None:
     return value
 
 
+async def persist_baseline(
```
hellohanchen, line 74: This logs and swallows every failure. Each one leaves the next send as another null-baseline commit, where the Server establishes the baseline at its latest committed seq — re-acknowledging messages the turn never saw. A persistently failing `set_context_baseline_seq` therefore degrades the freshness boundary silently and indefinitely, with nothing retrying or escalating. A counter, or a one-time louder warning, would keep it from being invisible.

src/puffo_agent/agent/keyless_invitation_state.py
```diff
@@ -0,0 +1,201 @@
...
+def save_keyless_invitations(
+    slug: str,
+    pending: dict[str, dict[str, Any]],
+) -> None:
+    """Atomically persist the whole invitation document.
+
+    Writes a sibling ``*.tmp`` file then ``os.replace`` over the target so
+    a crash mid-write never leaves a truncated document; raises ``OSError``
+    on failure so the caller restores its in-memory mirror.
+    """
```
hellohanchen, line 198: The docstring says "a crash mid-write never leaves a truncated document". `os.replace` is atomic for the rename, but without an `fsync` of the temp file (and the directory) a power loss can still leave a short file after the rename is visible.

The failure mode is safe — `parse_keyless_invitation` rejects the row and the operator gets re-prompted — so I'd correct the claim rather than add the fsync, unless you want the stronger durability guarantee for decision state.

hellohanchen:
Addendum to my review above — I ran the repository's own structure gate against `e8a7e5a`:

```
$ python3 tools/check_python_structure.py
src/puffo_agent/agent/send_coordinator.py:1: file is 2010 lines (limit 2000)
src/puffo_agent/portal/cli.py:468: cmd_agent_create is 101 lines (limit 100)
Python structural check failed: 2 violation(s).
```

Both are introduced by this PR — `send_coordinator.py` crosses the 2,000-line limit and the workspace-link call added at `cli.py:539` pushes `cmd_agent_create` one line over. The Codex reviewer flagged the same two; this is just independent confirmation that the gate actually fails rather than a style opinion.

Worth noting the Validation section lists Ruff but not this check, so it looks like it wasn't run. `AGENTS.md` treats both limits as active refactoring work rather than advisory, so I'd expect pre-commit/CI to reject the commit as it stands.

None of this changes the substantive findings in my review — flagging it because it blocks the merge mechanically and is cheap to fix.

src/puffo_agent/agent/send_coordinator.py
```diff
@@ -38,6 +38,7 @@
     coordinator_config,
     http_error_detail,
     optional_response_int,
+    persist_baseline,
```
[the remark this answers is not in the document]
Glimmer0x: Fixed. The semantic send models were extracted before this follow-up, and e4791e6 keeps send_coordinator.py below the repository limit. python3 tools/check_python_structure.py and the pre-commit structural hook both pass.

src/puffo_agent/portal/cli.py
```diff
@@ -534,6 +536,7 @@ def cmd_agent_create(args: argparse.Namespace) -> int:
         created_at=int(time.time()),
     )
     cfg.save()
+    ensure_workspace_shared_link(cfg.resolve_workspace_dir(), shared_fs_dir())
```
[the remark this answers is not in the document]
Glimmer0x: Fixed. Shared-workspace CLI setup now lives in the focused _prepare_cli_shared_workspace helper, so cmd_agent_create remains below the method limit. The structural check and full pre-commit suite pass.

src/puffo_agent/agent/global_inbox_runtime.py
```diff
@@ -726,6 +751,98 @@ async def _admit(self, planned: PlannedTurn, event: ProviderAdmissionEvent) -> N
...
+        immutable snapshot the durable requeue/process just used.
+        """
+        if self.status_lifecycle is None or not terminal:
+            return
+        try:
+            await self.status_lifecycle.on_turn_terminal(
+                message_ids=tuple(self.active.message_ids),
+                succeeded=succeeded,
+                error_text=error_text,
+            )
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in e4791e6. Processing run IDs are reconstructed deterministically from the durable turn ID plus message ID, preserving the existing run_ + 32-hex wire shape. Recovery now settles status on successful resume, cancellation, activated unwind, stale/current-turn unwind, and orphan startup recovery. Server processing/start is idempotent for the same run_id and batch end is an upsert. The crash-recovery regression verifies the restarted process reuses and ends the original run.

src/puffo_agent/portal/worker_run.py
```diff
@@ -87,6 +94,69 @@ def stop(self):
...
+        self,
+        *,
+        message_ids: tuple[str, ...],
+        succeeded: bool,
+        error_text: str | None,
+    ) -> None:
+        runs = self._build_runs(message_ids, succeeded, error_text) if self._began else []
+        try:
+            if runs:
+                await self._reporter.end_turn_batch(runs)
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in e4791e6. The keyless end_turn_batch branch now selects the first failed entry and forwards its bounded error_text through the bridge instead of emitting a detail-free error state. The focused keyless status regression covers success and failure.

src/puffo_agent/portal/workspace_layout.py
Glimmer0x: Fixed in e4791e6. Callers now retain the workspace preparation status, conflict/unavailable are surfaced in logs or CLI output, and generated AGENTS.md/CLAUDE.md text is selected from the actual state. A conflict explicitly says the private directory is not connected; unavailable tells the Agent not to rely on shared/. Stale, dangling, empty-directory migration, conflict preservation, and prompt truthfulness are covered.

src/puffo_agent/agent/harness/docker_runtime.py
Glimmer0x: Fixed in e4791e6 using the defensive Docker shape suggested here. Docker preparation removes a prior local-runtime symlink and bind-mounts the host shared root directly at /workspace/shared; real existing content yields conflict and is never hidden by the mount. The container layout version is bumped so existing containers rebuild. Besides the argv regression, a real Docker Desktop local-to-Docker smoke read host shared data successfully through /workspace/shared.

src/puffo_agent/agent/send_response_validation.py
Glimmer0x: Confirmed deliberate against the deployed Server contract; no Agent change is needed. Merged puffo-server #281 keeps context_baseline_seq optional on Held and returns the request value unchanged, while only a successful Sent response establishes and returns an integer baseline. Strict held equality therefore detects contract drift instead of accepting an invented boundary. The native/keyless null-baseline lifecycle regression covers null held followed by established sent.

src/puffo_agent/agent/send_response_validation.py
Glimmer0x: Fixed in e4791e6. Baseline persistence now retries once, emits an error after exhaustion, returns failure to the coordinator, and appends a visible warning to the otherwise committed send result. The message remains correctly reported as sent while the caller is told that a later send may need context resynchronization. Both native and keyless paths are covered.

src/puffo_agent/agent/keyless_invitation_state.py
Glimmer0x: Fixed in e4791e6 by narrowing the documentation claim. It now promises atomic replacement for ordinary process failure and explicitly states that power-loss durability is not guaranteed without file and directory fsync.

Glimmer0x:
Review follow-up pushed in `e4791e6`.

Addressed all actionable comments:
- structural limits pass for `send_coordinator.py` and `cmd_agent_create`;
- crash recovery reconstructs and settles the original processing run;
- keyless batch failures retain diagnostics;
- shared-workspace status is propagated into operator output and generated prompts;
- stale/dangling/empty workspace migration is covered;
- Docker removes a prior local symlink, does not hide conflicting real content, and uses a direct `/workspace/shared` bind mount;
- null held-baseline validation is retained intentionally and matches merged Server #281;
- baseline persistence retries and reports visible degradation without misreporting an already committed send;
- invitation durability documentation no longer overclaims fsync guarantees.

Validation:
- `uv run pre-commit run --all-files` passed
- `uv run pytest -q` passed, with the existing Windows-only skip
- focused review regressions passed
- real Docker Desktop local-to-Docker shared-workspace smoke passed
- `git diff --check` passed

All review threads have a scoped reply and are resolved for re-review.

PR #235, opened 2026-08-14
Revert to 1.2.0 — move Agent Foundation 2.0 to the v2 branch

hellohanchen:
Restores `main`'s tree to the 1.2.0 release commit `431bb77`, reverting the 47 commits merged since.

**Nothing is lost.** The 2.0 work is preserved verbatim on [`v2`](https://github.com/puffo-ai/puffo-agent/tree/v2) at `bd8d6ea`, which is byte-identical to `main` as of this PR. This is a normal commit on top of `main`, not a force-push — the history stays intact and this is revertible in one click.

## Why

Production has been rejecting agent runtime events with **409 `invalid_turn_transition`** since **2026-08-11 19:00 UTC**:

| Day (UTC) | Rejections |
|---|---|
| 08-11 | 1,007 |
| 08-12 | 1,992 |
| 08-13 | 4,087 |
| 08-14 | 3,489 |

Zero before that — verified over a 20-day window, and specifically 0 across 08-07 → 08-11 12:00.

**It is not a server regression.** puffo-server ran the same binary (`3565e2df`) in production from 08-07 to 08-13, and both the `require_running` check and its logging have existed since 07-30.

The producer is what changed. The runtime-event outbox and its `turn_ref` / `session_ref` do not exist at 1.2.0 — they arrived on `main` with #225 *"Integrate Agent Foundation across native and keyless cloud runtimes"* at **08-11 00:37**, ahead of the `2.0.0a1` / `2.0.0a2` alphas. The first rejections follow ~18 hours later.

## What users see

`agent_runtime_events.rs` allows an agent exactly one active turn. Once the server's projection leaves `running`, that turn accepts nothing further, and the agent cannot open a replacement either — `turn.started` is refused with *"Agent already has an active turn"*.

So the agent stops responding and cannot recover on its own. **Restarting it is the only way out**, because a restart opens a fresh `session_ref`. That matches the reports: agents that go quiet, receive nothing, reply to nothing, and come back only after a restart.

The retry pattern in the logs is a stable stuck loop — three groups of five attempts ~1.07s apart, repeating every 5 minutes, indefinitely.

## Effect

1.2.0 emits no runtime events at all, so merging this stops the rejections outright rather than mitigating them.

## Not addressed here

This is a stop-the-bleeding revert, not a fix for the turn lifecycle. The underlying defect is still unidentified: the failing `POST /v2/agent-runtime/events:append` is logged with only `failure_code` and `status` — **no `agent_slug`, `turn_ref`, or `session_ref`** — so the logs cannot say which agents are affected, nor distinguish the two distinct causes the server collapses into one code:

- `"Agent already has an active turn"` (a `turn.started` against a live turn)
- `"turn.finished does not match the active turn"`

Both surface as `invalid_turn_transition`. Adding those fields server-side is cheap and would make the next investigation a log query instead of a code read.


Glimmer0x:
Closing as superseded by the repaired 2.0 path. #237 is now merged with daemon-owned Inbox turns and runtime observability kept off the execution path, which addresses the stuck-turn failure that motivated this emergency rollback. #238 restores the 1.2 flat-memory compatibility surface on top of that architecture. Reverting the repository to 1.2 now would discard the validated fixes and the later compatibility work.

PR #243, opened 2026-08-17
(Reopen) PUF-430: on archive, leave every space so the agent drops out of channel rosters

hellohanchen:
Reverts puffo-ai/puffo-agent#242


Glimmer0x (changes requested):
v2 blocker: the native signed archive path is covered and passes, but bridge/keyless agents have no local identity and no bridge/server leave+revoke lifecycle primitive. In the current implementation archive returns without convergence: pending_leave remains true, pending_revoke is not written, and retries cannot complete, so an archived keyless agent can remain in space/channel rosters. Please either add the keyless server/bridge lifecycle contract and route archive through it, or explicitly scope this PR to native transport with a warning and track bridge roster cleanup as a separate release blocker. CI is green; this is a product-contract blocker, not a test failure.

PR #255, opened 2026-08-18
fix(runtime): recover stuck turns after a runtime crash or an unresumable session

hellohanchen:
## Root cause: two distinct bugs in the same failure family

Fixes the 2.0.0a15–a17 regression that got fleet agents stuck permanently in a `RuntimeStateError` → `turn.failed` → requeue loop, unable to recover on their own.

### Bug 1: crashed-runtime abandons never retried

`RuntimeManager._consume_events()` already tags `TURN_ABANDONED` events caused by a crashed provider runtime (`RUNTIME_EXITED`, event-persistence failure, timeout cleanup) with `data["retryable"] = True`. But `RuntimeManagerAdapter._consume_turn_events()` discarded that flag and always raised a bare `RuntimeStateError` — not an `AgentAPIError`, so `global_inbox_runtime._invoke_turn_with_retries()` never entered its session-transfer retry path (`handle_global_inbox_retry` / `transfer_turn_session`, added in `b0c13b5`) and fell through to a dumb requeue that never repairs the persisted session pointer.

**Fix**: `_consume_turn_events()` now raises `AgentAPIError` when the terminal event says `retryable=True`. Genuine provider-reported failures (`turn.completed` with outcome `"failed"`/`"cancelled"` and no `retryable` flag) keep raising `RuntimeStateError` exactly as before.

### Bug 2: an invalid `--resume` target isn't always a synchronous open failure

claude-code CLI doesn't always fail `driver.open()` when `--resume` targets a session id that no longer exists. Sometimes the process spawns and initializes fine, and the invalid resume only surfaces later as an ordinary `turn.completed` with `outcome="failed"` (stderr `"No conversation found with session ID: ..."`); other times it crashes the whole process (`RUNTIME_EXITED`) instead. Neither carries a `retryable` flag on its own, so under Bug 1's fix both stayed non-retryable too — but nothing ever detected that the *persisted* `native_session_id` itself was permanently dead and replaced it, so the agent retried the identical broken id forever.

**Fix**: `RuntimeManager` now tracks `_resume_unconfirmed` (true from a resumed open until its first turn succeeds). A first-turn failure on an unconfirmed resume discards the session id and forces the terminal event retryable so it flows through the same `AgentAPIError` / session-transfer retry path as Bug 1, landing on a genuinely fresh session. Once a turn succeeds, the session is confirmed good and later failures go back to plain `RuntimeStateError`.

**Refined in #256** (merged into this branch as a stacked follow-up, see that PR for its own write-up): rather than gating the reset purely on "is this the first turn after resume", the driver now parses the claude-code `result` frame's structured `errors` field for the exact `"No conversation found with session ID:"` diagnostic and tags it `error_code="invalid_resume"`; `RuntimeManager` resets specifically on that precise signal (not on any first-turn failure), which avoids misfiring on an unrelated provider error that happens to land on turn one. The `RUNTIME_EXITED`-crash variant (no structured signal available) still uses the `_resume_unconfirmed` heuristic. #256 also fixes a related issue where context admission could degrade before the first turn because claude-code's native autocompact capability/threshold wasn't learned until `system/init`, and adds a short retry "kick" when a native session survives across a retry.

Both exception types carry an optional `error_code` (mirroring the terminal event's own `error_code`/`outcome`) through the existing allowlisted `error_code` observability field, so failure sites can be told apart in production logs without ever exposing raw exception text.

`AgentAPIError` lives in `agent/errors.py` with no adapter/harness imports, so this adds no circular dependency.

### DM messages never surfaced their thread_root_id

Separately reported live: an agent replying inside a DM thread could see the DM target (`target_type="dm" target_ref="dm:<peer>"`) but never the `thread_root_id`, even though the row always carries one and `send_message`'s `root_id` parameter already fully supports DM replies (`_validate_outgoing_root_scope` has a dedicated `dm_peer` branch). The read side just never told the agent what `root_id` to pass back.

**Fix**: `target_label()`'s DM branch now also resolves `_effective_thread_root()` and appends `thread_root_id` to the header when present, exactly like the channel branch already does. `target_type` stays `"dm"` and `target_ref` stays `dm:<peer>` — DMs aren't independently routable per thread the way channel threads are, so this is presentation-only, with zero change to routing, coalescing, or `canonical_target_parts()`.

## Tests

- Updated the existing `test_runtime_manager_failures.py` assertions that exercised retryable abandons to expect `AgentAPIError`, with `is_auth`/`error_code` coverage, plus new tests for both Bug 2 failure shapes and the confirmed-session/non-retryable paths that must stay unaffected.
- `test_claude_runtime_recovery.py` (added in #256) covers the `invalid_resume` precision detection and autocompact capability learning.
- `test_dm_projection_exposes_thread_root_id_but_keeps_dm_routing` covers the DM projection fix.

Full suite runs clean (`pytest`, excluding two GUI test files that can't collect without the optional `PySide6` extra — pre-existing, unrelated to this change).

## Live verification against the fleet

Installed this branch into the production pipx daemon (`pipx install git+...`) and restarted, iterating as each newly-discovered failure shape came up. All four previously-stuck agents recovered and completed real turns: `equation-7256-87f7`, `old-coder-7968-8e7b0303`, `solution-5805-d1be`, and `calculatio-4282-36fa`. No regressions observed across the 24-agent fleet. Production is currently running this branch's code.

### DIFF
--- a/docs/claude-runtime-recovery-investigation-2026-08-18.md
+++ b/docs/claude-runtime-recovery-investigation-2026-08-18.md
@@ -1,0 +1,172 @@
+# Claude Runtime Recovery Investigation
+
+Date: 2026-08-18
+
+## Question
+
+Does `puffo-agent` recover reliably when a Claude Code Agent cannot continue
+its persisted provider session, and do PRs #253 and #255 cover every observed
+way that the Agent can become permanently stuck?
+
+## Observed Failure Families
+
+### 1. Context admission blocks before a provider turn starts
+
+One production Claude Agent has 79 durable pending messages. Its last
+successful turn ended on 2026-08-17. Since then the daemon has repeatedly
+planned work but has never started another provider turn.
+
+Representative admission evidence:
+
+```text
+used_tokens=495678
+projected_tokens=503475
+context_window=1000000
+outcome=degraded
+```
+
+The context controller's soft target is 50% of the provider window. Once the
+projected turn crosses that target, admission requires native compaction or a
+rollover. Claude reports neither capability at process open when
+`auto_compact_threshold_pct` is unset. The Claude stream-json process only
+advertises `/compact` in `system/init`, and current Claude Code does not emit
+that frame until it receives its first user frame. Admission therefore blocks
+the frame that would disclose the capability needed to unblock admission.
+
+PR #253 makes a configured `auto_compact_threshold_tokens` available at open
+time and injects the matching `--autocompact` argument. It does not help
+existing/default Agent configurations where both compact threshold fields are
+unset. The affected production Agent is still in that default state after the
+2.0.0a17 restart.
+
+Two other Claude Agents remained in the same degraded state for roughly 20
+hours before a process restart happened to reopen them with lower reported
+context use. No native compaction event was recorded, so that recovery is not
+a deterministic fix.
+
+### 2. A started provider turn fails on a crashed or invalid resumed session
+
+PR #255 addresses a later failure boundary. A provider turn has already been
+started, but the runtime either exits or reports a failed first turn because
+the persisted `--resume` target is no longer usable.
+
+Two defects are plausible and independently useful to fix:
+
+1. `TURN_ABANDONED(retryable=True)` was converted into a non-retryable
+   `RuntimeStateError`, bypassing Global Inbox's session-transfer retry path.
+2. Claude can accept process startup for an invalid resume ID and report the
+   failure only after the first user frame, so synchronous open fallback is
+   insufficient.
+
+PR #255 propagates retryable terminal events as `AgentAPIError` and tracks an
+unconfirmed resumed runtime until one provider turn succeeds. It clears the
+native session ID and retries from a fresh session when that first resumed
+turn fails or the runtime exits.
+
+## Review Concerns In PR #255
+
+The retryable terminal propagation follows the existing Global Inbox recovery
+contract and appears directionally correct.
+
+The unconfirmed-resume classification is currently broader than its evidence:
+every non-successful first turn on a resumed session is treated as proof that
+the resume ID is invalid. A valid resumed session can have a first turn fail
+for unrelated reasons such as rate limiting, authentication, policy rejection,
+provider execution failure, cancellation, or malformed input. Resetting that
+session loses provider-native history and replays only Puffo's reconstructed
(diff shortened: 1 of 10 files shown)


Glimmer0x:
I reproduced and reviewed the remaining stuck-Agent paths against the current Claude CLI and ran an independent Claude Fable 5 source review. PR #256 targets this branch and completes the recovery contract.

The important deltas are: pre-turn `get_context_usage` now discovers native compact capability without a guessed default; only the explicit missing-conversation diagnostic becomes `invalid_resume`; invalid session state is cleared before terminal persistence; confirmed sessions survive resource reloads; and retry payload selection now distinguishes a preserved session from a fresh replacement.

Full pytest, pre-commit, structural limits, Ruff, and diff checks pass.

Glimmer0x:
Updated this branch onto current `origin/main` (`dd8fa3e`, 2.0.0a19) and resolved the RuntimeManager conflict with #260's native-session continuity work.

Conflict resolution preserves both contracts:

- the failed driver is closed once by the current-main open-failure path;
- `_clear_native_session()` clears the invalid native ID and its confirmed-session marker together;
- resource/account reloads still preserve a valid confirmed native session;
- pre-turn Claude context usage can publish newly learned `/compact` capability through `RuntimeManager.current_capabilities()`;
- only the explicit `invalid_resume` provider result retires an otherwise valid resumed session.

Fresh verification after the merge:

- targeted context/recovery/session-continuity suite: 171 tests passed;
- full `uv run --extra dev pytest -q`: passed, one existing Windows-only skip;
- `uv run pre-commit run --all-files`: passed, including Ruff, structural limits, and conflict detection.

Glimmer0x:
Final bounded-context fallback is now included in `fca6298`.

After native compaction is unavailable or exhausts its two bounded attempts, and even the minimum Inbox batch cannot fit below the context soft target, the RuntimeManager now opens a fresh Claude/Codex native session while preserving the Puffo logical `session_ref`. Normal turns, successful compaction, batch shrinking, profile/resource reloads, and confirmed native-session resume remain unchanged.

Fresh verification on the merged current-main branch:
- targeted runtime/context/recovery suite: 172 passed;
- full `uv run --extra dev pytest -q`: passed with one existing Windows-only skip;
- `uv run pre-commit run --all-files`: passed.

PR #268, opened 2026-08-19
fix(runtime): resolve Claude's autocompact threshold from the static table

hellohanchen:
## Follow-up to #265

#265 fixed the compounding-threshold crash loop (`equation-7256-87f7`, `old-coder-7968-8e7b0303` stuck permanently on `--autocompact` argument `'90000' is invalid`) by pinning the threshold to whatever was already on `spec`/`status` and never recomputing it. That's safe, but doesn't match how this problem was actually solved before, in **#219**.

## What #219 already established

When Claude Code stopped honoring the legacy `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` env var and we switched to computing `--autocompact <tokens>` ourselves, #219 explicitly called out that Claude echoes the *configured* ceiling back as its own `rawMaxTokens` once `--autocompact` is active — not the model's real capacity. Verified again empirically here: launching with `--autocompact 300000` makes the CLI report `rawMaxTokens: 300000` right back (opus-4-7's real window is 1,000,000). `build_context_runtime()` (still used by `worker.py`/`agent_detail.py` for status/UI) handles this by always resolving from the static per-model table whenever a pct override is configured — never trusting the live-reported window.

`runtime_manager.py`'s `_context_threshold()` (added in #265) doesn't do that yet — it just freezes on `configured` for Claude, meaning the threshold is only ever as correct as whatever happened to be set at launch, and any later pct config change has no path to apply.

## This PR

- `_context_threshold()`: for `driver_name == "claude-code"` with a pct configured, resolve via `claude_autocompact_tokens(model, pct)` against the static table — the identical call `local_runtime.py` makes at launch. Deterministic for a fixed model+pct, so it can never disagree with what's already running from a normal launch.
- `get_context_snapshot()`: the reload-on-mismatch branch was gated `driver_name != "claude-code"` (codex only). Extended to cover claude-code too, now that the resolution is stable — it only fires once, to apply a pct that changed after launch, not on every poll.
- Restored the "corrects a stale threshold exactly once" regression test that didn't survive #265's squash-merge.

## Verification
- Full test suite + ruff: clean
- `test_context_snapshot_stays_pinned_once_it_matches_launch` (already existed): unaffected — resolving from the static table lands on the same 300_000 the launch spec already has, so no reload
- `test_context_snapshot_corrects_a_stale_claude_threshold_exactly_once` (new): spec starts with pct configured but no token value yet; first poll resolves 300_000 and reloads once; second poll doesn't reload again

Not deployed yet — production is currently running #265's version (pinned, no static-table resolution or drift-correction), which is safe and already fixes the crash loop. This is a refinement, not a hotfix.

### DIFF
--- a/src/puffo_agent/agent/harness/runtime_manager.py
+++ b/src/puffo_agent/agent/harness/runtime_manager.py
@@ -49,10 +49,7 @@ COMPACTION_WAIT_SECONDS = 120.0
 
 
 def _resolve_claude_autocompact_tokens(*, model: str, pct: float) -> int | None:
-    """Same call local_runtime.py makes at launch -- static per-model window,
-    never the driver's live-reported one (see PR #219: once --autocompact is
-    active, Claude echoes that configured ceiling back as its own window).
-    """
+    """Same call local_runtime.py makes at launch -- static per-model window."""
     from ...portal.control.context_telemetry import claude_autocompact_tokens
 
     return claude_autocompact_tokens(model=model, pct=pct)
@@ -894,11 +891,8 @@ class RuntimeManagerAdapter(Adapter):
         if pct is None:
             return configured
         if self.manager.driver_name == "claude-code":
-            # Claude echoes the configured --autocompact ceiling as its live
-            # context window, so window * pct compounds smaller on every
-            # observation (PR #219). Resolve from the static per-model
-            # table instead -- deterministic for a fixed model+pct, same
-            # call local_runtime.py already made at launch.
+            # Claude echoes our configured ceiling as its live window, so
+            # window * pct would compound smaller each time (PR #219).
             return (
                 _resolve_claude_autocompact_tokens(
                     model=self.manager.spec.model, pct=pct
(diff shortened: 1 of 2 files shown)


Glimmer0x (changes requested):
Blocking runtime-contract mismatch: the new recomputation does not always use the same resolution inputs as launch.

`DockerRuntimePreparer._prepare_claude_spec()` calls `claude_autocompact_tokens(..., env=self.agent_cfg.env_overrides)`, intentionally isolating Docker from the daemon process environment. `_resolve_claude_autocompact_tokens()` omits `env`, so it reads `os.environ`. With `CLAUDE_CODE_AUTO_COMPACT_WINDOW=500000` on the daemon, model `claude-opus-4-7`, and pct `50`, Docker launches at `500000` from the static 1M model window, but the first context snapshot recomputes `250000` from the host override and reloads the provider with the wrong value. The existing Docker test at `tests/test_context_telemetry_threshold.py:539` already establishes that the host override is ignored for Docker launch.

Please make the resolver use the launch spec environment (or retain a non-null launch-owned token and resolve only when both spec/status are missing), and add one cross-boundary regression case covering Docker launch -> RuntimeManager snapshot. Also note that normal `env_overrides` edits already restart the worker and rebuild the full spec via `_worker_needs_restart`, so telemetry should not overwrite a valid launch-owned token merely to apply a later pct edit.

Glimmer0x:
Superseded by #269, which preserves this PR's missing-threshold repair while keeping the launch-resolved threshold authoritative and using the RuntimeSpec launch environment for fallback resolution. #269 passed the full test suite, CI/CodeQL, and an independent Fable 5 review, and is now merged.

PR #276, opened 2026-08-22
feat(inbox): message covers — explicit disposition, finalize reconciliation, one-shot redelivery

Glimmer0x:
## What

The safety net for the audited "read but never answered" message loss (45 confirmed net-loss cases across prod agents): the model now **declares** which inbound messages each outbound action disposes of, and the daemon reconciles at turn end.

### Covers
- `send_message` / `create_reminder` accept `covers=[message_id...]` — the reply/deferral's disposition claim (message-level, cross-target allowed). No covers = covers nothing; there is deliberately no target-level inference (per design discussion with Glimmer).
- New `mark_covered` tool settles a message that needs no reply, or backfills a forgotten declaration.
- Covers land in a `message_covers` table (doubles as reply-traceability data).

### Finalize reconciliation + one-shot redelivery
- At turn end: human messages read − covers declared = uncovered set. A `turn.uncovered_messages` observation event always fires; behind the `covers_renotice` flag (daemon.yml; `PUFFO_COVERS_RENOTICE` env as test override) uncovered rows are redelivered **once**, marked `uncovered_redelivery=true`, with a `renotified` bit guaranteeing no loops.
- The notice announces the backlog as `[uncovered message_count=N]` metadata.
- Naming is deliberate: *uncovered* (no declaration recorded), not *unaddressed* — the ledger can't know whether a message was handled, only whether a claim was made.
- Prompt layer: dedicated **Covers** section in the platform primer; turn cue requires per-message disposition before ending a turn.

### Hardening from an 8-angle adversarial review (~31 findings, all triaged)
- One completion transaction serves `mark_processed` and the renotice path (restores the exact-membership pre-check the copy had dropped).
- Renoticed rows release notice contributions **by row** — redelivery no longer depends on incidental recovery paths.
- `model_visible_at` survives renotice (no freshness-watermark regression → no spurious held sends).
- A cover recorded during the redelivery window settles the row.
- Sender classification is by exclusion (agent/system exempt): non-operator human DMs with no roster facts are inside the net, not outside it.
- Crash recovery completes through the same reconciliation as normal finalize.
- `uncovered_redelivery` renders only while the redelivery is live; settled rows stop nagging history readers forever.
- Held sends report `covers_dropped`; one `_parse_covers` wire validator for all three RPC routes (falsy wrong types are 400s, not silent drops); `mark_covered` rejects blank-only lists.
- Rolling-upgrade shims: old-daemon `create_reminder` retries without covers and reports them dropped; `mark_covered` explains unavailability.
- Reviewed-and-rejected: the claimed "back-page permanent silence" P1 and a finalize/admission race were both refuted by direct testing (regression tests pin the correct behavior: `tests/test_inbox_backpage_renotice.py`, `test_renotice_survives_contribution_from_earlier_generation`).

## Testing

Full suite green (3240+ tests, 0 failures), including 20+ new tests covering the covers lifecycle, redelivery edge cases, wire validation, and the refuted-audit regressions.

## Rollout

Observation events are unconditional; redelivery ships **off** by default. Suggest enabling `covers_renotice: true` in the test environment first, watching `turn.uncovered_messages` rates, then prod.


Glimmer0x:
Blocking failure-mode issue: finalize catches any cover-reconciliation error and then falls through to mark the full turn processed. I reproduced this by making the covered-id lookup raise with covers renotice enabled; the human message ended processed and was not renotified. That makes a failure in the safety net silently acknowledge exactly the work it is supposed to protect. Please fail closed for this path by requeueing or leaving the turn recoverable, while keeping duplication bounded. One focused regression is enough. Non-blocking boundary note: the cover store currently treats any locally known row as coverable, including outbound or otherwise non-inbound rows; narrowing this to eligible inbound messages would better match the declared contract and keep traceability honest.

Glimmer0x:
Re-review complete: the original blocker is resolved. A reconciliation read failure now emits an ERROR event and uses an atomic store-side partition to requeue every still-unrenotified turn member once, while already-renotified rows settle. This preserves the bounded duplicate-over-loss behavior for the reproduced failure. Focused tests passed (17 tests). Non-blocking residual: if both reconciliation and the fallback write fail, the code deliberately completes plainly; that is an explicit availability-over-loss escape hatch rather than fully fail-closed behavior and should remain documented/observable.

PR #277, opened 2026-08-22
fix(data-client): read failures raise DataUnavailable instead of rendering empty

Glimmer0x:
## What

Audited P1: `DataClient` swallowed data-service 5xx responses and transport errors into `return []` / `return None` on five read paths (channel history, DM history, channel roots, thread messages, single message). `read_history` then rendered the failure as a definite "no messages" window — the model draws a confidently wrong conclusion (e.g. "this thread has no replies") from a transient outage, with no error trace.

## Fix

- New `DataUnavailable(RuntimeError)` (in `message_store_models`, re-exported beside `DataNotFound`); the five read paths raise it on ≥400 (except 404) and on transport errors.
- The MCP tool layer surfaces it as an explicit tool error the model can retry. Consumers that already degrade gracefully keep doing so — `resolve_outgoing_root` catches broadly and sends top-level with an honest note.
- 404 absence semantics unchanged (`[]` / `None` / `DataNotFound` as before).
- `lookup_channel_space` deliberately stays lenient (auxiliary lookup; `None` is treated as unknown, not as fact) — noted for a follow-up if it ever matters.

## Testing

New `tests/test_data_client_honest_errors.py` (real aiohttp server: 500s, connection-refused transport errors, and 404-preservation). Full suite green, 0 failures.

### DIFF
--- a/src/puffo_agent/agent/message_store.py
+++ b/src/puffo_agent/agent/message_store.py
@@ -20,6 +20,7 @@ from .reminder_store import ReminderStoreMixin
 PRIOR_CONTEXT_MAX_ITEMS = _models.PRIOR_CONTEXT_MAX_ITEMS
 PRIOR_CONTEXT_MAX_BYTES = _models.PRIOR_CONTEXT_MAX_BYTES
 DataNotFound = _models.DataNotFound
+DataUnavailable = _models.DataUnavailable
 ReceiptDisposition = _models.ReceiptDisposition
 ProcessingState = _models.ProcessingState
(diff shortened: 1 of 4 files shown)


Glimmer0x:
Mechanism review complete. The five read paths preserve the important semantic distinction: 404 remains genuine absence, while 5xx and transport failures surface as retryable uncertainty instead of an empty fact. The exception reaches the MCP boundary without being converted back to an empty window. Targeted tests passed locally. No blocking findings.

PR #278, opened 2026-08-22
fix(inbox): preserve unverifiable thread roots instead of erasing them

Glimmer0x:
## What

Audited P2 from the message-reliability investigation: when a reply referenced a thread root that was not in the local store (root predates the agent joining the channel, aged out, or failed to decrypt), the inbound path **wiped** `thread_root_id` and permanently demoted the reply to a channel-level message. Cross-store evidence showed real thread replies rendered as top-level posts, and the idempotent redelivery path never backfilled the lost reference.

## Fix

- `resolve_incoming_thread_root` now distinguishes an **affirmative mismatch** (wrong channel/space/DM, cycle, over-deep chain — still wiped: those prove the claim wrong) from a **locally unverifiable** root, which keeps the claimed id with a new `thread_root_unverified` flag instead of erasing it. Grouping and thread replies keep working off the claimed id.
- New `PARENT_MISSING` sentinel in `validated_parent` carries the distinction; `validate_incoming_parent_id` behavior is unchanged.
- **Deferred verification on root arrival**: when the claimed message later arrives, `_settle_unverified_thread_claims` applies the same ownership rules the live path uses — matching channel/space (or DM conversation participants) settles the flag; a claimed "root" that is itself a reply re-roots claimants to its own root; anything else wipes the claim.
- The flag is persisted (schema migration with default 0), round-trips through `StoredMessage`, and renders as `thread_root_unverified=true` in message projections so the model knows the thread context may be incomplete.

## Testing

New `tests/test_thread_root_preservation.py` (9 cases): missing-root preservation, mismatch still wipes, persistence + projection round-trip, settle-on-arrival (channel and same-conversation DM), wipe-on-arrival for foreign channel and foreign DM conversation, re-rooting through a late reply, and verified rows untouched. Full suite green.

### DIFF
--- a/src/puffo_agent/agent/bridge_transport.py
+++ b/src/puffo_agent/agent/bridge_transport.py
@@ -658,6 +658,15 @@ async def _bridge_storage_row(client, payload: MessagePayload) -> dict[str, Any]
         if payload.sender_slug == client.slug
         else payload.sender_slug
     ) if payload.envelope_kind == "dm" else ""
+    thread_root_id, thread_root_unverified = (
+        await client._resolve_incoming_thread_root(
+            payload.thread_root_id,
+            payload.channel_id,
+            payload.space_id,
+            expected_envelope_kind=payload.envelope_kind,
+            expected_dm_peer=dm_peer,
+        )
+    )
     return {
         "envelope_id": payload.envelope_id,
         "envelope_kind": payload.envelope_kind,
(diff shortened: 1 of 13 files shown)


Glimmer0x:
Blocking user-flow gap: the inbound half now preserves a missing root as thread_root_unverified, but the outbound half still calls the existing root resolver, which returns no root whenever that same id is absent locally. Reproduction on this PR head: store an inbound reply with missing-root marked unverified, then resolve an outbound send to missing-root; it is still sent top-level. So grouping is preserved, but replying in the thread is not, contrary to the PR's stated flow. Please close the outbound path with a scoped trust rule without weakening cross-channel or DM ownership checks, and add one focused end-to-end regression from inbound preserved reply to outbound threaded send. No need for broader test volume.

Glimmer0x:
Re-review complete: the original blocker is resolved. The preserved unknown inbound root is now replyable outbound through a scoped local claimant, while wrong-channel and wrong-DM claimants are still rejected by the existing scope validator. The thread read path is also closed for rootless local threads. Focused tests passed (86 tests across thread-root preservation and route resolution).

PR #280, opened 2026-08-22
fix(agent): replay processed receipts durably

Glimmer0x:
## Problem

A transient failure while posting `/messages/processing/end:batch` could permanently lose per-message processed status. The local message was already complete, but the server never received the green-dot acknowledgement.

## Architecture

```mermaid
flowchart LR
  A[Turn completion or stale catch-up] --> B[Per-agent messages.db outbox]
  B --> C{Serialized dispatcher}
  C -->|success| D[Server processing runs]
  D --> E[Delete acknowledged rows]
  C -->|transient failure| F[1m / 5m / 10m / hourly]
  F --> C
  C -->|permanent batch 4xx| G[Isolate to singleton]
  G --> C
  H[Restart or reconnect] --> I[0-30s jitter]
  I --> C
```

## What changed

- Adds one durable processing-report outbox to each existing `messages.db`.
- Attempts terminal reports immediately, then retries at 1m, 5m, 10m, and hourly.
- Force-drains all pending rows after restart or transport reconnect.
- Sends at most 50 oldest-first records per request, with one in-flight request and one request per second per Agent.
- Uses deterministic run IDs and allows a recovered terminal result to replace an older pending result for the same run.
- Isolates permanent batch failures so one poison row cannot block valid rows.
- Replaces the stale catch-up in-memory buffer with the same durable path.
- Re-asserts current status after background or mixed replay batches so old reports cannot leave the Agent status stale.
- Makes reporter/dispatcher run latches reusable across `ws-local` attach/detach cycles.
- Leaves keyless/cloud bridge behavior unchanged.

## Verification

- Full `uv run pytest -q`: passed; one existing Windows-only skip.
- Focused processing, staleness, status, lifecycle, and ws-local suites: passed.
- `ruff check`, Python structure check, and `git diff --check`: passed.
- sdist and wheel build: passed.
- Claude Fable 5 High final review: no P0/P1 findings; its only P2 cancellation rollback finding was fixed before this PR.

## Server contract

This relies on the already-merged server PR #304 behavior: duplicate heartbeats are rate-limited, while real status or message-anchor transitions are accepted immediately.


Glimmer0x:
Review blockers are addressed in b8f1316. Processing-report acknowledgements now conditionally mutate the exact persisted payload/version, so a replacement queued during an in-flight upload survives. Stale catch-up now commits the local receipt first and enqueues server completion only after an ACK. Focused tests, pre-commit, CI, and the combined five-PR full suite pass.

PR #287, opened 2026-08-24
feat(send): channel is_encrypted drives the outbound channel format

hellohanchen:
## What

Ports PR #212's format rules onto the 2.0.0 architecture, against puffo-server's coordinated plaintext support (server PR #321, merged):

1. **Channel sends follow `channels.is_encrypted`** — encrypted → sealed envelope (unchanged); plaintext → signed plaintext envelope through the **same** `/v2/agent-runtime/messages:send` freshness pipeline (require_current / held / send_anyway all apply to plaintext channels). Plaintext sends skip device-key resolution entirely.
2. **DMs stay E2EE** — the DM path is untouched; the agent cannot choose a format.
3. **`CHANNEL_FORMAT_MISMATCH` recovery** — on the server's mismatch error the coordinator re-reads the channel policy from the server, updates the cache, and resends **once** in the other format; an unchanged policy surfaces the failure (`error_kind: channel_format_mismatch`).

**Policy cache** (`PuffoCoreMessageClient`): in-memory `_channel_encrypted` populated by space warms (the channel list now carries `is_encrypted`), persisted in `disk_cache` (policy preserved across policy-less rewrites), updated live by the new `channel_update` WS frame, evicted alongside the other channel caches. Unknown/missing/legacy always fails safe to **encrypted**. The coordinator takes the client as an optional `channel_policy_source`; without one (keyless MCP fallback) it stays always-encrypted — keyless channel sends already route through the bridge where the **server** picks the format.

Not needed anymore from the original plan: the old `send_mode` rules (turn-bundle / thread-root) were already deleted upstream in the 2.0.0 release, and synthetic system-message rows keep the store's fail-safe `is_encrypted=True` default.

## Tests

13 new (`test_channel_format_policy.py`): mismatch detector contract, disk-cache policy round-trip + preservation, sealed/plaintext format selection off the policy, mismatch → refresh → cross-format resend, mismatch with unchanged policy fails once, missing-policy-source default, warm/WS-update/ensure/refresh/eviction cache behavior, WS `channel_update` dispatch. Existing manual-construction fixtures updated for the new cache attr. Full suite: **0 failures** (2 pre-existing environment skips).

Depends on a puffo-server deployment carrying #321 (already on server main).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- **Channel sends follow the channel's format policy.** The channel's
+  `is_encrypted` field alone decides sealed vs signed-plaintext for channel
+  sends; DMs stay encrypted and the agent cannot choose the format. The
+  policy is cached (disk + memory), kept live by `channel_update` frames,
+  and a `CHANNEL_FORMAT_MISMATCH` rejection refreshes the policy once and
+  resends in the channel's current format. (#212)
+
 ## [2.0.0] - 2026-08-23
 
 > Stable Agent Foundation 2.0 release, promoted from `2.0.0a25` without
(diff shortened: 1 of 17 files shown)


hellohanchen:
Superseded: the 2.0.0 port now lives on PR #212 (branch puf-411-agent-send-decision updated to this implementation).

PR #289, opened 2026-08-24
perf(startup): expose daemon readiness before worker prep

Glimmer0x:
## Summary

- report daemon readiness once local WS/RPC/data/control services are initialized
- run the expensive MCP fingerprint scan off the event-loop thread, before the first worker reconcile
- lazily import the daemon in the CLI so detached startup does not import the daemon in both parent and child
- add stage timing logs and one lifecycle-order regression test

## Why this is stacked

This PR is based on #288, which fixes the correctness bug where a live daemon was killed after a fixed 10-second startup observation window. This follow-up reduces the actual cold-start wait without changing worker startup ordering.

## Ordering contract

`local control plane ready -> MCP/session worker preparation -> first worker reconcile`

The ready marker no longer waits for the MCP scan, but no worker can start before that scan completes. Worker preparation remains best-effort and logs failures instead of terminating an already-ready daemon.

## Evidence

- isolated detached cold start on macOS: 3.25s baseline -> 0.25s
- control-plane ready log: 6ms
- worker preparation log: 121ms
- focused startup/background/MCP tests: 33 passed
- full test suite: passed (1 existing Windows-only skip)
- `pre-commit run --all-files`: passed
- `python tools/check_python_structure.py`: passed

## Deferred

Windows tray startup still imports Qt before launching the daemon thread. That is a separate process/lifecycle change and is intentionally not included here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Background startup now distinguishes slow, stalled, stopped, and exited
+  daemons without terminating a live process after a fixed observation window.
+- Detached startup reports success when a concurrently launched child exits
+  because another daemon won startup, and avoids redundant daemon imports in
+  the parent CLI process.
+
 ## [2.0.0] - 2026-08-23
 
 > Stable Agent Foundation 2.0 release, promoted from `2.0.0a25` without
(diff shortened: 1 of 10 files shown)


Glimmer0x:
Addressed the Opus 5 fresh-context review and follow-up review in commits `7b3ff11` and `42015af`:

- accept a redundant detached child only when it exits 0 and one validated winner PID is alive
- distinguish `starting`, `stalled during startup`, `stopping`, and `stop stalled` without terminating a live process
- publish the stop sentinel for SIGINT/SIGTERM shutdowns too
- check an old stalled daemon before waiting through another observation window
- replace the lazy wrapper with an inline lazy import
- add only the lifecycle tests covering these branches and document the behavior

Verification after the final follow-up:
- focused lifecycle tests: passed
- full pytest suite: passed (1 existing Windows-only skip)
- pre-commit: passed
- Python structure check: passed
- isolated detached startup: 0.26s to CLI success; control plane ready in 6ms; worker preparation completed in 127ms

PR #290, opened 2026-08-24
Fix codex resume recovery: invalid_resume classification + session preservation

hellohanchen:
## Context

大范围 codex agent 掉线(8/19 ~11:30)的根因修复。两个 bug,一个死亡链条:

1. daemon 重启/瞬时失败 → turn start 失败路径把 runtime retire 时 `preserve_session=False`,持久化的 native session id 被清空
2. 下次启动 `thread/start` 新建 thread;若该 thread 在首个 turn 落盘前随 app-server 一起死掉,codex 不会写 rollout jsonl
3. 再次 resume 该 thread → JSON-RPC -32600 "no rollout found" → `_classify_jsonrpc_error` 把 detail 洗成通用 "The provider could not complete the turn."
4. `_native_resume_is_unavailable` 的文本 marker 匹配不上 → fresh-session fallback 失效 → 盲目重试 3 次 → worker 死亡(claude driver 因走 `error_code="invalid_resume"` 通道而幸免)

## Fix

**Bug #1 — codex_driver**: "no rollout found" / "rollout not found" / "thread not found" → `AgentAPIError(detail, error_code="invalid_resume")`,保留 native detail;`_native_resume_is_unavailable` 优先检查 `error_code == "invalid_resume"`(双通道覆盖)。

**Bug #2 — runtime_manager**: turn start 失败的 retire 路径改为 `preserve_session=True`。driver close 已防止 old turn overlap;session 真死时由 fix #1 的 fallback 兜底转 fresh。steer/cancel/timeout 等 driver 明确宣告 session 不可复用的路径保持原语义不变。

## Tests

- `test_jsonrpc_missing_rollout_is_classified_as_invalid_resume`(3 param):分类 + detail 保留
- `test_native_resume_falls_back_on_invalid_resume_error_code`:文本被洗时 error_code 通道仍触发 fresh fallback
- `test_failed_turn_start_preserves_the_native_session`:失败 start 后 session 保留且下次 open resume 同一 thread
- 两个既有测试的断言更新为新语义(失败 start 保留 session)

三个 fix 点均已 mutation-verified(逐一 revert → 对应测试变红 → 恢复)。全量:3189 passed, 5 skipped(PySide6/Windows-only,与本 PR 无关)。

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **Codex resume recovery.** A resume against a thread with no rollout on disk
+  ("no rollout found") is classified as `invalid_resume` and falls back to a
+  fresh session instead of dying in a blind retry loop, and failed turn starts
+  keep the persisted native session id so restarts resume the same thread
+  instead of creating a new one. (#290)
+
 ## [2.0.0] - 2026-08-23
 
 > Stable Agent Foundation 2.0 release, promoted from `2.0.0a25` without
--- a/src/puffo_agent/agent/harness/codex_driver.py
+++ b/src/puffo_agent/agent/harness/codex_driver.py
@@ -168,8 +168,7 @@ def _classify_jsonrpc_error(error: Any) -> Exception:
     detail = f"Codex JSON-RPC error code={code}: {message}"
     context = _jsonrpc_error_context(error)
     if _is_invalid_resume_error(message):
-        # Missing rollout: keep native detail so the fresh-session
-        # fallback (_native_resume_is_unavailable) can match it.
+        # Missing rollout: keep detail for the fresh-session fallback.
         logger.warning(
             "codex provider request failed (invalid_resume): %s", detail
         )
--- a/src/puffo_agent/agent/harness/codex_driver.py
+++ b/src/puffo_agent/agent/harness/codex_driver.py
@@ -148,6 +148,18 @@ def _jsonrpc_error_status(error: Any) -> int | None:
     return None
 
 
+_INVALID_RESUME_MARKERS = (
+    "no rollout found",
+    "rollout not found",
+    "thread not found",
+)
+
+
+def _is_invalid_resume_error(message: str) -> bool:
+    lowered = message.lower()
+    return any(marker in lowered for marker in _INVALID_RESUME_MARKERS)
+
+
 def _classify_jsonrpc_error(error: Any) -> Exception:
     """Translate a Codex JSON-RPC error into Global Inbox recovery semantics."""
     error_obj = error if isinstance(error, dict) else {}
@@ -155,6 +167,13 @@ def _classify_jsonrpc_error(error: Any) -> Exception:
     message = _safe_jsonrpc_error_message(error_obj.get("message", error))
     detail = f"Codex JSON-RPC error code={code}: {message}"
     context = _jsonrpc_error_context(error)
+    if _is_invalid_resume_error(message):
+        # Missing rollout: keep native detail so the fresh-session
+        # fallback (_native_resume_is_unavailable) can match it.
+        logger.warning(
+            "codex provider request failed (invalid_resume): %s", detail
+        )
+        return AgentAPIError(detail, error_code="invalid_resume")
     error_code = classify_provider_failure(
         status=_jsonrpc_error_status(error),
         diagnostic=context,
--- a/src/puffo_agent/agent/harness/runtime_manager.py
+++ b/src/puffo_agent/agent/harness/runtime_manager.py
@@ -93,12 +93,18 @@ _TRANSIENT_RESUME_ERROR_CODES = frozenset({
     "quota_exhausted",
 })
 
-# Unclassified resume failures tolerated before assuming the session is gone.
+# Consecutive unclassified resume failures tolerated before assuming the
+# session is gone.
 _RESUME_FAILURE_FALLBACK_AFTER = 3
 
 
 def _resume_error_is_transient(exc: Exception) -> bool:
     if getattr(exc, "is_auth", False):
+        return True
+    # AgentAPIError is recoverable-with-backoff by contract (invalid_resume
+    # is intercepted earlier); only categorized non-retryable failures
+    # (ProviderFailureError) and raw exceptions count toward the streak.
+    if isinstance(exc, AgentAPIError):
         return True
     return getattr(exc, "error_code", None) in _TRANSIENT_RESUME_ERROR_CODES
 
@@ -249,6 +255,8 @@ class RuntimeManager:
                 raise
             if not _native_resume_is_unavailable(exc):
                 if _resume_error_is_transient(exc):
+                    # Consecutive semantics: transients break the streak.
+                    self._resume_failure_streak = 0
                     raise
                 self._resume_failure_streak += 1
                 if self._resume_failure_streak < _RESUME_FAILURE_FALLBACK_AFTER:
--- a/src/puffo_agent/agent/harness/runtime_manager.py
+++ b/src/puffo_agent/agent/harness/runtime_manager.py
@@ -86,6 +86,8 @@ class RuntimeStateError(RuntimeError):
 
 def _native_resume_is_unavailable(exc: Exception) -> bool:
     """Whether the provider explicitly rejected the saved native session."""
+    if getattr(exc, "error_code", None) == "invalid_resume":
+        return True
     text = str(exc).lower()
     identifies_session = any(
         marker in text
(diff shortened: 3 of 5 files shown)


Glimmer0x (changes requested):
Blocking: preserving `native_session_id` for every failed `turn/start` breaks the retry payload contract.

`RuntimeManagerAdapter.run_retry_turn()` (lines 1265-1271) uses the mere presence of `native_session_id` as proof that the original input is already in the provider transcript, so it sends only the cheap kick. After this PR, `_discard_failed_start_locked()` retains that id even when `turn/start` failed before the provider admitted the input (for example, an immediate provider rejection or a write failure). The retry then sends `continue processing` to a thread that never received the task, silently losing the durable input.

There is a second ordering variant: if the retained native session is actually dead, `run_retry_turn()` chooses the kick *before* `run_turn()` calls `open()`; `open()` may then fall back to a fresh session on `invalid_resume`, but the already-selected content is still the kick, not `fallback_user_message`.

I reproduced the first case with a focused regression: let `_FailingStartDriver.start_turn()` raise `RuntimeError("provider refused the turn")`, then call `run_retry_turn("continue processing", "full durable input", ...)`. Expected input for the retry is the durable payload; actual input is `continue processing`. The new test fails exactly as expected, while the two adjacent PR tests pass.

Please decouple “keep the saved thread id” from “the original input was admitted”. A failed start needs explicit admission state (and the retry must use the durable payload when admission is false/unknown), or a stable correlation/idempotency scheme that makes replay safe. `native_session_id` alone cannot carry both meanings. Add a regression covering the end-to-end failed-start → retry payload path, including the invalid-resume → fresh-session ordering case.

Glimmer0x:
Additional blocking recovery case (surfaced by Linus, reproduced locally on head 731d5fd): a dead native session can loop forever when the provider changes its human-readable error wording. For example, a resume error `Codex rollout missing` does not match `_INVALID_RESUME_MARKERS`; three consecutive `RuntimeManager.open()` attempts all retry `SessionRef("native-old")`, raise, and never attempt `open(..., None)`.

This is distinct from the original-input-loss path in my review, though both come from treating retained `native_session_id` as stronger evidence than it is. Please add a bounded/structured fallback so an unclassified resume failure cannot permanently wedge the agent, while still avoiding premature session loss on transient auth/network/rate-limit failures. A regression should demonstrate eventual fresh-session recovery for an unknown missing-session wording; it should not add another phrase-specific marker.

Glimmer0x:
Rechecked current head 731d5fd against the existing blocking review.

The focused Codex/runtime-manager tests pass, but the failed-start to retry contract is still untested and still broken in the implementation: _discard_failed_start_locked preserves native_session_id for an exception or negative receipt, while run_retry_turn chooses the cheap kick solely from native_session_id. That ID does not prove the original durable input was admitted, and the invalid-resume fallback happens only after the kick has already been selected.

The unknown missing-session wording case also remains phrase-marker based, so an unclassified provider wording can still retry the same dead session indefinitely.

There are no commits after the cited review/comments, so both findings remain blocking on this head.

hellohanchen:
Both blockers addressed in 9e89404:

**Retry payload contract.** The manager now carries explicit input-admission state instead of overloading `native_session_id`:
- `_input_admitted` is set **only** by an accepted start receipt, and cleared by every failed start (`_discard_failed_start_locked` — exception, negative receipt, unsupported) and by `_clear_native_session`. Unknown admission reads as not-admitted: duplication is recoverable, silent input loss is not.
- `run_retry_turn` now calls `manager.open()` **before** choosing its payload, so the invalid-resume fresh-session fallback (which clears admission) happens ahead of the choice — fixing the ordering variant.
- The kick is sent only when `native_session_id and input_admitted`; every other state replays the exact durable payload.

Regressions: `test_failed_start_retry_replays_the_durable_payload` (end-to-end failed start → retry carries the durable payload into the driver), `test_retry_kick_is_reconsidered_after_a_fresh_session_fallback` (dead session → fallback → durable payload, asserting the resume/None open sequence), and the rewritten `test_retry_payload_depends_on_input_admission` unit contract.

**Unclassified resume failures.** No new phrase markers. `_open_locked` keeps three tiers: classified-transient (auth / rate_limit / provider_unavailable / quota_exhausted, via `error_code`/`is_auth`) retries the same session indefinitely; explicit invalid-resume falls back immediately; anything unclassified falls back after a bounded streak of 3 consecutive failures (`_RESUME_FAILURE_FALLBACK_AFTER`, reset on successful open and session replacement). Regressions: `test_unclassified_resume_failures_fall_back_after_a_bounded_streak` (uses the exact "Codex rollout missing" wording — two raises, then fresh) and `test_classified_transient_resume_failures_never_lose_the_session` (5 rate-limit failures, session intact).

All four protections mutation-verified (admission gate, open-before-choice ordering, bounded streak, transient guard — each reverted individually, its regression goes red). Full suite: 3193 passed, 5 skipped.

Glimmer0x:
Re-reviewed current head `9e894040eaace4fca7f699542d5e29e6752df7cc`. The original two blockers are substantially fixed: explicit input-admission state now prevents cheap-kick input loss, payload selection happens after `open()`, and the unknown missing-session wording gets a bounded fallback. The new end-to-end regressions cover those paths.

One blocking recovery boundary remains in `_resume_error_is_transient`: it only protects `is_auth` or four explicit error codes. An untagged `AgentAPIError` is therefore counted as an unclassified resume failure and loses the saved native session on the third attempt. That contradicts the `AgentAPIError` contract (recoverable provider failure with backoff), the existing `test_native_resume_keeps_session_on_transient_error` fixture (`AgentAPIError("provider temporarily unavailable", is_auth=False)`), and the requirement that transient network/provider failures not cause premature session loss.

I reproduced this with the existing `_ResumeBoundaryDriver` and `AgentAPIError("temporary transport failure")`: attempts 1 and 2 raise while retaining `native-old`; attempt 3 calls `open(..., None)` and replaces it with `native-session-1`. A real fresh open that also fails would already have cleared the persisted old id. Please keep recoverable `AgentAPIError` (and any raw transport exceptions the drivers intentionally expose) outside the destructive fallback counter, and add a repeated-failure regression—not just the current one-attempt transient test.

Secondary consistency point: `_resume_failure_streak` is not reset when a classified transient occurs, so it counts cumulative unclassified failures rather than the “3 consecutive failures” described in the PR comment. Please either make the state transition match that contract and test an interleaved transient, or document/test the cumulative policy explicitly.

Local verification on the PR head: focused Codex/runtime-manager suite `60 passed`; `pre-commit run --all-files` passed. GitHub currently reports the PR as conflicting with `main` (the direct merge conflict is in `CHANGELOG.md`), and this head currently has CodeQL results but no CI pytest/pre-commit checks.

hellohanchen:
Addressed in 0843973 (+ merge commit 4e80d6d resolving the CHANGELOG conflict):

**Recoverable failures exempted from the destructive counter.** `_resume_error_is_transient` now treats every `AgentAPIError` as transient — it is recoverable-with-backoff by contract, and the `invalid_resume` case is intercepted earlier by `_native_resume_is_unavailable`. The streak therefore only counts categorized non-retryable failures (`ProviderFailureError`, e.g. the codex driver's laundered `provider_error`) and raw exceptions. Repeated-failure regression: `test_untagged_agent_api_errors_never_lose_the_session` (5 consecutive untagged `AgentAPIError` opens — session intact, every resume targets `native-old`). The wedge fix keeps its coverage from both sides: `test_unclassified_resume_failures_fall_back_after_a_bounded_streak` (raw `RuntimeError`) and new `test_categorized_provider_errors_hit_the_bounded_fallback` (`ProviderFailureError(error_code="provider_error")` — the exact class the codex driver emits for unknown wordings).

**Consecutive semantics.** A classified transient now resets `_resume_failure_streak`, matching the "3 consecutive unclassified failures" contract. Regression: `test_transient_resume_failures_reset_the_unclassified_streak` interleaves 2 unclassified + 1 rate-limit + 3 unclassified and asserts fallback on the 6th attempt, not the 4th.

Both new protections mutation-verified (AgentAPIError exemption flipped → repeated-failure test red; streak reset removed → interleaved test red).

**Housekeeping.** Merged `origin/main` and resolved the `CHANGELOG.md` conflict (kept #212's Changed entry, folded this PR's expanded Fixed entry alongside main's startup fixes) — GitHub now reports MERGEABLE. CI is running on this head. Full suite post-merge: 3226 passed, 5 skipped; `pre-commit` structural check passes.

Glimmer0x (approved):
Approved on current head `0843973edca432e29cfc493ea0b0d861f05bdf71`. The remaining recovery blocker is resolved: every recoverable `AgentAPIError` now stays outside the destructive resume-fallback counter, and classified transients reset the counter so it represents consecutive unclassified failures. The new regressions cover repeated untagged API errors, an interleaved transient reset, and bounded fallback for categorized non-retryable provider errors.

The earlier input-admission and open-before-payload fixes remain intact after merging current `main`. Local focused verification: `63 passed` (`tests/test_codex_driver_jsonrpc_errors.py` + `tests/test_runtime_manager_failures.py`); `pre-commit run --all-files` passed. GitHub CI and CodeQL are green, and the PR is mergeable with no remaining conflict. No further blocking findings from me.

Glimmer0x:
Post-merge follow-up, not a claim that the merged fix is wrong:

I independently reproduced a conditional gap on current origin/main (1dd4871, including squash merge ad1203c) by composing the real Codex JSON-RPC classifier with RuntimeManager.

Four session-loss phrasings recover to a fresh session within the bounded streak: sess-9 is stale/start a new one, session expired, rollout missing, and thread unavailable. Two timeout-worded phrasings do not recover after 30 attempts: conversation timed out; and session timeout, please open a new conversation.

Mechanism: classify_provider_failure sees timeout first and returns retryable provider_unavailable; _classify_jsonrpc_error then exposes only the generic public message through AgentAPIError; RuntimeManager treats every AgentAPIError as transient, resets the resume-failure streak, and never reaches fresh-session fallback.

Evidence boundary: this behavior is reproduced in real components, but I have no evidence that Codex currently emits either wording. The wording is also ambiguous: counting all timeout failures toward destructive fallback could discard a valid session during a real provider outage. I therefore do not recommend reopening #290 or adding timeout tokens to the invalid-resume allowlist. This should remain a follow-up design question: preserve enough structured/original diagnostic context to distinguish provider outage from native-session rejection, or choose and document a bounded policy for ambiguous failures.

PR #295, opened 2026-08-26
feat(membership): announce invite-link channel joins to the agent

hellohanchen:
## Summary

puffo-server #325 emits one synthetic `add_to_channel` event per channel joined through an invite-link redemption, delivered only to that channel's active members. puffo-agent previously dropped these events silently — channel members never learned who joined via a link (especially in private channels).

- mirror `add_to_channel` in the agent's `EventKind` enum
- route it through the existing channel membership announcement path
- actor = `added_slug`, action = `joined`, inviter label = `signer_slug` (the link's creator, set intentionally by the server)
- existing guards apply unchanged: channels the agent isn't in are skipped (matching the server's members-only delivery), self-joins are skipped, per-event-id dedup absorbs WS redelivery, and the deterministic envelope id makes cross-restart replay idempotent

## Validation

- 7 new WS-routing unit tests (happy, unknown channel, self, missing `added_slug`, redelivery dedup, multi-channel grant, dispatch fall-through) and 7 cross-layer E2E tests driving raw WS frames through `PuffoCoreWsClient._handle_frame` → `PuffoCoreMessageClient` → real sqlite `MessageStore` → projection/routing (private channel, concurrency, store failure)
- new + adjacent tests green 3 consecutive runs; full suite 3289 passed, 1 skipped (Windows-only)
- 100% line and branch coverage of the new executable code (coverage --branch, per-line verified)
- ruff + structural checks pass
- live smoke on a 25-agent daemon against a server with #325 deployed: multi-channel link redemption announced per channel by two member agents; private-channel grant invisible to non-member agents; exactly-once across a WS reconnect and a full daemon restart

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- **Invite-link channel joins are announced to the agent.** When someone
+  joins channels by redeeming an invite link, the server's synthetic
+  `add_to_channel` events now produce a per-channel system message
+  ("X joined channel #name (invited by Y)", with Y the link's creator)
+  in each affected channel the agent is a member of, alongside the
+  existing space-level join announcement.
+
 ## [2.0.2] - 2026-08-25
 
 ### Fixed
--- a/src/puffo_agent/agent/event_kinds.py
+++ b/src/puffo_agent/agent/event_kinds.py
@@ -24,4 +24,5 @@ class EventKind(StrEnum):
     REMOVE_FROM_SPACE = "remove_from_space"
     REMOVE_FROM_CHANNEL = "remove_from_channel"
     CREATE_CHANNEL = "create_channel"
+    ADD_TO_CHANNEL = "add_to_channel"
     REDEEM_INVITE_CAPABILITY = "redeem_invite_capability"
--- a/src/puffo_agent/agent/membership_actions.py
+++ b/src/puffo_agent/agent/membership_actions.py
@@ -245,11 +245,22 @@ async def maybe_announce_membership_change(
     inviter_by_event_id: Mapping[str, str],
     processed_event_ids: set[str],
     enqueue_message: AsyncCall,
+    rewarm_channels: AsyncCall | None = None,
     log: Log,
 ) -> None:
     channel_id = payload.get("channel_id") or ""
-    if not channel_id or channel_id not in channel_spaces:
+    if not channel_id:
         return
+    if channel_id not in channel_spaces:
+        if kind != EventKind.ADD_TO_CHANNEL or rewarm_channels is None:
+            return
+        # The server delivers add_to_channel only to that channel's
+        # members, so a cache miss usually means the connect-time warm
+        # task hasn't landed yet — rewarm once before deciding the
+        # channel isn't ours.
+        await rewarm_channels()
+        if channel_id not in channel_spaces:
+            return
     actor = _channel_membership_actor(
         kind=kind,
         event=event,
--- a/src/puffo_agent/agent/membership_actions.py
+++ b/src/puffo_agent/agent/membership_actions.py
@@ -283,6 +283,12 @@ def _channel_membership_actor(
     if kind == EventKind.ACCEPT_CHANNEL_INVITE:
         inviter = _inviter_slug(payload, inviter_by_event_id)
         return event.get("signer_slug") or "", "joined", "", inviter
+    if kind == EventKind.ADD_TO_CHANNEL:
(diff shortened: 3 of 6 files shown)


Glimmer0x (changes requested):
One blocking reliability issue:

- `src/puffo_agent/agent/puffo_core_client.py:510-512`: `add_to_channel` immediately enters `maybe_announce_membership_change`, whose first guard drops the event when `channel_id` is absent from `_channel_space`. On first connect/restart that map starts empty, while `_on_ws_connect` only schedules `_warm_member_caches()` in a fire-and-forget task. A join event arriving before that task finishes is therefore silently lost even though the server delivered it precisely because this agent is an active channel member; WS membership events have no durable client replay.

Please rewarm/verify the channel mapping on an `add_to_channel` cache miss before deciding it is not our channel (for example, await the serialized `rewarm_channel_caches()` path, then re-check), and add a regression where `_channel_space` begins empty, the rewarm populates the delivered channel, and the join is announced. Keep a second case where rewarm still does not reveal the channel and the event remains a no-op.

The wire schema matches puffo-server #325, focused membership/invite tests and Ruff pass, and hosted CI is green.

hellohanchen:
@Glimmer0x Fixed in bb4be28 — confirmed real: `_on_ws_connect` only fire-and-forgets `_warm_member_caches()`, so a join delivered right after (re)connect could hit an empty `_channel_space` and be dropped with no client-side replay.

The miss-handling now lives in `maybe_announce_membership_change`: on an `add_to_channel` cache miss it awaits the injected `rewarm_channels` callback (the client passes the existing serialized + 5s-debounced `rewarm_channel_caches()`), then re-checks before deciding the channel isn't ours. Other kinds keep the plain drop-on-miss behavior.

Regressions added at both levels:
- routing: empty cache + rewarm reveals the channel → announced (and rewarm not called on a warm-cache hit); rewarm still doesn't reveal → no-op, no announcement; `leave_channel` miss → no rewarm round-trip; no `rewarm_channels` callback → plain drop.
- E2E: raw WS frame racing the connect warm (empty `_channel_space`, stubbed warm reveals the channel, real `rewarm_channel_caches` lock/debounce path) → message persisted exactly once; unknown-after-rewarm → nothing durable.

New lines remain at 100% line+branch coverage; focused tests green 3 consecutive runs; full suite 3294 passed, 1 skipped.

Glimmer0x (changes requested):
复核后仍有一个阻塞竞态：`maybe_announce_membership_change()` 在 cache miss 时调用的是现有 `rewarm_channel_caches()`，但该方法带 5 秒 debounce。若任何一次 rewarm 刚在事件到达前完成，而当时新频道尚未出现在列表中，`_last_rewarm` 仍在窗口内；这次 `add_to_channel` 会跳过实际 warm，重查仍 miss，然后永久丢掉一次性 WS 事件。实测把 `_last_rewarm = time.monotonic()`、缓存置空、让下一次真实 warm 可发现频道后，处理事件得到 `warm_calls=0, announcements=0`。新增两条测试覆盖了 rewarm 执行后的 found/not-found，但没有覆盖 debounce-suppressed 路径。建议让这个事件驱动的 authoritative recheck 能绕过 debounce（仍共用串行锁），或等待/复用一个能保证发生在该事件之后的 warm，并加“recent prior rewarm + add_to_channel”回归测试。其余新增/相邻测试均通过（47 passed）。

hellohanchen:
@Glimmer0x Confirmed — the debounce window was a real second hole: a warm finishing just before the grant landed left `_last_rewarm` inside the 5s window, so the event-driven rewarm no-op'd, the recheck still missed, and the one-shot event was gone.

Fixed in d645c6d: `rewarm_channel_caches()` gains `force: bool = False` which skips only the debounce while still serializing on `_rewarm_lock`; the `add_to_channel` recheck passes `force=True`, and the other on-miss callers (data service, ws-local) keep the debounced behavior. Stampede exposure stays bounded: the forced path only runs on a cache miss, and the server delivers `add_to_channel` solely to members, so misses are the connect/just-granted races, not a steady state.

New regressions: E2E "recent prior rewarm + add_to_channel" (empty cache, `_last_rewarm = now`, warm reveals the channel → exactly one persisted message, `warm_calls == 1`) and a debounce-intact check (two unforced calls → one warm; forced → runs). Changed lines at 100% line+branch coverage; focused tests green 3 consecutive runs; full suite 3410 passed, 1 skipped.

Glimmer0x (approved):
Re-review complete. The event-driven cache miss now bypasses only the 5-second debounce while retaining the shared rewarm lock, and the new end-to-end regression proves a recent prior rewarm no longer drops the one-shot add_to_channel event. Ordinary callers remain debounced. Full local pytest, pre-commit, diff check, and all hosted CI/CodeQL checks pass.

PR #298, opened 2026-08-27
Surface swallowed worker-task exceptions via universal spawn helper

hellohanchen:
## Summary

Every `asyncio.ensure_future` / `asyncio.create_task` in the daemon now
routes through a single helper (`puffo_agent.tasks.spawn`) that attaches a
universal done-callback. Unhandled exceptions in worker tasks are logged
with the full traceback (`worker task died: <name>`) instead of vanishing
into the event loop. Shutdown cancellation stays silent; awaited-companion
paths do not double-surface.

Motivating incident: the 2026-08-26 Prof-Puffo restart hang. A
`ValueError: message backup key has invalid length` fired at
`keystore.py:149` on every startup and was swallowed by an unmonitored
`reminder_task`, producing zero log output. Diagnosis took a full day and
six misdirected stages. With this helper in place, that failure surfaces
on restart #1.

## Change shape

- **New helper** — `src/puffo_agent/tasks.py` (37 lines):
  - `spawn(coro, *, name=None)` — `get_running_loop().create_task(coro, name=name)`
    for coroutines (preserves the `RuntimeError("no running event loop")`
    contract callers already rely on); `asyncio.ensure_future(awaitable)` +
    conditional `set_name` for non-coroutine awaitables.
  - `_report_task_death(task)` — `Task.done()` callback. Skips
    `task.cancelled()` and `task.exception() is None`. Logs
    `ERROR "worker task died: %s"` with `exc_info=<exception>`.
- **Sweep** — 26 modules migrated from bare `asyncio.ensure_future` /
  `asyncio.create_task` to `tasks.spawn`. `bridge_transport.py` layers
  `_report_task_death` alongside its existing `client._ack_tasks.discard`
  housekeeping callback (housekeeping preserved).
- **Fleet-lint** — two AST tests
  (`test_no_bare_task_spawn_remains_in_tree`,
  `test_helper_is_the_only_ensure_future_owner`) walk
  `src/puffo_agent/**/*.py` (excluding `tasks.py`) and fail on any
  `asyncio.ensure_future` / `asyncio.create_task` / `loop.create_task` /
  `loop.ensure_future` call outside the helper. Guards future regressions.

## Test coverage

- **New unit tests** — `tests/test_task_spawn.py`, 13 tests:
  - Happy path spawn + await.
  - Unhandled exception surfaces via logger at ERROR with full traceback
    and task name.
  - `CancelledError` from cooperative shutdown stays silent.
  - `exception() is None` (normal completion) stays silent.
  - `name=` propagates to callback log.
  - Pre-built Task / plain Future path (non-coroutine `ensure_future`).
  - No-running-loop → `RuntimeError` (contract preserved).
  - Awaited-companion double-surface guard (caller `await` still raises
    once; callback logs once; no duplicate ERROR).
  - Two AST fleet-lint tests as described above.
- **Updated tests** —
  `tests/test_auth_failed_dm_notification.py`,
  `tests/test_drained_state.py`: adjusted where the previous
  `asyncio.create_task` call-site was patched or asserted.
- **Full suite** — 3341 passed, 0 failed, 0 error, 1 skipped
  (pre-existing Windows-only skip), exit 0.

## Race / concurrency coverage

Explicitly tested:
- Callback FIFO ordering with pre-existing housekeeping callback
  (bridge_transport) — housekeeping and death-log both fire; order does
  not swap.
- Awaited path — a caller that awaits the returned task still receives
  the exception normally; the callback still logs; no double-report of
  the same exception on the same task.
- Shutdown-cancel path — `Task.cancel()` mid-run does not trigger the
  ERROR log.
- Non-coroutine awaitable path — plain `asyncio.Future` handled without
  attempting `set_name` on objects that lack it.

## Non-changes

- No wire protocol change; no schema change; no persisted-state change.
- No dependency change.
- No behavioral change on any happy path — successful task completion is
  untouched. Only silently-swallowed exceptions are now surfaced as
  `ERROR` log lines.

## Fleet numbering

Fleet PUF-454 · Linear PUF-353.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,15 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   them too. Quota-exhausted refresh probes no longer count toward the
   `refresh_broken` streak. (#228)
 
+### Fixed
+
+- **Worker tasks no longer die unclaimed.** Every background task the daemon
+  spawns now carries a done-callback that logs `worker task died: <name>` with
+  the full traceback when the task ends on an unhandled exception, instead of
+  the exception vanishing into the event loop. Shutdown cancellation stays
+  silent. A start-up failure such as an unreadable message-backup key now
+  surfaces on the first restart rather than presenting as a silent hang.
+
 ## [2.0.2] - 2026-08-25
 
 ### Fixed
--- a/src/puffo_agent/agent/adapters/cli_session.py
+++ b/src/puffo_agent/agent/adapters/cli_session.py
@@ -48,6 +48,7 @@ from ..context_controller import (
     ToolResultAdmission,
     normalize_context_snapshot,
 )
+from ...tasks import spawn
 
 logger = logging.getLogger(__name__)
 
@@ -899,13 +900,14 @@ class ClaudeSession:
                 self.agent_id,
                 INIT_TIMEOUT_SECONDS,
             )
-            self._stderr_drain_task = asyncio.ensure_future(
-                self._drain_stderr(self._proc)
+            self._stderr_drain_task = spawn(
+                self._drain_stderr(self._proc),
+                name="drain_stderr",
             )
             return
         if sid and sid != self._session_id:
             self._save_session_id(sid)
-        self._stderr_drain_task = asyncio.ensure_future(self._drain_stderr(self._proc))
+        self._stderr_drain_task = spawn(self._drain_stderr(self._proc), name="drain_stderr")
 
     async def _read_init(self, proc: asyncio.subprocess.Process) -> str:
         while True:
@@ -1347,8 +1349,9 @@ class ClaudeSession:
             text = block.get("text", "") or ""
             state.reply_parts.append(text)
             if text and not is_silent(text):
-                asyncio.ensure_future(
-                    reporter.emit(self.agent_id, "assistant_text", {"text": text})
+                spawn(
+                    reporter.emit(self.agent_id, "assistant_text", {"text": text}),
+                    name="reporter.emit:assistant_text",
                 )
             if self.audit is not None and text:
                 self.audit.write("assistant.text", text=text)
@@ -1368,7 +1371,10 @@ class ClaudeSession:
         tool_event = {"tool": name}
         if "send_message" in name and isinstance(tool_input.get("text"), str):
             tool_event["content"] = tool_input["text"][:STATUS_PREVIEW_CHARS]
-        asyncio.ensure_future(reporter.emit(self.agent_id, "tool_use", tool_event))
+        spawn(
+            reporter.emit(self.agent_id, "tool_use", tool_event),
+            name="reporter.emit:tool_use",
+        )
         if name == "mcp__puffo__send_message":
             state.send_message_targets.append(
                 {
--- a/src/puffo_agent/agent/bridge_client.py
+++ b/src/puffo_agent/agent/bridge_client.py
@@ -26,6 +26,7 @@ from typing import Any, AsyncIterator, Awaitable, Callable, Optional
 import aiohttp
 
 from ..limits import MAX_INBOUND_ATTACHMENT_BYTES
+from ..tasks import spawn
 
 logger = logging.getLogger(__name__)
 
@@ -152,7 +153,7 @@ class CloudBridgeClient:
         logger.info("cloud bridge: WS connected (slug=%s)", self._slug)
         # Start the heartbeat only after a clean handshake so no failure
         # path leaves a live heartbeat task behind.
-        self._heartbeat_task = asyncio.create_task(self._heartbeat_loop())
+        self._heartbeat_task = spawn(self._heartbeat_loop(), name="heartbeat_loop")
         for callback in tuple(self._connected_callbacks):
             try:
                 await callback()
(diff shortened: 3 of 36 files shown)


Glimmer0x:
Peter review follow-up on head 46f0f66ad26e7fecc610f9db848026539f10f7e1:

- Rebased cleanly onto main after #296; the original four commits were range-diff identical.
- Fixed duplicate failure reporting by making supervised/awaited paths opt out of the detached-worker callback via report_failure=False. The owning path remains the single authoritative reporter.
- Made reporter registration idempotent for a pre-built Task/Future with a weak registration set.
- Strengthened the AST fleet guard to catch chained loop calls, arbitrary attribute owners, direct imports, and aliases, with negative-control tests.

Verification: full local pytest passes with the agent-injected CODEX_HOME removed from the test environment; one expected Windows-only skip. Relevant targeted suites and all pre-commit hooks pass. Fresh GitHub CI is green on this head. Independent re-review is still pending; this comment does not claim the second gate is closed.

Glimmer0x:
Follow-up review fixes are now on `ae81526`:\n\n- Cache warm tasks keep central reporters and are gathered together, so an early member failure cannot orphan/silence the channel warmer; a regression test proves two failed siblings produce two authoritative traceback records.\n- ws-local attached tasks keep the central reporter because cleanup is DEBUG-only; a regression test proves one consumer failure produces exactly one authoritative record.\n- The fleet lint is import-aware for direct and aliased asyncio imports, while explicit tests exempt `asyncio.TaskGroup` owners and unrelated `domain.create_task(...)` methods.\n- `report_failure=False` documentation now states the strict guarantee required: the owner must emit exactly one alertable traceback-bearing record.\n\nValidation: affected suite passed, pre-commit passed, and the full pytest suite passed with `CODEX_HOME` removed (one expected Windows-only skip). Fresh CI is running on this exact head.

Glimmer0x:
Follow-up fix pushed at exact head `eb92f1b` for the two remaining `report_failure=False` violations found in full-callsite review.

- `InboxCoalescer._sleep_or_pull`: both wait tasks now retain central failure reporting; cleanup cancels unfinished siblings and gathers both. Regression test injects a failing sleep and asserts exactly one traceback-bearing `puffo_agent.tasks` ERROR.
- `CredentialRefresher._sleep_until_next_tick`: both wait tasks now retain central failure reporting; cleanup cancels unfinished siblings and gathers both. Regression test injects two failed waits and asserts exactly one traceback-bearing central ERROR for each.
- Full source scan now leaves only deliberately owner-awaited `report_failure=False` callsites plus the helper contract test.

Validation on this head: targeted scheduler/refresher tests pass; `pre-commit run --all-files` passes; full `pytest -q` passes (one expected Windows-only skip); `git diff --check` passes. #298 remains HOLD pending independent exact-head review and fresh exact-head CI.

Glimmer0x:
Addressed the simultaneous-completion HOLD at exact head `1ddd4f3ffa395cac6e8c4a5fae456ec234e609d5`.

Changes:
- restored central failure reporting for both `reminder_scheduler.run` and `coalescer.wait_for_burst`;
- always settles the burst sibling with `gather(..., return_exceptions=True)` before propagating an owned reminder failure, including when both tasks complete in the same loop turn;
- added a double-failure regression asserting exactly two traceback-bearing `puffo_agent.tasks` ERROR records and no loop-level unhandled task exception.

Validation:
- focused runtime/task/worker/ws-local/docker set: 146 passed;
- full pytest: passed (one expected Windows-only skip);
- pre-commit: all passed;
- `git diff --check`: passed; worktree clean.

Remaining `report_failure=False` source sites were re-evaluated by completion ordering (child first, owner/listener first, simultaneous): listener/runtime pairing explicitly inspects simultaneous completion; docker communicate is directly awaited or kill+reaped; worker and ws-local runtime tasks flow through that listener/runtime settlement helper and cleanup awaits them. Requesting a fresh independent exact-head review.

Glimmer0x:
Structural stop-rule fix pushed at exact head `c8498af39ac9fe9771bcc8a76ae2c9b61791f734`.

Changes:
- removed `report_failure` from `spawn()` entirely; every spawned failure now gets the central traceback reporter and callers cannot opt out
- removed all four production opt-outs
- added an API-contract test proving reporting cannot be disabled
- strengthened listener/runtime simultaneous-failure control to require exactly two authoritative traceback-bearing task records and zero loop-unhandled contexts

Validation:
- 155 affected-path tests passed
- full pytest passed with one expected Windows-only skip when the live agent `CODEX_HOME` was removed from the test environment (running with the live override predictably contaminates seven unrelated temporary-home Codex config tests)
- pre-commit passed
- `git diff --check` passed
- worktree clean

HOLD pending fresh independent review of this exact head and exact-head CI.

Glimmer0x:
Follow-up exact head: `ff8aec650a85be975a5162e4e5fbf3f55ddcb935`

Documented the accepted supervised-reporting tradeoff in `spawn` and pinned it in the API contract test: a single underlying failure may yield additional authoritative supervisor records, so incident counts must deduplicate by exception rather than count `puffo_agent.tasks` ERROR lines.

Local verification:
- `env -u CODEX_HOME uv run pytest -q tests/test_task_spawn.py` — pass
- `env -u CODEX_HOME uv run pytest -q` — pass (one expected Windows-only skip)
- `env -u CODEX_HOME uv run pre-commit run --all-files` — pass
- `git diff --check` — pass

PR remains HOLD pending fresh independent review and exact-head CI.

PR #304, opened 2026-08-28
feat(harness): add constrained LingTai driver authority

Glimmer0x:
> [!NOTE]
> **评审已转移**:本 PR 已并入统一交付 [puffo-agent #328](https://github.com/puffo-ai/puffo-agent/pull/328)(agent 仓汇总,head `fcc1ddc7`;统一评审入口及 11→3 映射见其描述)。原评审正文与 head 保留如下,未改动。

## Summary

- add the Puffo-owned Driver authority server for constrained LingTai ACP launches
- pass exactly one issued POSIX authority endpoint to the puffo-v0 profile and strip caller-supplied authority state
- cover endpoint binding, single claim, derived-launch depth, provider adjudication, malformed frames, and spawn wiring
- ship the fixed-SHA D6 acceptance oracle

## Lineage and evidence boundary

The previously frozen implementation was Puffo e62c10c179245f81b152c9ac1edcb6527b7d00c9 against LingTai d69dacc400cf277fc380af10193c6f9af6e44779. That Puffo branch was based on an integration lineage later merged to main as rewritten/split commits, so opening it directly would have produced a 70-file, roughly 9,600-line PR.

This PR is a clean port onto current main. Its head is 695972bbd8d5aeea6657e942d8a3864198465c0c. The old PASS does not automatically transfer: the validation below is a fresh delivery-author run on this new SHA. Boris and Linus have separately claimed full independent reruns; their results are not yet recorded here.

## Validation by delivery author

- ruff check on all four changed files: pass
- pytest tests/test_driver_authority_server.py tests/test_acp_driver.py: 27 passed
- clean-clone D6 against LingTai d69dacc400cf277fc380af10193c6f9af6e44779: baseline=0, replay=1, fanout=1
- full pytest: seven failures, all reproduced identically on untouched origin/main; they are existing Codex host-MCP/config tests and do not touch this four-file diff

## Review and merge boundary

Acceptance, merge authorization, and deployment authorization are separate. Whole-tree equality against e62c10c1 is not a valid carry-forward check because main has advanced. This port must be judged on its own SHA and behavior. Any later conflict resolution or content change requires validation of the changed result; do not infer acceptance from a branch name or from the old frozen SHA.

---

## PR 文档（what / how）— 2026-09-09 交叉审 · Linus
**比较基线 base `069f7d75d27a`(main) → head `0e8a0d83`** · rev2（按 Peter 第一轮核对修订：§5 共触事实更正、§4 判断移入建议）· 事实与建议分列。

### 1. 问题
LingTai ACP 子进程需要向 Driver 请求受控能力（派生端点等），但子进程状态不可信：若权限面接受子进程自报的角色/lineage/能力，受限子进程即可越权。

### 2. 做了什么
新增 Driver-owned POSIX 权限服务器：继承的 fd 只是定位器（一条已连接 AF_UNIX socket），全部权威（role/lineage/depth/capability/provider）保存在 Driver 侧 endpoint record，不从子进程状态接受。

### 3. 怎么做的（机制）
`harness/driver_authority_server.py`：PROTOCOL_VERSION=1；帧上限 64KB；派生端点上限 16；lease 三态 issued/claimed/closed、claim 超时 60s；审计环 4096 条；被拒请求的随附 fd 全部收割关闭。`drivers/acp.py` 经 `LINGTAI_DRIVER_AUTHORITY_FD` 注入；`scripts/verify_lingtai_driver_authority.py` 独立核验。

### 4. 建模关系（事实）
在 Driver 与 ACP 子进程之间新增一个此前不存在的权限面组件；复用既有 Driver/ACP 进程模型与 fd 传递机制。

### 5. 重叠面（事实，rev2 更正）
**与 #319 共触 `drivers/acp.py` 与 `tests/test_driver_authority_server.py`**（rev1 写「零共触」是错的）。共同文件≠重复模型：#304 管调用权限/端点 lifecycle，#319 管工具投影与结果接纳——该分工判断待两份文档对照后在跨 PR 轮确认。另与 #311（echo driver authority call IDs）语义相邻（call_id 出现在审计记录），接口兼容性在 #311 审时对照。

### 6. 验证
`tests/test_driver_authority_server.py`。**历史结果（非本轮重跑）**：Peter 修两 P2 后我 delta 复核 PASS（两个逆变异亲跑）；CI 7/7 @0e8a0d83。状态：等 Old Coder 授权。

### 建议（非事实，rev2 自 §4 移入）
「新模型且必要／无既有结构可扩展」是待讨论判断，交叉轮以 #319 文档对照裁定。

固定版本入口：[driver_authority_server.py](https://github.com/puffo-ai/puffo-agent/blob/0e8a0d831cf3891ab30422e3f7d08d10af9a13f2/src/puffo_agent/agent/harness/driver_authority_server.py) · [acp.py](https://github.com/puffo-ai/puffo-agent/blob/0e8a0d831cf3891ab30422e3f7d08d10af9a13f2/src/puffo_agent/agent/harness/drivers/acp.py)


Glimmer0x:
## Acceptance archive for `695972bb`

**Status: PASS from two independent reviewer seats, with zero evidence transfer.** The PR head remains exactly `695972bbd8d5aeea6657e942d8a3864198465c0c` on base `069f7d75d27a65364f7d4d90208fec79e679b001`.

Both reviewers independently verified:

- the six-path port classification and the four-file PR diff;
- focused authority-server tests and the full test suite;
- D6 clean reproduction at LingTai `d69dacc400cf277fc380af10193c6f9af6e44779`, yielding `0/1/1`;
- all four D2–D5 mutations, with the expected frozen assertions failing and recovery returning clean;
- the current ACP integration preserves the constrained authority semantics and does not weaken the adapted assertions.

GitHub CI and CodeQL are also green for this exact head.

### Table B — non-blocking durability items

#### D6 evidence-scope correction (added after acceptance)

The archived D6 `0/1/1` result is real, but its demonstrated scope is narrower than the original wording implied. The oracle's `RecordingProvider.generate` path is synchronous and same-thread; it does not cross an executor or thread boundary between admission and actual provider I/O. The LingTai owner subsequently reported a reproducible `#1545` production-path probe in which provider I/O crosses `APICallGate`'s thread pool without propagating the `ContextVar`, so the actual I/O observes `audit_id=None`. That LingTai-side finding is recorded here as a reproducible probe in the other repository, not as an independent reproduction by this PR's reviewers.

Accordingly, **D6 PASS here means that calls through the named `generate()` entry point have correct counts and cross-side audit binding when there is no executor/thread hop between admission and actual provider I/O.** Its execution-boundary coverage is:

- **synchronous, same-thread admission → provider I/O: verified** (this is the oracle's path by construction);
- **executor/thread-pool hop: not exercised** by this oracle; the LingTai `#1545` probe reports lost `ContextVar` propagation and `audit_id=None` at I/O, but neither reviewer independently reproduced that other-repository probe;
- **async task boundary: not exercised**;
- **cross-process boundary: not exercised**.

This narrows the stated coverage of the LingTai-side evidence only. Inspection of `driver_authority_server.py` found no `ContextVar`/`contextvars` use: the authority server generates the audit id and writes it directly into the response frame within the same request handling, without implicit context propagation. No Puffo code or reviewed byte changed, and the two-seat PASS for `695972bb` remains intact.

For future acceptance oracles, **"which execution boundaries does this evidence cover?"** should be an explicit acceptance item. Synchronous/same-thread, thread pool, async task, and cross-process must each be marked **verified** or **not exercised**. A green synchronous oracle looks identical whether or not those other boundaries are covered; only the explicit checklist makes the scope durable. This is an evidence-scope disclosure requirement, not an implementation or authority boundary.

#### Cross-driver spawn-discipline guard

Current main removed the former shared framed-child spawn chokepoint and its cross-driver structural guard tests. The authority-specific D1 properties are still directly tested and passed by both reviewers for this port, so this is **not a current acceptance defect**.

The repository currently has **no mechanical cross-driver enforcement** for this discipline: `pi`, `codex`, `opencode`, `claude_code`, and `acp` each spawn directly. A future driver can therefore be added without any test forcing review of its environment-replacement or fd-inheritance behavior. Merely planning to remember this when another driver arrives would be a silent-failure-prone convention, not a boundary.

The cheapest recorded rebuilding shape is an audit gate: a test asserting that the discovered driver-module set exactly equals a checked-in expected list. Adding a driver would then make CI fail until the list changes, placing that change in the review diff and forcing an explicit spawn-discipline review. This would be a **change threshold and audit signal, not an authority boundary**, because reviewers could still approve the list update. It is recorded for a separate future change and is not implemented by this PR.

### Environment note

Both independent reviewers completed the full suite with zero failures (`CODEX_HOME` unset). The delivery author saw seven Codex host-MCP/config failures, then reproduced the same seven failures on untouched `origin/main`; neither reviewer reproduced them. This supports the recorded classification as an author-environment/baseline issue rather than a port regression.

This archives acceptance only. Merge remains an explicit owner decision; no merge was performed.

src/puffo_agent/agent/harness/driver_authority_server.py
```diff
@@ -0,0 +1,421 @@
...
+
+        def read_exact(count: int) -> bytes:
+            while len(record.buffer) < count:
+                data, ancdata, flags, _ = record.server_socket.recvmsg(
+                    MAX_FRAME_BYTES + 4,
+                    socket.CMSG_SPACE(array.array("i", [0]).itemsize),
+                )
+                if not data:
+                    raise EOFError
+                if flags & socket.MSG_CTRUNC:
```
hellohanchen: **A1 — blocker: descriptors are leaked into the Driver when ancillary data is truncated.**

This check runs **before** the `ancdata` harvest loop below it. On truncation the kernel has already *installed* the descriptors that fit, so raising here means they never reach `received_fds` — and the `finally: for fd in received_fds: os.close(fd)` cannot close them. They leak into the Driver process and outlive the endpoint.

The ancillary buffer is `socket.CMSG_SPACE(4)` = **24 bytes**; `CMSG_LEN` for 1/2/3 fds is 20/24/28. So two descriptors fit exactly and three truncate:

```
sent 1 fd(s): MSG_CTRUNC=False  installed=1  harvested=1  -> LEAKED=0
sent 2 fd(s): MSG_CTRUNC=False  installed=2  harvested=2  -> LEAKED=0
sent 3 fd(s): MSG_CTRUNC=True   installed=2  harvested=2  -> LEAKED=2
sent 8 fd(s): MSG_CTRUNC=True   installed=2  harvested=2  -> LEAKED=2
```

End-to-end through the real `DriverAuthorityServer`, diffing `/proc/self/fd` across the whole exchange including endpoint teardown:

```
P1 fds present after teardown that were not before: {'4': '/dev/null', '6': '/dev/null'}
P1 LEAKED /dev/null descriptors: 2
```

The leaked descriptors are **attacker-chosen**: a constrained child hands the Driver a live reference to any descriptor it holds, which the Driver then never closes and never accounts for. That is what makes this security-relevant rather than untidy.

It is two per endpoint on its own, because the `ValueError` tears the endpoint down — but A2 supplies unlimited endpoints.

Note this is exactly the shape that survives review-by-reading: the guard is present and its condition is right, only its ordering relative to the harvest is wrong.

src/puffo_agent/agent/harness/driver_authority_server.py
```diff
@@ -0,0 +1,421 @@
...
+        }:
+            return self._mismatch(record, "authorize_derived_launch"), None
+        child_binding = _EndpointBinding(
+            launch_id=f"launch_{uuid.uuid4().hex}",
+            role="derived",
+            parent_launch_id=binding.launch_id,
+            depth=1,
+            capability=capability,
+        )
+        child_record, child_endpoint = self._issue_endpoint(child_binding)
```
hellohanchen: **A2 — blocker: a root child can issue unbounded derived launches; nothing is capped or pruned.**

`authorize_derived_launch` is child-initiated and there is no limit on how many times a root may call it. Each grant allocates a `socketpair` plus a dedicated `threading.Thread` (`_start_record` on the next line), appends to `self._records` (`:151`), and appends to `self._audits` (`:322`). None of the three is ever pruned — `_serve` sets `record.state = CLOSED` on exit but leaves the record in the list for the life of the process.

Driven through the real server, holding the grandchild descriptors open (releasing them is *cooperative* behavior; a hostile root will not):

```
P2 after 200 derived launches (fds held): records=201 threads 2->202 fds 7->407 audits=200
```

200 requests cost the Driver 200 threads and 400 descriptors. At a common `RLIMIT_NOFILE` of 1024 that is roughly 500 requests to exhaust the Driver's descriptor table — and the Driver is the trusted supervisor, so what dies is the daemon, not the misbehaving child.

Scope matters for the fix, so being precise: the thread/descriptor exhaustion needs a hostile root, but **`_records` and `_audits` grow without bound even under a perfectly well-behaved one**. Those are two different fixes — a cap on concurrent derived endpoints, and pruning/bounding the two lists.

This sits inside the threat model the module states in its own docstring, which is why I am filing it as a blocker rather than hardening.

src/puffo_agent/agent/harness/driver_authority_server.py
```diff
@@ -0,0 +1,421 @@
...
+            del record.buffer[:count]
+            return value
+
+        try:
+            size = struct.unpack("!I", read_exact(4))[0]
+            if size <= 0 or size > MAX_FRAME_BYTES:
+                raise ValueError("authority request frame is out of bounds")
+            raw = read_exact(size)
+            if received_fds:
+                raise ValueError("authority request must not carry descriptors")
```
hellohanchen, line 482: **A3 — nit: this guard, and the `MSG_CTRUNC` guard above it, are both unpinned.**

Two mutations survive all 15 tests:

- `if received_fds:` → `if False:` (never reject descriptors on a request)
- `if flags & socket.MSG_CTRUNC:` → `if False:` (ignore truncation)

No test in this file sends ancillary data on a *request* at all — the `_send` helper only ever calls `sendall`. That is the coverage hole A1 hides in.

One test that sends 3 descriptors with a request and asserts the Driver's `/proc/self/fd` count is unchanged after teardown would pin A1's fix and both guards together.

src/puffo_agent/agent/harness/driver_authority_server.py
```diff
@@ -0,0 +1,421 @@
...
+        binding = record.binding
+        if binding.role != "root":
+            return self._decision(
+                record,
+                "authorize_derived_launch",
+                "denied",
+                "nested_derived_launch_denied",
+            ), None
+        capability = request.get("capability")
+        if request.get("launch_id") != binding.launch_id or capability not in {
```
hellohanchen, line 296: **A4 — nit: this capability allowlist is unpinned.**

Dropping `capability not in {"daemon", "avatar"}` (keeping only the `launch_id` check) survives all 15 tests.

In the current code the real blast radius is limited: `provider_capability` collapses anything that is not `"daemon"` to `"avatar_child"`, so an unlisted capability could not gain provider authority. But the `hello` response echoes `binding.capability` back to the child verbatim, so what an unlisted value means to the consumer depends on LingTai's adapter — outside this repo, and I could not check it.

Two assertions — `"daemon"` and `"avatar"` accepted, anything else denied with `endpoint_binding_mismatch` — would close it.

src/puffo_agent/agent/harness/driver_authority_server.py
```diff
@@ -0,0 +1,421 @@
...
+
+    def _claim(self, record: _EndpointRecord) -> dict[str, Any]:
+        with self._lock:
+            if record.state is not _LeaseState.ISSUED:
+                return self._record_decision_locked(
+                    record, "hello", "denied", "endpoint_already_claimed"
+                )
+            if self._claim_probe is not None:
+                self._claim_probe()
+            record.state = _LeaseState.CLAIMED
```
hellohanchen, line 272: **A5 — nit: successful claims are not audited.**

The denied path a few lines up appends an `AuthorityAuditRecord` (`endpoint_already_claimed`), but this success path returns the binding without one. Every other decision in the server is audited.

For an authority server whose audit trail is the acceptance oracle, "endpoint N was claimed at all" seems like the one event you would most want in the record — it is the only evidence that a given endpoint was ever taken up, and by implication when.

hellohanchen (commented):
## Review — constrained LingTai driver authority @ `695972bbd8d5aeea6657e942d8a3864198465c0c`

Verdict: **BLOCK** — two resource-exhaustion defects reachable from the untrusted side of this module's own trust boundary (A1, A2). Both are probe-reproduced, neither is a design problem: the authority model itself holds up well.

Base for the diff: `069f7d75d27a65364f7d4d90208fec79e679b001` (merge-base), +1077/−12. CI is green on this SHA (7/7), and so is my local run: full suite on py3.12 with `tests/test_pi_tool_bridge.py` deselected gives **3763 tests, 0 failures, 0 errors, 4 skipped** (JUnit XML counts, not the exit code); that one file fails identically on the merge-base, so it is environmental on my host. `ruff check src tests scripts`: clean. Nothing below is a test failure — both blockers are reachable states the suite does not construct.

### The core design is sound, and the tests pin it

The central claim — *"All authority stays in the Driver-owned endpoint record: role, lineage, depth, capability, and provider are never accepted from child process state"* — holds under mutation. I deleted each load-bearing guard one at a time (baseline 15 passed):

| # | mutation | site | outcome |
|---|---|---|---|
| N3 | `if binding.role != "root"` → `if False` (let a derived child spawn a nested child) | `:227` | killed |
| N5 | `if record.state is not _LeaseState.ISSUED` → `if False` (re-claim an endpoint) | `:213` | killed |
| N6 | drop `request.get("capability") == binding.provider_capability` | `:267` | killed |
| N7 | drop `request.get("launch_id") == binding.launch_id` | `:265` | killed |
| N1 | `if received_fds: raise` → never reject descriptors on a request | `:392-393` | **SURVIVED** |
| N2 | `if flags & socket.MSG_CTRUNC: raise` → ignore truncation | `:374-375` | **SURVIVED** |
| N4 | drop the `capability not in {"daemon", "avatar"}` allowlist | `:237-240` | **SURVIVED** |

Lineage, depth, single-use claim, and per-call binding are all genuinely covered. The sanitization on the client side is right too: `DRIVER_AUTHORITY_FD_ENV` is popped from `plan.environment` on **every** spawn path — the generic local spawn passes the popped `environment`, and the `process_factory` path substitutes a `replace(plan, ...)` — so a caller cannot smuggle in an ambient descriptor. `_uses_lingtai_driver_authority` correctly handles both `--profile puffo-v0` and `--profile=puffo-v0`, and bounds the lookahead.

---

### A1 — descriptors are leaked into the Driver when ancillary data is truncated

`_recv_frame` checks `MSG_CTRUNC` **before** it harvests `ancdata`:

```python
if flags & socket.MSG_CTRUNC:
    raise ValueError("authority request ancillary data was truncated")
for level, kind, raw in ancdata:      # <-- never reached on truncation
    ...
    received_fds.extend(items.tolist())
```

On truncation the kernel has **already installed** the descriptors that fit. Because the raise happens first, they never reach `received_fds`, so the `finally: for fd in received_fds: os.close(fd)` cannot close them. They leak into the Driver process and outlive the endpoint.

The ancillary buffer is `socket.CMSG_SPACE(4)` = **24 bytes**, and `CMSG_LEN` for 1/2/3 fds is 20/24/28 — so two descriptors fit exactly and three truncate. Measured:

```
sent 1 fd(s): MSG_CTRUNC=False  installed=1  harvested=1  -> LEAKED=0
sent 2 fd(s): MSG_CTRUNC=False  installed=2  harvested=2  -> LEAKED=0
sent 3 fd(s): MSG_CTRUNC=True   installed=2  harvested=2  -> LEAKED=2
sent 8 fd(s): MSG_CTRUNC=True   installed=2  harvested=2  -> LEAKED=2
```

And end-to-end through the real `DriverAuthorityServer`, comparing `/proc/self/fd` before and after, including endpoint teardown:

```
P1 fds present after teardown that were not before: {'4': '/dev/null', '6': '/dev/null'}
P1 LEAKED /dev/null descriptors: 2
```

Two descriptors of the child's choosing, permanently installed in the Driver. The endpoint does tear down afterwards, so it is two per endpoint rather than an unbounded drip on its own — but A2 supplies the endpoints.

Worth stressing what makes this the security-relevant half rather than a tidiness issue: the leaked descriptors are *attacker-chosen*. A constrained child that has some other descriptor open hands the Driver a live reference to it that the Driver never closes and never accounts for.

### A2 — a root child can issue unbounded derived launches; nothing is capped or pruned

`authorize_derived_launch` is child-initiated, and there is no limit on how many times a root may call it. Each grant allocates a `socketpair` plus a dedicated `threading.Thread`, appends to `self._records`, and appends to `self._audits`. None of those three is ever pruned — `_serve` sets `record.state = CLOSED` on exit but leaves the record in the list forever.

Driven through the real server, with the grandchild descriptors held open (which is what a hostile root does — releasing them is cooperative behavior):

```
P2 after 200 derived launches (fds held): records=201 threads 2->202 fds 7->407 audits=200
```

200 requests cost the Driver 200 threads and 400 descriptors. At a common `RLIMIT_NOFILE` of 1024 that is roughly 500 requests to exhaust the Driver's descriptor table — and the Driver is the trusted supervisor, so what dies is the whole daemon, not the misbehaving child.

I want to be precise about scope, since it changes the fix: even a *well-behaved* root leaves `_records` and `_audits` growing without bound for the life of the process. The thread/descriptor exhaustion needs a hostile root; the memory growth does not.

Both of these sit squarely inside the threat model this module states for itself in its own docstring, which is why I am filing them as blockers rather than hardening suggestions.

### A3 — nit: the two descriptor-handling guards are unpinned

N1 and N2 above. No test in `test_driver_authority_server.py` sends ancillary data on a *request* at all — the `_send` helper only ever calls `sendall`, so both `MSG_CTRUNC` handling and the "must not carry descriptors" rejection are untested. That is also why A1 survived review-by-reading: the guard looks present and correct, and only its ordering relative to the harvest is wrong.

A test that sends 3 descriptors with a request and asserts the Driver's `/proc/self/fd` count is unchanged after teardown would pin A1's fix and both guards at once.

### A4 — nit: the derived-capability allowlist is unpinned

N4 above: dropping `capability not in {"daemon", "avatar"}` survives all 15 tests. In the current code the real blast radius is limited — `provider_capability` collapses anything that is not `"daemon"` to `"avatar_child"`, so an unlisted capability could not gain provider authority. But the `hello` response echoes `binding.capability` back to the child verbatim, so what an unlisted value means to the consumer depends on LingTai's adapter, which is outside this repo and which I could not check. Two assertions — `"daemon"` and `"avatar"` accepted, anything else denied with `endpoint_binding_mismatch` — would close it.

### A5 — nit: successful claims are not audited

`_claim` audits the *denied* path (`endpoint_already_claimed`) but the success path returns the binding without appending an `AuthorityAuditRecord`. Every other decision in the server is audited. For an authority server whose audit trail is the acceptance oracle, "endpoint N was claimed at all" seems like the one event you would most want in the record.

---

`scripts/verify_lingtai_driver_authority.py` I read but did not execute: it requires an exact LingTai checkout via `--lingtai-src`, which I do not have. Its baseline/replay/fanout structure is the right shape for a delivery oracle. Flagging that it is unrun on my side rather than implying coverage I do not have.

All experiments were run in my own worktree and reverted; `git status --porcelain` is empty at `695972bb`.

PR #306, opened 2026-09-01
feat(mcp): monid_spend tool to fetch paid external data via the server

KurtneySh:
## Goal

Give an agent one generic tool to fetch paid, read-only external data through Monid, with the server acting as the spend-control middle layer. The agent never holds the Monid API key or money.

This is the agent-side piece of the Monid spend gateway; the server-side endpoint it forwards to is a separate change.

## Why

- An agent should be able to buy external data (e.g. product data, reviews) without ever holding the paid API key or an unbounded budget.
- The budget, the key, and the receipt all live on the server; the agent just asks and gets a result plus what it cost.

## Scope

Adds one MCP tool, `monid_spend`, in its own registrar module `core_monid_tools` wired into `register_core_tools` (matching the per-domain tool-module layout), and registers it in the tool allowlist (`config.py`):

- The agent calls it with a natural-language `query`, the structured `input` for the capability, and a hard per-call `max_cost_micro` ceiling (micro-dollars, 1,000,000 = $1). The server discovers a matching read-only capability, checks the agent's budget, pays Monid, and returns the result.
- It forwards to `POST /v2/monid/spend` via the **native signed client** — no new auth path. The agent authenticates as itself (subkey signature); the server resolves the agent from that identity. The keyless bridge transport is out of scope here (that path is unsigned and could not reach the subkey-gated spend route).
- It validates the query and the positive cost ceiling before any network call — a UX pre-guard; the real cap is enforced server-side. It reports a still-resolving (pending reconcile) spend clearly instead of as a fake result, and surfaces the server's own mapped error message on failure rather than a raw HTTP blob.

## How to verify

Needs Python ≥ 3.11 and the `mcp` v1 SDK (the code uses `FastMCP`; `mcp` v2 renamed it, so pin `mcp<2`):

```
pytest tests/test_monid_tools.py
pytest
```

Result on my run: **`tests/test_monid_tools.py` 5 passed** (forwards-and-formats-result, reports-pending-reconcile, rejects-bad-input-without-calling-the-server, surfaces-server-error-message, keyless-unsupported). Full suite **3786 passed, 4 skipped**. No linter/type-checker is configured in this repo; the tests are the gate.

This branch is based on current `main`. The real-money end-to-end run is done separately on the maintainer's own machine with the key set locally; this branch is built and tested against a fake client.

**This is a draft and is not intended to merge as-is** — pending code-level review.

### DIFF
--- a/src/puffo_agent/mcp/config.py
+++ b/src/puffo_agent/mcp/config.py
@@ -74,6 +74,7 @@ PUFFO_CORE_TOOL_NAMES = (
     "add_dm_allowlist",
     "update_dm_blocklist",
     "refresh",
+    "monid_spend",
     # M3 memory tools (registered by mcp.memory_tools).
     "create_note",
     "patch_note",
--- a/src/puffo_agent/mcp/core_monid_tools.py
+++ b/src/puffo_agent/mcp/core_monid_tools.py
@@ -1,0 +1,127 @@
+"""Monid spend MCP tool registration.
+
+One generic tool, ``monid_spend``, lets an agent fetch paid, read-only
+external data through Monid with the server as the spend-control middle
+layer. The agent never holds the Monid key or money: it forwards to
+``POST /v2/monid/spend`` via the native signed client, and the server
+discovers a capability, checks the budget, pays Monid, and returns the
+result. Step one targets native (key-holding) agents; the keyless bridge
+transport is out of scope here (that path is unsigned and could not reach
+the subkey-gated spend route).
+"""
+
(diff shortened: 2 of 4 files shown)


KurtneySh:
Superseded by #308. The single-shot tool here required a request template hardcoded for one endpoint; #308 replaces it with generic monid_prepare + monid_spend that build the request from each endpoint's own schema. #308 is a single clean commit in which the removed template never appears. Closing in favor of the general PR.

PR #309, opened 2026-09-02
feat(cli): add cross-platform daemon autostart

Glimmer0x:
## Summary
- add `puffo-agent autostart enable|disable|status`
- register per-user startup with launchd on macOS, systemd user units on Linux, and HKCU Run on Windows
- support optional Linux linger and preserve actionable disable/reload failures
- quote and escape systemd environment and executable values safely

## Validation
- `env -u CODEX_HOME uv run --extra dev pytest -q tests/test_autostart.py tests/test_cli_parser.py` (24 passed)
- `env -u CODEX_HOME uv run --extra dev ruff check src/puffo_agent/portal/autostart.py src/puffo_agent/portal/cli.py src/puffo_agent/portal/cli_parser.py tests/test_autostart.py tests/test_cli_parser.py`
- `env -u CODEX_HOME uv run --extra dev python tools/check_python_structure.py`
- `git diff --check f0691564..af31e9f`
- `env -u CODEX_HOME uv run --extra dev pytest -q` (100% passed, 2 expected skips)

### DIFF
--- a/src/puffo_agent/portal/autostart.py
+++ b/src/puffo_agent/portal/autostart.py
@@ -66,7 +66,15 @@ def _service_env() -> dict[str, str]:
     # shell export the user enabled from, and "defaulted at enable time,
     # different default at boot" is exactly the class of silent
     # divergence this feature must not introduce.
-    return {"PUFFO_AGENT_HOME": str(home_dir())}
+    env = {"PUFFO_AGENT_HOME": str(home_dir())}
+    # Login services get a minimal PATH. Known harness CLIs ride
+    # cli_bin's broader-than-PATH resolver, but generic/ACP runtime
+    # configs spawn bare commands (``opencode acp``, ...) exactly as
+    # written — persist the enable-time PATH so those still resolve.
+    path = os.environ.get("PATH")
+    if path:
+        env["PATH"] = path
+    return env
 
 
 # ── result / status types ────────────────────────────────────────────────────
@@ -411,16 +419,36 @@ def windows_run_command() -> list[str]:
     return command
 
 
+def _windows_login_home() -> Path:
+    """The PUFFO_AGENT_HOME a login-started daemon will see: the
+    persisted per-user value if any (where ``setx`` writes), else the
+    default. Comparing against the current effective home — not against
+    "is the variable set" — is what keeps ``setx`` + re-run from being
+    refused forever."""
+    winreg = _load_winreg()
+    try:
+        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment") as key:
+            value, _ = winreg.QueryValueEx(key, "PUFFO_AGENT_HOME")
+    except (FileNotFoundError, OSError):
+        value = ""
+    if str(value):
+        return Path(str(value)).expanduser()
+    return Path.home() / ".puffo-agent"
+
+
 def enable_windows() -> ActionResult:
-    if os.environ.get("PUFFO_AGENT_HOME"):
-        # Run-key entries carry no per-entry environment; an override
-        # enabled here would silently start the daemon against the
-        # default home at next login. Refuse rather than diverge.
+    login_home = _windows_login_home()
+    if login_home != home_dir():
+        # Run-key entries carry no per-entry environment; enabling under
+        # a session-only override would silently start the daemon
+        # against a different home at next login. Refuse rather than
+        # diverge.
         return ActionResult(
             False,
             [
-                "PUFFO_AGENT_HOME is overridden in this environment; a Run-key "
-                "entry cannot carry it. Persist it first (`setx "
+                f"PUFFO_AGENT_HOME resolves to {home_dir()} here but would "
+                f"resolve to {login_home} at login; a Run-key entry cannot "
+                "carry the override. Persist it first (`setx "
                 "PUFFO_AGENT_HOME ...`) or unset it, then re-run "
                 "`puffo-agent autostart enable`."
             ],
--- a/src/puffo_agent/portal/autostart.py
+++ b/src/puffo_agent/portal/autostart.py
@@ -92,7 +92,14 @@ Runner = Callable[[list[str]], subprocess.CompletedProcess]
 
 
 def _run(cmd: list[str]) -> subprocess.CompletedProcess:
-    return subprocess.run(cmd, capture_output=True, text=True, check=False)
+    try:
+        return subprocess.run(cmd, capture_output=True, text=True, check=False)
+    except FileNotFoundError:
+        # No launchctl/systemctl on this box (e.g. a non-systemd distro):
+        # report it like any failed call instead of tracebacking.
+        return subprocess.CompletedProcess(
+            cmd, 127, stdout="", stderr=f"{cmd[0]}: command not found"
+        )
 
 
 def _failure_detail(proc: subprocess.CompletedProcess) -> str:
@@ -212,7 +219,11 @@ def status_macos(run: Runner = _run) -> AutostartStatus:
 
 
 def systemd_unit_path() -> Path:
-    return Path.home() / ".config" / "systemd" / "user" / SYSTEMD_UNIT_NAME
+    # systemd reads user units from $XDG_CONFIG_HOME/systemd/user and,
+    # like systemd itself, a non-absolute XDG_CONFIG_HOME is ignored.
+    xdg = os.environ.get("XDG_CONFIG_HOME", "")
+    base = Path(xdg) if os.path.isabs(xdg) else Path.home() / ".config"
+    return base / "systemd" / "user" / SYSTEMD_UNIT_NAME
 
 
 def systemd_unit(command: list[str], env: dict[str, str]) -> str:
@@ -254,14 +265,26 @@ def enable_linux(run: Runner = _run, *, linger: bool = False) -> ActionResult:
     path = systemd_unit_path()
     content = systemd_unit(daemon_command(), _service_env())
     changed = not (path.exists() and path.read_text() == content)
+    # Queried before `enable --now` can start anything: an already-running
+    # service keeps executing the old definition until restarted below.
+    was_active = (
+        changed
+        and run(["systemctl", "--user", "is-active", SYSTEMD_UNIT_NAME]).returncode == 0
+    )
     if changed:
         path.parent.mkdir(parents=True, exist_ok=True)
         path.write_text(content)
     lines = [f"wrote {path}" if changed else f"unit unchanged at {path}"]
-    for cmd in (
+    commands: list[list[str]] = [
         ["systemctl", "--user", "daemon-reload"],
         ["systemctl", "--user", "enable", "--now", SYSTEMD_UNIT_NAME],
-    ):
+    ]
+    if was_active:
+        # `enable --now` only starts an inactive unit — the launchd
+        # branch's bootout/bootstrap equivalent for a changed definition.
+        commands.append(["systemctl", "--user", "restart", SYSTEMD_UNIT_NAME])
+        lines.append("restarting the running service to apply the changed definition")
+    for cmd in commands:
         proc = run(cmd)
         if proc.returncode != 0:
             lines.append(
@@ -285,9 +308,26 @@ def enable_linux(run: Runner = _run, *, linger: bool = False) -> ActionResult:
 def disable_linux(run: Runner = _run) -> ActionResult:
     path = systemd_unit_path()
     if not path.exists():
-        # Nothing configured; a failing systemctl here (e.g. no user
-        # manager at all) leaves nothing behind to report.
-        return ActionResult(True, ["autostart was not enabled"])
+        # The unit file can be gone while systemd still runs the service
+        # (deleted by hand, an earlier config root, ...). "Not enabled"
+        # over a live daemon would be a lie — stop it. A failing
+        # is-active (including: no user manager at all) means nothing is
+        # running and nothing is left behind to report.
+        if run(["systemctl", "--user", "is-active", SYSTEMD_UNIT_NAME]).returncode != 0:
+            return ActionResult(True, ["autostart was not enabled"])
+        proc = run(["systemctl", "--user", "stop", SYSTEMD_UNIT_NAME])
+        if proc.returncode != 0:
+            return ActionResult(
+                False,
+                [
+                    f"no unit file at {path}, but the service is running and "
+                    f"`systemctl --user stop` failed: {_failure_detail(proc)}"
+                ],
+            )
+        return ActionResult(
+            True,
+            [f"no unit file at {path}; stopped the still-loaded service"],
+        )
     proc = run(["systemctl", "--user", "disable", "--now", SYSTEMD_UNIT_NAME])
     if proc.returncode != 0:
         # Deleting the unit anyway would orphan a still-running /
--- a/src/puffo_agent/portal/autostart.py
+++ b/src/puffo_agent/portal/autostart.py
@@ -71,9 +71,15 @@ def _service_env() -> dict[str, str]:
     # cli_bin's broader-than-PATH resolver, but generic/ACP runtime
     # configs spawn bare commands (``opencode acp``, ...) exactly as
     # written — persist the enable-time PATH so those still resolve.
-    path = os.environ.get("PATH")
-    if path:
-        env["PATH"] = path
+    # Absolute entries only: an empty or relative entry resolves from
+    # the cwd, which an interactive shell controls but a daemon doesn't.
+    entries = [
+        entry
+        for entry in os.environ.get("PATH", "").split(os.pathsep)
+        if entry and os.path.isabs(entry)
+    ]
+    if entries:
+        env["PATH"] = os.pathsep.join(entries)
     return env
 
 
--- a/src/puffo_agent/portal/autostart.py
+++ b/src/puffo_agent/portal/autostart.py
@@ -1,0 +1,452 @@
+"""Login-time autostart registration for the puffo-agent daemon.
+
+``puffo-agent autostart enable`` registers the daemon to start
+automatically after the user logs in; ``disable`` removes the
+registration; ``status`` reports both what is configured on disk and
+what the platform's service manager currently has loaded — the two can
+diverge (file present but never loaded, interpreter upgraded away, ...)
+and conflating them would report an autostart that will not fire.
+
+This is deliberately per-user, login-session autostart — not a
+pre-login system service. Credentials the daemon needs (Keychain
+entries, ``~/.pi``, OpenCode auth) live in the user session, so a
+system-level service would start into an environment where every
+harness reports signed-out. On Linux, ``--linger`` optionally upgrades
+to boot-time start via ``loginctl enable-linger`` (best-effort; not all
+systems permit it).
+
+Platform mechanisms:
+
+- macOS: a LaunchAgent plist. launchd supervises the foreground
+  ``start`` directly; ``KeepAlive.SuccessfulExit=false`` restarts the
+  daemon after a crash but not after a clean ``puffo-agent stop`` (and
+  not after the pid-lock "already running" no-op, which exits 0).
+- Linux: a systemd user unit with ``Restart=on-failure``.
+- Windows: an ``HKCU`` Run key. Windows offers no supervision on this
+  path — the daemon starts at login and is not restarted on crash;
+  ``status`` says so rather than implying launchd/systemd semantics.
+
+Every registration embeds the absolute interpreter path
+(``sys.executable -m puffo_agent.portal.cli start`` — the repo's
+established PATH-free re-invocation) and the effective
+``PUFFO_AGENT_HOME``, because login-time environments have neither the
+user's shell PATH nor its exports.
+"""
+
+from __future__ import annotations
+
+import os
+import plistlib
+import shlex
+import subprocess
+import sys
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from .state import home_dir
+
+LAUNCHD_LABEL = "ai.puffo.agent"
+WINDOWS_RUN_VALUE = "PuffoAgent"
+_WINDOWS_RUN_KEY = r"Software\Microsoft\Windows\CurrentVersion\Run"
+SYSTEMD_UNIT_NAME = "puffo-agent.service"
+
+
+def daemon_command() -> list[str]:
+    """The exact argv a service manager should supervise."""
+    return [sys.executable, "-m", "puffo_agent.portal.cli", "start"]
+
+
+def autostart_log_path() -> Path:
+    return home_dir() / "logs" / "autostart.log"
+
+
+def _service_env() -> dict[str, str]:
+    # Always explicit: a login-time environment does not inherit the
+    # shell export the user enabled from, and "defaulted at enable time,
+    # different default at boot" is exactly the class of silent
+    # divergence this feature must not introduce.
+    return {"PUFFO_AGENT_HOME": str(home_dir())}
+
+
+# ── result / status types ────────────────────────────────────────────────────
+
+
+@dataclass
+class ActionResult:
+    ok: bool
+    lines: list[str] = field(default_factory=list)
+
+
+@dataclass
+class AutostartStatus:
+    """``configured`` is what's on disk; ``active`` is what the service
+    manager currently has loaded. Both false ⇒ disabled."""
+
+    configured: bool
+    active: bool
+    lines: list[str] = field(default_factory=list)
+
+
+Runner = Callable[[list[str]], subprocess.CompletedProcess]
+
+
+def _run(cmd: list[str]) -> subprocess.CompletedProcess:
+    return subprocess.run(cmd, capture_output=True, text=True, check=False)
+
+
+def _failure_detail(proc: subprocess.CompletedProcess) -> str:
+    detail = (proc.stderr or proc.stdout or "").strip()
+    return detail.splitlines()[0] if detail else f"exit code {proc.returncode}"
+
+
+# ── macOS (launchd LaunchAgent) ──────────────────────────────────────────────
+
+
+def launchagent_path() -> Path:
+    return Path.home() / "Library" / "LaunchAgents" / f"{LAUNCHD_LABEL}.plist"
+
+
+def launchagent_plist(
+    command: list[str], log_path: Path, env: dict[str, str]
+) -> bytes:
+    return plistlib.dumps(
+        {
+            "Label": LAUNCHD_LABEL,
+            "ProgramArguments": command,
+            "RunAtLoad": True,
+            "KeepAlive": {"SuccessfulExit": False},
+            "StandardOutPath": str(log_path),
+            "StandardErrorPath": str(log_path),
+            "EnvironmentVariables": env,
+        },
+        sort_keys=True,
+    )
+
+
+def _launchd_domain() -> str:
+    return f"gui/{os.getuid()}"
+
+
+def _launchd_loaded(run: Runner) -> bool:
+    return run(["launchctl", "print", f"{_launchd_domain()}/{LAUNCHD_LABEL}"]).returncode == 0
+
+
+def enable_macos(run: Runner = _run) -> ActionResult:
+    path = launchagent_path()
+    log_path = autostart_log_path()
+    content = launchagent_plist(daemon_command(), log_path, _service_env())
+    if path.exists() and path.read_bytes() == content and _launchd_loaded(run):
+        return ActionResult(True, ["already enabled (LaunchAgent loaded, definition current)"])
+    log_path.parent.mkdir(parents=True, exist_ok=True)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_bytes(content)
+    lines = [f"wrote {path}"]
+    if _launchd_loaded(run):
+        # Definition changed while loaded: launchd only re-reads a plist
+        # across bootout/bootstrap. This restarts a launchd-run daemon.
+        run(["launchctl", "bootout", f"{_launchd_domain()}/{LAUNCHD_LABEL}"])
+        lines.append("reloading changed LaunchAgent definition")
+    proc = run(["launchctl", "bootstrap", _launchd_domain(), str(path)])
+    if proc.returncode != 0:
+        # Older macOS has no bootstrap subcommand.
+        proc = run(["launchctl", "load", "-w", str(path)])
+    if proc.returncode != 0:
+        lines.append(
+            "launchctl could not load the LaunchAgent: "
+            f"{_failure_detail(proc)}. The file is in place; it will "
+            "load at next login, or run `launchctl bootstrap "
+            f"{_launchd_domain()} {path}` manually."
+        )
+        return ActionResult(False, lines)
+    lines.append("daemon will start automatically after login (supervised by launchd)")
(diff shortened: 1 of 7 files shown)


src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+    command = daemon_command()
+    interpreter = Path(command[0])
+    windowless = interpreter.with_name("pythonw.exe")
+    if interpreter.name.lower() == "python.exe" and windowless.exists():
+        command[0] = str(windowless)
+    return command
+
+
+def enable_windows() -> ActionResult:
+    if os.environ.get("PUFFO_AGENT_HOME"):
```
hellohanchen: **U1 — blocker: this refuses for any user with a custom home, and the remedy it prints cannot ever clear it.**

The error tells the user:

> Persist it first (`setx PUFFO_AGENT_HOME ...`) or unset it, then re-run `puffo-agent autostart enable`.

`setx` writes to the persistent **user** environment. Every console opened afterwards inherits it, so `os.environ.get("PUFFO_AGENT_HOME")` is truthy on the retry and this refusal repeats — permanently. The check cannot tell a persisted user-level variable from an ad-hoc `set` in the current shell; both are just "present in `os.environ`".

Driven through the real `enable_windows` with a fake registry:

```
case 1  no PUFFO_AGENT_HOME set             -> ok = True
case 2  PUFFO_AGENT_HOME == default home    -> ok = False
case 3  after `setx` (persisted, inherited) -> ok = False
```

Case 2 is the sharpest: the variable is set to *exactly* the default home, so there is provably no divergence to prevent, and it still refuses. The guard asks "is the variable set" when what it means to ask is "would the login-time value differ from this one".

What lifts this above a Windows-only annoyance: this PR also makes `machine link` enable autostart by default (`cli.py:1303`). So on Windows with a custom home, **every** `puffo-agent machine link` now ends in a stderr warning pointing at a command that will also fail — on the default onboarding path.

To be clear, the *intent* is right: a Run key carries no per-entry environment, and silently starting against the wrong home would be worse. It is the test for that condition that is too broad. Note `test_windows_enable_refuses_unpersistable_home_override` locks in the current version, so it has to move with the fix.

src/puffo_agent/portal/cli.py
```diff
@@ -1270,17 +1289,41 @@ def cmd_link(args: argparse.Namespace) -> int:
...
                 name,
                 open_browser=not args.not_open,
-                code=getattr(args, "code", None),
+                code=args.code,
             )
         )
     except KeyboardInterrupt:
         print("\nlink: cancelled.")
         return 1
+    if link_rc == 0 and not args.no_autostart:
```
hellohanchen, line 1303: **U2 — nit: `--no-autostart` is unpinned.**

Deleting `not args.no_autostart` from this line — so the flag is ignored and linking always enables autostart — survives all 42 tests in `test_autostart.py` / `test_cli_parser.py` / `test_portal_onboarding.py`.

The counterpart mutation (never auto-enabling at all) *is* killed, so the default behavior is covered. It is only the opt-out that is not — and for a flag whose whole purpose is "do not touch my login items", that is the direction worth asserting.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+def daemon_command() -> list[str]:
+    """The exact argv a service manager should supervise."""
+    return [sys.executable, "-m", "puffo_agent.portal.cli", "start"]
+
+
+def autostart_log_path() -> Path:
+    return home_dir() / "logs" / "autostart.log"
+
+
+def _service_env() -> dict[str, str]:
```
hellohanchen, line 64: **U3 — nit: this helper's wiring is unpinned.**

Making `_service_env()` return `{}` survives all 42 tests.

The three tests that assert `PUFFO_AGENT_HOME` reaches the plist and the unit pass the env dict in explicitly (`launchagent_plist(..., {"PUFFO_AGENT_HOME": ...})`), so they exercise the *formatter* and never the wiring. Nothing asserts that `enable_macos` / `enable_linux` actually embed the effective home.

That is precisely the invariant the comment right below this line singles out — *"defaulted at enable time, different default at boot" is exactly the class of silent divergence this feature must not introduce*. Under this mutation the daemon starts at login against the default home, with no error anywhere.

Asserting on the bytes `enable_macos` / `enable_linux` actually write, rather than on what the formatter does with a hand-supplied dict, closes it.

hellohanchen (commented):
## Review — cross-platform daemon autostart @ `7a2b330f6304bb16d22fc16f4489497950f83247`

Verdict: **BLOCK** — one user-facing defect on Windows (U1), reproduced. The rest is nits. This is otherwise the most carefully built of the four PRs I have looked at in this batch: the platform semantics are right, the differences between them are stated rather than papered over, and the tests pin nearly everything that matters.

Base for the diff: `f0691564161193fd40142754c1d63d6678ed3dc6` (merge-base), +1021/−6. CI green 7/7 on this SHA.

### What holds up

Mutation matrix (baseline 42 passed across `test_autostart.py`, `test_cli_parser.py`, `test_portal_onboarding.py`):

| # | mutation | outcome |
|---|---|---|
| T1 | `_systemd_quote` stops escaping `%` | killed |
| T2 | `_systemd_quote` never quotes | killed |
| T3 | `enable_windows` drops the `PUFFO_AGENT_HOME` refusal | killed |
| T4 | `status_macos` drops the stale-interpreter check | killed |
| T5 | `disable_linux` deletes the unit even when `systemctl disable` failed | killed |
| T6 | `enable_macos` skips `bootout` when the definition changed | killed |
| T8 | `cmd_link` never auto-enables autostart | killed |
| T7 | `cmd_link` ignores `--no-autostart` | **SURVIVED** |
| T9 | `_service_env()` returns `{}` | **SURVIVED** |

The systemd quoting is genuinely correct, which is the part I expected to break. `%` is doubled *before* the quoting decision, and `_unit_interpreter` undoes it *after* `shlex.split`, so the round-trip survives paths containing spaces, quotes, backslashes, `%`, and even a literal `%%`. I checked each of those by hand against `_systemd_quote` → `_unit_interpreter`.

The three-way platform split is honest in the way that matters: `status` separates "configured on disk" from "loaded in the service manager", the Windows path states outright that Run entries are unsupervised instead of implying launchd/systemd semantics, and `disable_linux` refuses to delete a unit it could not disable rather than reporting a success that orphans a running service. `KeepAlive.SuccessfulExit=false` is the right launchd choice for not fighting a deliberate `puffo-agent stop`.

Local full suite on py3.12 with `tests/test_pi_tool_bridge.py` deselected: **3815 tests, 0 failures, 0 errors, 4 skipped** (JUnit counts, not the exit code); that file fails identically on the merge-base, so it is environmental on my host. `ruff check src tests`: clean.

---

### U1 — blocker: on Windows, autostart is unreachable for any user with a custom home, and the error's own remedy cannot clear it

`enable_windows` refuses whenever `PUFFO_AGENT_HOME` is present in the environment, and tells the user:

> Persist it first (`setx PUFFO_AGENT_HOME ...`) or unset it, then re-run `puffo-agent autostart enable`.

`setx` writes to the persistent **user** environment. Every console opened afterwards inherits it, so `os.environ.get("PUFFO_AGENT_HOME")` is truthy on the retry and the refusal repeats — permanently. The check has no way to tell a persisted user-level variable from an ad-hoc `set` in the current shell; both are just "present in `os.environ`".

Driven through the real `enable_windows` with a fake registry:

```
case 1  no PUFFO_AGENT_HOME set             -> ok = True
case 2  PUFFO_AGENT_HOME == default home    -> ok = False
case 3  after `setx` (persisted, inherited) -> ok = False

os.environ cannot distinguish persisted-by-setx from ad-hoc shell export:
  both are just PUFFO_AGENT_HOME present -> 'C:\\custom\\puffo'
=> the first remedy the error text offers can never clear the refusal.
```

Case 2 is the sharpest one: the variable is set to *exactly* the default home, so there is provably no divergence to prevent, and it still refuses. The guard is testing "is the variable set" when what it means to ask is "would the login-time value differ from this one".

What raises this above a Windows-only annoyance is that this PR also makes `machine link` enable autostart by default (`cli.py:1303`). So on Windows with a custom home, **every** `puffo-agent machine link` now ends with a stderr warning and a pointer to a command that will also fail. That is the default onboarding path, not an opt-in one.

Worth saying plainly: the *intent* is right — a Run key cannot carry per-entry environment, and silently starting against the wrong home would be worse. It is the test for that condition that is wrong. `test_windows_enable_refuses_unpersistable_home_override` currently locks in the too-broad version, so it will need to move with the fix.

### U2 — nit: `--no-autostart` is unpinned

T7: deleting `not args.no_autostart` from `cli.py:1303` — so the flag is ignored and linking always enables autostart — survives all 42 tests. T8 (never auto-enabling) is killed, so the *default* behavior is covered; only the opt-out a user reaches for when they specifically do not want this is not. For a flag whose entire purpose is "do not touch my login items", that is the one direction worth an assertion.

### U3 — nit: `_service_env()`'s wiring is unpinned

T9: making `_service_env()` return `{}` survives all 42 tests. The three tests that assert `PUFFO_AGENT_HOME` reaches the plist and the unit pass the env dict in explicitly (`launchagent_plist(..., {"PUFFO_AGENT_HOME": ...})`), so they exercise the formatter, never the wiring. Nothing asserts that `enable_macos` / `enable_linux` actually embed the effective home.

That is the invariant this module's own docstring singles out — *"defaulted at enable time, different default at boot" is exactly the class of silent divergence this feature must not introduce* — and under T9 the daemon would start at login against the default home with no error anywhere. Asserting on the bytes `enable_macos`/`enable_linux` write, rather than on what the formatter does with a hand-supplied dict, closes it.

---

All experiments were run in my own worktree and reverted; `git status --porcelain` is empty at `7a2b330f`.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,544 @@
...
+def _windows_login_home() -> Path:
+    """The PUFFO_AGENT_HOME a login-started daemon will see: the
+    persisted per-user value if any (where ``setx`` writes), else the
+    default. Comparing against the current effective home — not against
+    "is the variable set" — is what keeps ``setx`` + re-run from being
+    refused forever."""
+    winreg = _load_winreg()
+    try:
+        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment") as key:
+            value, _ = winreg.QueryValueEx(key, "PUFFO_AGENT_HOME")
```
hellohanchen, line 437: **P3 (non-blocking) — two legitimate ways of persisting `PUFFO_AGENT_HOME` still get refused.**

This function answers "what home will a login-started daemon see?" by reading `HKEY_CURRENT_USER\Environment` and nothing else, and by taking `QueryValueEx`'s value while discarding its type (`value, _ = ...`). Windows resolves a login environment from more than that.

**1. `setx /M` writes to HKLM, not HKCU.** Windows composes the user environment from the machine-wide `HKLM\...\Session Manager\Environment` and then `HKCU\Environment`, so a machine-wide value *is* visible at login. This function does not see it, falls through to the default, and the comparison against the (inherited, therefore overridden) session value refuses — telling the user to run `setx`, which is precisely what they did, with `/M`.

**2. `QueryValueEx` does not expand `REG_EXPAND_SZ`.** The Environment Variables dialog writes that type whenever the value contains a `%VAR%`. The raw string comes back and `Path(str(value)).expanduser()` only handles `~`, so `%USERPROFILE%\elsewhere` is compared literally against a resolved path and never matches. `winreg.ExpandEnvironmentStrings` is the counterpart, and the type is already in hand — it is just being dropped.

`FakeWinreg` cannot express either case (one hive, `OpenKey` ignores its hive argument, `QueryValueEx` always reports `REG_SZ`), so I wrote a double that can and ran all three through the real `enable_windows`:

```
1. setx      (HKCU, REG_SZ)                    -> ENABLED   <- control
2. setx /M   (HKLM, REG_SZ)                    -> REFUSED
3. GUI dialog, %USERPROFILE%  (HKCU, REG_EXPAND_SZ) -> REFUSED
```

(Run on Linux, so the printed paths look POSIX — the registry read and the path comparison are both plain Python and platform-independent, which is what the probe exercises.)

**Why P3 and not higher.** I checked the direction of failure before rating it, because the dangerous outcome here is the opposite one — enabling under a divergence the check failed to notice. Neither case does that. For HKLM: if the machine-wide value is set, the session inherits it too, so `home_dir()` carries it and the comparison errs toward refusing. For `REG_EXPAND_SZ`: an unexpanded literal cannot coincidentally equal a resolved path. Both are spurious refusals, and both are recoverable by running plain `setx`. So this is the same class of bug U1 fixed, narrowed to two persistence mechanisms rather than left wide open — worth closing, not worth holding the merge.

If you do take it: read HKCU first and fall back to HKLM, and pass the value through `winreg.ExpandEnvironmentStrings` when the returned type is `REG_EXPAND_SZ`. `FakeWinreg` will need a second hive and a value type before a test can pin either half.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,544 @@
...
+    # Always explicit: a login-time environment does not inherit the
+    # shell export the user enabled from, and "defaulted at enable time,
+    # different default at boot" is exactly the class of silent
+    # divergence this feature must not introduce.
+    env = {"PUFFO_AGENT_HOME": str(home_dir())}
+    # Login services get a minimal PATH. Known harness CLIs ride
+    # cli_bin's broader-than-PATH resolver, but generic/ACP runtime
+    # configs spawn bare commands (``opencode acp``, ...) exactly as
+    # written — persist the enable-time PATH so those still resolve.
+    path = os.environ.get("PATH")
```
hellohanchen: **P3 (non-blocking) — copying the whole enable-time `PATH` copies entries a daemon should not search.**

The reasoning for persisting a `PATH` is sound and the comment states it well: generic/ACP runtime configs spawn bare commands, those skip `cli_bin`'s broader resolver, and a login service otherwise gets a minimal `PATH`. No argument with the goal.

The part worth a second look is that this takes `os.environ["PATH"]` verbatim. An interactive shell's `PATH` routinely carries entries that make sense for that shell and not for a background daemon started at login — an activated virtualenv's `bin`, a project-local `node_modules/.bin`, a direnv-injected directory. Those are baked into the unit until the next `enable`, and they will silently no longer exist, or worse, be a directory someone else can write to.

The sharpest case is the cheapest to rule out: an **empty** `PATH` entry means the current directory. `PATH=/usr/bin:` and `PATH=/usr/bin::/bin` both contain one, and they are common enough to hit by accident. For an interactive shell whose cwd is wherever the user is standing that is merely bad practice; for a daemon whose working directory is not something the user is thinking about, it means bare-command resolution can pick up a binary from an arbitrary directory.

Filtering to absolute, non-empty entries when building the value would keep everything this change is for and drop the part that is only meaningful to the enabling shell:

```python
path = os.pathsep.join(
    p for p in os.environ.get("PATH", "").split(os.pathsep) if os.path.isabs(p)
)
```

Low severity — it needs an unusual `PATH` to bite, and the feature is opt-in. But it is a one-line narrowing of what a login-started daemon will execute, and the assertion in `test_macos_enable_writes_plist_and_bootstraps` would just need the same filter applied to its expected value.

hellohanchen (commented):
## Round 2 — `11631cd20159b36c7740c9b4801da2c0264aa8ee`

**PASS.** No blocker. Every round-1 finding is closed and pinned, including the two mutants that survived last round. Two P3s remain, both residual cases of the same Windows check — details inline, with a reproduction.

Delta-reviewed against my own round-1 SHA `7a2b330f6304bb16d22fc16f4489497950f83247` (confirmed a linear ancestor; 2 commits, 176 insertions / 16 deletions across 3 files). CI rollup green 7/7.

### The round-1 findings

**U1 — the Windows refusal.** The judgement is now "does the login-time home differ from the effective one", not "is the variable set", which is the right question. I checked the thing that would quietly break it: `_windows_login_home`'s fallback is `Path.home() / ".puffo-agent"` and `home_dir()`'s fallback (`portal/state.py:59-63`) is the same expression, so the no-persisted-value case compares equal instead of refusing forever. The error message now names both resolved paths, which is what makes it actionable.

**U3 — enable actually writing the env.** `test_macos_enable_writes_plist_and_bootstraps` now reads the plist back and asserts `PUFFO_AGENT_HOME == home_dir()` and `PATH == os.environ["PATH"]`. That is the wiring test that was missing; the Linux side asserts against the unit it wrote too.

**U2 / T7 — `--no-autostart`.** You were right about why the old probe could not fail: `_enable_autostart_after_link` swallows the exception into a warning, so a raising monkeypatch was invisible. The recording assertion fixes it, and I confirmed the fix rather than taking it:

```
E9  cli.py: `if link_rc == 0 and not args.no_autostart:` -> `if link_rc == 0:`   FAILED
```

That mutant **survived** in round 1. It is dead now.

**T9 — `_service_env()` wiring.** Also survived in round 1; also dead now (E1 below).

### Mutation results — 9 mutants, 9 killed

Baseline `tests/test_autostart.py` + `tests/test_portal_onboarding.py` → 46 passed.

```
E1  _service_env: drop the enable-time PATH                        FAILED  2 failures
E2  _windows_login_home: always return the default                 FAILED  1 failure
E3  restore the over-wide `if os.environ.get("PUFFO_AGENT_HOME")`  FAILED  2 failures
E4  enable_linux: was_active = False (never restart)               FAILED  1 failure
E5  disable_linux: always report "not enabled"                     FAILED  1 failure
E6  _run: remove the FileNotFoundError catch                       FAILED  1 failure
E7  systemd_unit_path: ignore XDG_CONFIG_HOME                      FAILED  1 failure
E8  systemd_unit_path: accept a relative XDG_CONFIG_HOME           FAILED  1 failure
E9  cli.py: ignore --no-autostart                                  FAILED  1 failure
```

E3 and E9 are the two counterfactuals that matter — they restore the exact round-1 behaviours I reported, and both go red. E8 is worth calling out separately: it is easy to write an XDG test that only proves the absolute case, and this one pins the systemd-matching *rejection* of a relative value too.

`enable_linux`'s restart is ordered correctly and the test asserts the ordering (`daemon-reload` before `restart`), not just the presence — restarting into a stale definition is exactly the failure that "the commands are all there" would miss.

### Two P3s — both the same residual, both reproduced

`_windows_login_home` reads **only `HKEY_CURRENT_USER`** and uses `QueryValueEx`'s value while discarding its type. Two ways of legitimately persisting the variable therefore still get refused. I built a winreg double that models two hives and unexpanded `REG_EXPAND_SZ` (which `FakeWinreg` cannot express) and ran all three cases through the real `enable_windows`:

```
1. setx      (HKCU, REG_SZ)              -> ENABLED   <- control
2. setx /M   (HKLM, REG_SZ)              -> REFUSED
3. GUI dialog with %USERPROFILE%
             (HKCU, REG_EXPAND_SZ)       -> REFUSED
```

Both refusals tell the user to run `setx` — which, in case 2, they already did. Detail and the direction-of-failure analysis are in the inline comment on `_windows_login_home`.

Severity is P3 deliberately: I checked whether either can produce the *dangerous* direction — a silent enable under a real divergence — and neither can. They only produce spurious refusals, and both are recoverable by running plain `setx`. It is the same class of bug U1 fixed, narrowed to two persistence mechanisms rather than left wide open.

There is a third P3 about copying the whole enable-time `PATH`, on `_service_env`.

### A note on why the suite cannot see the P3s

`FakeWinreg` (`tests/test_autostart.py:357`) exposes only `HKEY_CURRENT_USER`, its `OpenKey` ignores the hive argument and returns one flat dict, and `QueryValueEx` always reports `REG_SZ`. So neither gap is expressible as a test against it — not an oversight in any individual test, but a ceiling the double imposes on the whole file. If you take the P3s, the double needs a second hive and a value type before a test can pin the fix.

### Verdict

**PASS.** The P3s are yours to take or leave; I would not hold the merge for them.

Suites: focused set → 46 tests, 0 failures, 0 errors. Full suite → **3820 tests, 0 failures, 0 errors** (counts read from `--junitxml`, not the exit code). `tests/test_pi_tool_bridge.py` is excluded: all 20 of its tests fail on my box with Node `ERR_UNKNOWN_FILE_EXTENSION` loading the bridge harness — environmental here, unrelated to this diff, green in CI. Worktree confirmed clean after every mutation; all mutations and probes reverted.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,550 @@
...
+    # configs spawn bare commands (``opencode acp``, ...) exactly as
+    # written — persist the enable-time PATH so those still resolve.
+    # Absolute entries only: an empty or relative entry resolves from
+    # the cwd, which an interactive shell controls but a daemon doesn't.
+    entries = [
+        entry
+        for entry in os.environ.get("PATH", "").split(os.pathsep)
+        if entry and os.path.isabs(entry)
+    ]
+    if entries:
```
hellohanchen, line 81: **<sub><sub>P3 Badge</sub></sub>  The `if entries:` guard is the security-relevant half, and it is unpinned**

Replacing these two lines with an unconditional `env["PATH"] = os.pathsep.join(entries)` leaves `tests/test_autostart.py` fully green (28 passed). But that mutation reintroduces exactly the vector this commit removes, because an **empty** `PATH` is not "no PATH" — it is one empty entry, which means the cwd.

Probed rather than argued, with a planted `opencode` in the cwd:

```
PATH-empty-string:  rc=0 stdout='PLANTED-RAN'
PATH-single-colon:  rc=0 stdout='PLANTED-RAN'
PATH-absent:        FileNotFoundError: 'opencode'
```

So on a machine whose enable-time `PATH` happens to contain no absolute entry, the guard is what turns "nothing usable to persist" into "inherit the service manager's default" instead of "`PATH=` → resolve from the daemon's cwd". Omitting the key is the safe outcome and the code gets it right — this is a coverage note, not a defect.

A test would be the same shape as the one you just wrote: `monkeypatch.setenv("PATH", "relative/bin::")` and assert `"PATH" not in env`.

Minor, same block: `if entry and os.path.isabs(entry)` — the `entry and` is redundant, since `os.path.isabs("")` is already `False`. Dropping it keeps the suite green and changes nothing. Fine to leave as documentation of intent.

Useful? React with 👍 / 👎.

hellohanchen (commented):
## Round 3 — `4cb8d54c7a67fd00c898501a2a9254cc765a28a7` — **PASS**

Delta-reviewed from my own round-2 PASS SHA `11631cd2` (confirmed a linear ancestor). One commit, `+13 / -6` across `autostart.py` and `test_autostart.py`.

### The fix

`_service_env` now persists only absolute, non-empty `PATH` entries, and omits the key entirely when none survive. Order and duplicates are preserved; nothing else in the enable path moved.

The scope is right: `_service_env` feeds only `launchagent_plist` (`:159`) and `systemd_unit` (`:280`). `enable_windows` builds its Run-key entry without it, so `os.path.isabs`'s different semantics under `ntpath` never come into play here.

### Mutation matrix

`tests/test_autostart.py`, baseline **28 passed**:

| # | mutation | result |
|---|---|---|
| R1 | drop `os.path.isabs(entry)`, keep `if entry` | 1 failed — killed |
| R2 | drop `entry and`, keep `os.path.isabs` | 28/28 — survived (benign: `isabs("") is False`) |
| R3 | revert to the whole pre-fix `path = os.environ.get("PATH")` form | 1 failed — killed |
| R4 | `if entries:` → assign unconditionally | **28/28 — survived** |
| R5 | drop the `PATH` block entirely | 2 failed — killed |

R3 confirms the red-first claim independently. R1 confirms the new assertion is pinned to the *filtering*, not merely to the presence of a `PATH` key — the dirty fixture `"/usr/bin::relative/bin:/opt/tools/bin:"` asserted against the exact `"/usr/bin:/opt/tools/bin"` is the right shape: it pins the empty entry, the relative entry, and the trailing empty separately.

R4 is filed inline. Short version: an empty `PATH` string is one empty entry, i.e. the cwd — I probed it — so the `if entries:` guard is the security-relevant half of this fix and is the one thing no test holds.

### Verification

- CI 7/7 SUCCESS on this SHA (`pre-commit`, `pytest py3.11`, `pytest py3.12`, CodeQL ×4).
- Full local suite: **3820 tests, 0 failures, 0 errors, 5 skipped**, exit 0 — excluding `tests/test_pi_tool_bridge.py`, which fails all 20 on this host with a Node `ERR_UNKNOWN_FILE_EXTENSION` loading the bridge harness (environmental, green in CI, and excluded identically in rounds 1–2).
- All mutations reverted; `git status --porcelain` empty before filing.

### Verdict

**PASS** — no blocker. One P3 coverage note inline. Round 3 of 3; my P3-1 / P3-2 (HKCU-only registry read, dropped `REG_EXPAND_SZ` type) stand as the agreed follow-up, both failing only in the over-refusing direction.

src/puffo_agent/portal/cli.py
Glimmer0x: Fixed in 11631cd20159b36c7740c9b4801da2c0264aa8ee: the test now records autostart calls and asserts none occur for --no-autostart, avoiding the prior exception-swallowing false positive. Mutation E9, which ignores the flag, fails.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+def autostart_log_path() -> Path:
+    return home_dir() / "logs" / "autostart.log"
+
+
+def _service_env() -> dict[str, str]:
+    # Always explicit: a login-time environment does not inherit the
+    # shell export the user enabled from, and "defaulted at enable time,
+    # different default at boot" is exactly the class of silent
+    # divergence this feature must not introduce.
+    return {"PUFFO_AGENT_HOME": str(home_dir())}
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in 11631cd20159b36c7740c9b4801da2c0264aa8ee by persisting the enable-time PATH, then hardened in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7 to retain only absolute, non-empty entries. The E1 mutation and round-3 filtering mutations confirm the path is pinned.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+    content = systemd_unit(daemon_command(), _service_env())
+    changed = not (path.exists() and path.read_text() == content)
+    if changed:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(content)
+    lines = [f"wrote {path}" if changed else f"unit unchanged at {path}"]
+    for cmd in (
+        ["systemctl", "--user", "daemon-reload"],
+        ["systemctl", "--user", "enable", "--now", SYSTEMD_UNIT_NAME],
+    ):
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in 6554baa and included in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: when the definition changes for an active unit, enable performs daemon-reload before an explicit restart. The ordering and active-service case are covered; mutation E4 fails.

src/puffo_agent/portal/autostart.py
Glimmer0x: Fixed in 11631cd20159b36c7740c9b4801da2c0264aa8ee and included in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: Windows now refuses only when login-time home (persisted HKCU value, otherwise the default) differs from the effective home. Plain setx and default-equal cases pass; mutations E2 and E3 fail.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+    configured: bool
+    active: bool
+    lines: list[str] = field(default_factory=list)
+
+
+Runner = Callable[[list[str]], subprocess.CompletedProcess]
+
+
+def _run(cmd: list[str]) -> subprocess.CompletedProcess:
+    return subprocess.run(cmd, capture_output=True, text=True, check=False)
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in 6554baa and included in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: the service-manager runner maps FileNotFoundError to a clean unavailable result instead of a traceback. Mutation E6 fails.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+        )
+    return ActionResult(True, lines)
+
+
+def disable_linux(run: Runner = _run) -> ActionResult:
+    path = systemd_unit_path()
+    if not path.exists():
+        # Nothing configured; a failing systemctl here (e.g. no user
+        # manager at all) leaves nothing behind to report.
+        return ActionResult(True, ["autostart was not enabled"])
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in 6554baa and included in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: disable/status query the service manager independently of file existence, and a loaded/running unit is stopped even when its file is missing. Mutation E5 fails.

src/puffo_agent/portal/autostart.py
Glimmer0x: The production guard is retained and the final SHA remains 4cb8d54c7a67fd00c898501a2a9254cc765a28a7, which has round-3 PASS. This is recorded as a non-blocking coverage follow-up: when no absolute PATH entries survive, the code correctly omits PATH rather than writing an empty value; no new head was created solely for the surviving test-only mutation.

src/puffo_agent/portal/autostart.py
Glimmer0x: Fixed in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: persisted PATH now keeps only absolute, non-empty entries, removing cwd and relative lookup vectors while preserving order and duplicates. Round-3 mutation R1 and the pre-fix reversion R3 fail.

src/puffo_agent/portal/autostart.py
Glimmer0x: Fixed in 11631cd20159b36c7740c9b4801da2c0264aa8ee: tests now inspect the plist/unit bytes written by enable_macos and enable_linux and assert they carry the effective home and PATH. The previous formatter-only gap is closed.

src/puffo_agent/portal/autostart.py
```diff
@@ -0,0 +1,476 @@
...
+        else "not loaded in launchd (takes effect at next login, or enable again)"
+    )
+    return AutostartStatus(configured, active, lines)
+
+
+# ── Linux (systemd user unit) ────────────────────────────────────────────────
+
+
+def systemd_unit_path() -> Path:
+    return Path.home() / ".config" / "systemd" / "user" / SYSTEMD_UNIT_NAME
```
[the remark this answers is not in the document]
Glimmer0x: Fixed in 6554baa and included in final PASS SHA 4cb8d54c7a67fd00c898501a2a9254cc765a28a7: absolute XDG_CONFIG_HOME is honored for the systemd user unit; relative values are rejected in line with systemd behavior. Mutations E7 and E8 fail.

src/puffo_agent/portal/autostart.py
Glimmer0x: Deferred by owner decision as a non-blocking Windows follow-up. Both HKLM setx /M and unexpanded REG_EXPAND_SZ fail in the safe direction (spurious refusal), and plain per-user setx recovers. A correct change needs HKLM fallback, ExpandEnvironmentStrings, and a multi-hive/multi-type winreg test double; this is recorded rather than misreported as fixed.

Issue #232, opened 2026-08-12
Healthy UNCHANGED refresh outcome is counted as a failure, permanently bricking agents (refresh_broken can never self-clear)


not-so-fat:
## Summary
`CredentialRefresher._propagate_outcome` in `puffo_agent/portal/credential_refresh.py` treats a `RefreshOutcome.UNCHANGED` result (i.e. "token still valid, nothing to refresh") exactly the same as `RefreshOutcome.FAILED` for the `_consecutive_non_success` streak. Only `RefreshOutcome.REFRESHED` resets the counter to 0.

Given `REFRESH_BROKEN_THRESHOLD = 2` (line 54) and a 120s poll interval, if a machine's Claude Code CLI credential legitimately doesn't need an OAuth round-trip for two consecutive poll ticks (~4 minutes) — the normal/healthy state — the daemon flips every registered `cli-local`/`cli-docker` agent into `health=refresh_broken`. Since `UNCHANGED` can never become `REFRESHED` on its own, the streak can never clear itself: every later tick just re-fires `_flip_refresh_broken` with a growing count, permanently.

This also blocks live delivery: `Worker`'s pre-delivery gate (`worker.py` ~1336-1355) calls `_ensure_fresh_token()`, hits the same logic, returns falsy, the batch gets 3 kick-retries then is abandoned (`health=api_error_abandoned`) — even though `claude` works fine when invoked directly on the same machine at the same time.

## Reproduction
1. On a machine where on-disk `~/.claude/.credentials.json`'s `claudeAiOauth.expiresAt` is older than `REFRESH_SAFETY_MARGIN_SECONDS` (600s) but the CLI itself is fully authenticated/functional (session managed separately from that timestamp), every tick legitimately returns `UNCHANGED` (log: `claude credential refresh ok ... but Keychain unchanged — token was still fresh; claude skipped the OAuth round-trip`).
2. Within 2 ticks: `flipping refresh_broken after 2 consecutive unchanged outcome(s)`.
3. Any message sent afterward: `pre-delivery token refresh failed; flagging auth_failed and deferring batch` → 3 kick-retries → `api-error-abandon`.
4. Restarting the daemon only resets the in-memory counter; it re-trips within a few ticks (observed as few as 2-6) since the token still never produces a `REFRESHED` outcome.
5. Confirmed independently: `claude -p "..."` succeeds directly at the same moment the agent shows `refresh_broken`/`auth_failed`.

## Suggested fix
[...]

PR #310, opened 2026-09-02
fix(transport): recover from sleep/wake executor death and report honest health

Glimmer0x:
**Root cause (8/30 incident)**: `--background` runs the daemon as a thread of the windowless Qt tray process, which macOS App Nap suspends wholesale. After resume the loop's default ThreadPoolExecutor is dead; aiohttp rides it for DNS/netrc on every request and `websockets.connect` uses `loop.getaddrinfo` on the same pool, so one dead executor killed all HTTP + WS forever (`cannot schedule new futures after shutdown` ×9136) while runtime.json health said "ok" for 14.5h.

**Fixes**
- `http_client`: pre-send no-op executor probe heals a dead pool (and drops the pre-suspension session) before any bytes go out; a dead-executor error mid-request is healed for the next attempt but re-raised, never replayed (aiohttp redirect hops re-enter the executor after the request is on the wire — double-send reproduced live during review).
- `ws_client` + `worker`: reconnect-failure streaks flow through a transport-state listener wired via the single `Worker._build_wired_client()` seam (both standard and ws-local paths); ≥5 consecutive failures flip runtime.json health `ok → server_unreachable` (~15–40s into an outage), one successful reconnect clears it, stronger signals never clobbered. WS loop heals the same dead-executor error itself (valid subkey ⇒ no prior HTTP call to heal it). `agent list` shows the new state.
- `tray`: best-effort NSProcessInfo `beginActivityWithOptions:` App Nap exemption (still allows system sleep); failure degrades to today's behavior.

**Review**: two rounds by @Peter (exact SHAs a043b4c → f2a0c73 → 56fbaa0), findings verified with live aiohttp-server repros; both incident paths re-run on final SHA with exactly one POST observed. All tests red-first; full suite exit 0; ruff clean.

**Deferred**: structural split (daemon out of the GUI process) → follow-up PR; #232 (refresh_broken can't self-clear) is the same "no recovery path" class, out of scope here.

### DIFF
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -191,6 +191,9 @@ class PuffoCoreMessageClient:
                 bridge_client=bridge_client,
             )
         )
+        # Assignable post-construction (the Worker owns the health policy);
+        # picked up when ``run()`` builds the WS client.
+        self.transport_state_listener: Callable[[bool, int], None] | None = None
         bridge_connected = getattr(self._bridge, "add_connected_callback", None)
         if callable(bridge_connected):
             bridge_connected(self._notify_connected_callbacks)
@@ -283,6 +286,14 @@ class PuffoCoreMessageClient:
         self._ws.on_channel_update = self._handle_channel_update
         # Re-warms caches on every (re)connect, first connect included.
         self._ws.on_connect = self._on_ws_connect
+        # Worker-installed observer (see Worker._transport_state_listener):
+        # feeds reconnect-failure streaks into runtime.json health so a
+        # dead transport stops reporting "ok" (8/30 App Nap incident).
+        # getattr: legacy test seeds construct this client without
+        # __init__, same tolerance as the bridge callback above.
+        self._ws.on_transport_state = getattr(
+            self, "transport_state_listener", None
+        )
         await self.store.open()
         try:
             await self._ws.run()
--- a/src/puffo_agent/agent/puffo_core_client.py
+++ b/src/puffo_agent/agent/puffo_core_client.py
@@ -1448,86 +1448,14 @@ class PuffoCoreMessageClient:
     _DM_NOTICE_INTERVAL_MS = 72 * 3600 * 1000
 
     async def _maybe_send_dm_notice(self, sender_slug: str) -> None:
-        """Operator FYI for every non-trusted sender (contacts included):
-        first DM immediately, then one per 72h; persisted."""
-        if not self.operator_slug:
-            return
-        try:
-            last = await self.store.get_dm_notice(sender_slug)
-        except Exception:
-            last = None
-        now_ms = int(time.time() * 1000)
-        if last is not None and now_ms - last < self._DM_NOTICE_INTERVAL_MS:
-            return
-        display = await self._fetch_display_name(sender_slug)
-        label = f"**{display}** ({sender_slug})" if display else f"@{sender_slug}"
-        try:
-            await self._send_dm(
-                self.operator_slug,
-                f"FYI, {label} is sending direct messages to me.",
-                root_id="",
-            )
-        except Exception as exc:
-            self._log.warning(
-                "dm_notice: failed to notify operator about %s: %s",
-                sender_slug,
-                exc,
-            )
-            return
-        try:
-            await self.store.set_dm_notice(sender_slug, now_ms)
-        except Exception as exc:
-            self._log.warning("dm_notice: failed to persist ts: %s", exc)
+        from . import dm_gate
+
+        await dm_gate.maybe_send_dm_notice(self, sender_slug)
 
     async def _maybe_allowlist_outbound_dm(self, recipient_slug: str) -> None:
-        """Agent DM'd a foreign peer first → allowlist them; best-effort."""
-        if not recipient_slug:
-            return
-        # Never allowlist a sender we're currently gating — the ack DM
-        # echoes back here and would pre-empt the operator's y/n.
-        if any(
-            m.get("sender_slug") == recipient_slug
-            for m in self._pending_dm_approvals.values()
-        ):
-            return
-        # Replying is not consent — only a genuinely agent-initiated
-        # first DM allowlists. A stored inbound DM means they wrote first.
-        try:
-            if await self.store.has_dm_from(recipient_slug):
-                return
-        except Exception:
-            return
-        # Trusted short-circuit before is_allowed can hit the network
-        # (the daemon DMs the operator constantly).
-        if not await self._is_foreign_dm_sender(recipient_slug):
-            return
-        if await self._contacts.is_allowed(recipient_slug):
-            return
-        try:
-            await self.http.post("/allowlists", {"slugs": [recipient_slug]})
-        except Exception as exc:
-            self._log.warning(
-                "dm_gate: outbound allowlist for %s failed: %s",
-                recipient_slug,
-                exc,
-            )
-            return
-        self._contacts.note_allowed(recipient_slug)
-        if not self.operator_slug:
-            return
-        display = await self._fetch_display_name(recipient_slug)
-        label = f"**{display}**(@{recipient_slug})" if display else f"@{recipient_slug}"
-        try:
-            await self._send_dm(
-                self.operator_slug,
-                f"Allowlisted {label} — I messaged them first, so their "
-                "replies won't need approval.",
-                root_id="",
-            )
-        except Exception:
-            self._log.exception(
-                "dm_gate: failed to notify operator of outbound allowlist",
-            )
+        from . import dm_gate
+
+        await dm_gate.maybe_allowlist_outbound_dm(self, recipient_slug)
 
     async def _maybe_gate_foreign_dm(
         self, *, sender_slug: str, text: str
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -60,8 +60,8 @@ def heal_if_dead_executor(exc: BaseException) -> bool:
     """Heal the loop's default executor iff ``exc`` is the dead-executor
     RuntimeError; returns whether it healed.
 
-    Shared by the HTTP retry wrapper and the WS reconnect loop: with a
-    valid subkey ``connect_once`` reaches ``websockets.connect`` (and its
+    Shared by the HTTP wrapper and the WS reconnect loop: with a valid
+    subkey ``connect_once`` reaches ``websockets.connect`` (and its
     ``loop.getaddrinfo``) without a single prior HTTP request, so the WS
     path cannot rely on an HTTP-side heal having happened first.
     """
@@ -71,13 +71,36 @@ def heal_if_dead_executor(exc: BaseException) -> bool:
     return True
 
 
-class _HealedRequest:
-    """One aiohttp request attempt with a single dead-executor heal.
+async def _probe_default_executor() -> bool:
+    """Run a no-op through the loop's default executor; heal if dead.
 
-    The dead-executor RuntimeError fires in the pre-connection stage
-    (threaded DNS / netrc lookup), before any bytes reach the wire, so
-    the one retry cannot double-send a request. Any other error — and a
-    second dead-executor failure after healing — propagates untouched.
+    ``run_in_executor(None, ...)`` is the same scheduling call aiohttp
+    makes for threaded DNS and ``trust_env`` netrc lookups, so a passing
+    probe means this request's own executor hops cannot raise the
+    shutdown error — short of a concurrent shutdown mid-request, which
+    is surfaced to the caller, never retried. Returns whether it healed.
+    """
+    loop = asyncio.get_running_loop()
+    try:
+        await loop.run_in_executor(None, lambda: None)
+        return False
+    except RuntimeError as exc:
+        if not heal_if_dead_executor(exc):
+            raise
+        return True
+
+
+class _HealedRequest:
+    """One aiohttp request attempt, healed *before* any bytes go out.
+
+    A dead default executor is detected by a no-op probe and replaced
+    prior to sending; the pooled connections from before the suspension
+    are dropped with it. A dead-executor RuntimeError that still fires
+    mid-request is healed so the caller's next attempt works, then
+    re-raised — never replayed: aiohttp re-enters the executor on every
+    automatic-redirect hop, so past the first hop the original request
+    is already on the wire and a replay double-sends it (reproduced
+    live: POST → 307, hop-2 executor death, replayed POST).
     """
 
     def __init__(self, client: PuffoCoreHttpClient, method: str, url: str, kwargs: dict):
@@ -87,19 +110,17 @@ class _HealedRequest:
         self._kwargs = kwargs
         self._ctx = None
 
-    async def _enter_once(self):
-        http = await self._client._get_session()
-        self._ctx = http.request(self._method, self._url, **self._kwargs)
-        return await self._ctx.__aenter__()
-
     async def __aenter__(self):
-        try:
-            return await self._enter_once()
-        except RuntimeError as exc:
-            if not heal_if_dead_executor(exc):
-                raise
+        if await _probe_default_executor():
             await self._client.close()
-            return await self._enter_once()
+        try:
+            http = await self._client._get_session()
+            self._ctx = http.request(self._method, self._url, **self._kwargs)
+            return await self._ctx.__aenter__()
+        except RuntimeError as exc:
+            if heal_if_dead_executor(exc):
+                await self._client.close()
+            raise
 
     async def __aexit__(self, *exc_info):
         return await self._ctx.__aexit__(*exc_info)
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -154,9 +154,12 @@ class PuffoCoreHttpClient:
         return _HealedRequest(self, method, url, kwargs)
 
     async def close(self) -> None:
-        if self._session and not self._session.closed:
-            await self._session.close()
-            self._session = None
+        # Detach before awaiting: a concurrent _get_session() during the
+        # await may install a fresh session, which a post-await
+        # ``self._session = None`` would orphan with a live connector.
+        session, self._session = self._session, None
+        if session is not None and not session.closed:
+            await session.close()
 
     def _load_signing_key(self) -> tuple[Ed25519KeyPair, str]:
         sess = self.keystore.load_session(self.slug)
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -52,8 +52,23 @@ def _heal_dead_default_executor() -> None:
     )
     logger.warning(
         "default executor was shut down while the loop kept running; "
-        "installed a fresh one and rebuilding the HTTP session"
+        "installed a fresh one"
     )
+
+
+def heal_if_dead_executor(exc: BaseException) -> bool:
+    """Heal the loop's default executor iff ``exc`` is the dead-executor
+    RuntimeError; returns whether it healed.
+
+    Shared by the HTTP retry wrapper and the WS reconnect loop: with a
+    valid subkey ``connect_once`` reaches ``websockets.connect`` (and its
+    ``loop.getaddrinfo``) without a single prior HTTP request, so the WS
+    path cannot rely on an HTTP-side heal having happened first.
+    """
+    if _DEAD_EXECUTOR_MESSAGE not in str(exc):
+        return False
+    _heal_dead_default_executor()
+    return True
 
 
 class _HealedRequest:
@@ -81,9 +96,8 @@ class _HealedRequest:
         try:
             return await self._enter_once()
         except RuntimeError as exc:
-            if _DEAD_EXECUTOR_MESSAGE not in str(exc):
+            if not heal_if_dead_executor(exc):
                 raise
-            _heal_dead_default_executor()
             await self._client.close()
             return await self._enter_once()
 
--- a/src/puffo_agent/crypto/http_client.py
+++ b/src/puffo_agent/crypto/http_client.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import asyncio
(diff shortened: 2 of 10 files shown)


src/puffo_agent/agent/puffo_core_client.py
```diff
@@ -191,6 +191,9 @@ def __init__(
                 bridge_client=bridge_client,
             )
         )
+        # Assignable post-construction (the Worker owns the health policy);
+        # picked up when ``run()`` builds the WS client.
+        self.transport_state_listener: Callable[[bool, int], None] | None = None
```
hellohanchen, line 196: **B1 — blocker: `pre-commit` is red on this SHA.**

```
python structural limits.................................................Failed
src/puffo_agent/agent/puffo_core_client.py:1: file is 2009 lines (limit 2000)
Python structural check failed: 1 violation(s).
```

Attribution so the effort goes to the right half: on the merge-base `f0691564` this file is **1998** lines; #310 adds **+11 net** and lands at **2009**. This PR crosses the line but did not create the crowding — main was already 2 lines under the cap, so the next PR to touch this file trips it regardless.

The +11 is this 3-line attribute plus an 8-line assignment at line ~289, both load-bearing. Shaving them is not the fix. This is a policy call — split the file, or move `FILE_LINE_LIMIT` in `tools/check_python_structure.py:9` — and it is yours to make, not mine.

src/puffo_agent/crypto/http_client.py
```diff
@@ -24,6 +26,106 @@ def __init__(self, status: int, body: str):
...
+    def __init__(self, client: PuffoCoreHttpClient, method: str, url: str, kwargs: dict):
+        self._client = client
+        self._method = method
+        self._url = url
+        self._kwargs = kwargs
+        self._ctx = None
+
+    async def __aenter__(self):
+        if await _probe_default_executor():
+            await self._client.close()
```
hellohanchen, line 115: **F2 — low: this newly-added `close()` on a live request path can orphan an `aiohttp` session.**

`close()` (line 156-159, unchanged by this PR) clears the field *after* the await:

```python
async def close(self) -> None:
    if self._session and not self._session.closed:
        await self._session.close()
        self._session = None      # <-- after the await
```

`aiohttp.ClientSession.close()` flips `closed` to True at the start of its connector teardown, before the coroutine returns. So while task A is parked inside that await, a concurrent task B calling `_get_session()` (line 148) sees `closed == True`, builds a fresh session, assigns it to `self._session` — and then A resumes and overwrites it with `None`. B's session is live and in use but no longer referenced: never closed (leaked connector/sockets, plus the `Unclosed client session` warning), and the next caller builds another.

The race is **newly reachable because of this line**. Before #310, `close()` was only called from teardown (`desired_install.py:330`, `cli.py:857`) — single-shot, no concurrency. This call puts it on a live request path, and the window opens on exactly the scenario the PR targets: App Nap resume, when several parked requests wake together. Concurrency is real there — `puffo_core_client.py:883` gathers two request paths. The same applies to the `close()` in the `except` clause at line 122.

Probe-confirmed, driving the real `close()` / `_get_session()` with a session double that flips `closed` where aiohttp does:

```
B got: <S s1> | c._session: <S s1>
after close() returns -> c._session: None
sessions created: [<S s0>, <S s1>]
ORPHANED LIVE SESSION: True
```

Low, not a blocker: it leaks fds per recovery event, it does not corrupt or hang anything.

src/puffo_agent/crypto/http_client.py
```diff
@@ -24,6 +26,106 @@ def __init__(self, status: int, body: str):
...
+def heal_if_dead_executor(exc: BaseException) -> bool:
+    """Heal the loop's default executor iff ``exc`` is the dead-executor
+    RuntimeError; returns whether it healed.
+
+    Shared by the HTTP wrapper and the WS reconnect loop: with a valid
+    subkey ``connect_once`` reaches ``websockets.connect`` (and its
+    ``loop.getaddrinfo``) without a single prior HTTP request, so the WS
+    path cannot rely on an HTTP-side heal having happened first.
+    """
+    if _DEAD_EXECUTOR_MESSAGE not in str(exc):
```
hellohanchen, line 68: **F3 — nit: this guard is unpinned by the tests.**

Mutating it to `if False:` — so `heal_if_dead_executor` heals **every** `RuntimeError` — survives all 4 tests in `test_transport_recovery.py`. The guard as written is correct; this is a coverage gap, not a defect.

Worth closing because the guard is not cosmetic everywhere. In the WS loop the two branches differ only in a log label, but in `_probe_default_executor` it is the whole decision:

```python
except RuntimeError as exc:
    if not heal_if_dead_executor(exc):
        raise
    return True
```

Lose it and an unrelated `RuntimeError` from `run_in_executor` — `Event loop is closed` being the obvious one — is swallowed, a healthy default executor is replaced, and the client's session is closed, instead of the error propagating.

One test asserting that a non-matching `RuntimeError` propagates and leaves `loop._default_executor` untouched would pin the claim this docstring already makes about the intentional-teardown variants.

src/puffo_agent/portal/worker.py
```diff
@@ -111,6 +111,14 @@ def _rebuild_managed_system_prompt(
 
 RECONNECT_BACKOFF_SECONDS = 5.0
 
+# Consecutive WS reconnect failures before runtime.json health flips to
+# "server_unreachable". With the WS client's 1s→30s doubling backoff the
+# fifth failure lands ~15s into an outage (~40s when each attempt burns a
+# connect timeout): a blip surviving one or two reconnects never flips,
+# while an operator checking `agent list` sees the truth within the
+# minute instead of "ok" for hours. One successful reconnect clears it.
+_WS_DEGRADE_THRESHOLD = 5
```
hellohanchen, line 120: **F4 — nit: this constant's value is unpinned.**

Mutating `5 → 1` survives the suite. That is structural rather than an oversight: `test_ws_streaks_flip_health_to_server_unreachable_and_back` imports the constant and loops `for streak in range(1, _WS_DEGRADE_THRESHOLD)`, so the test moves with the value it is meant to pin — at `1` the pre-threshold loop is simply empty.

The comparison direction *is* covered (mutating `or rt.health != "ok"` away, and killing the recovery branch, both go red). Only the literal is loose. Two hard-coded assertions — streak 4 leaves health `"ok"`, streak 5 flips it — would pin the policy number this comment argues for.

hellohanchen (commented):
## Review — sleep/wake executor recovery @ `56fbaa0bab0431318c61f7e26ac9ad23e923cc2e`

Verdict: **BLOCK** — one required check is red (B1). No correctness blocker found in the product logic; the remaining three items are low/nit.

Base for the diff: `f0691564161193fd40142754c1d63d6678ed3dc6` (merge-base), +521/−24.

### What I verified as correct

- **The heal is loop-global and reaches WS.** `_heal_dead_default_executor` calls `set_default_executor` on the running loop, so one heal on the HTTP path also revives `websockets.connect`'s `loop.getaddrinfo`. The separate `except RuntimeError` clause in the WS reconnect loop (`ws_client.py:417`) is correctly ordered *after* `OSError` and *before* the catch-all `except Exception`, so it does not shadow either.
- **The "never replay" decision is right and is tested.** `_HealedRequest.__aenter__` heals a mid-request death and re-raises rather than retrying. That is the correct call: aiohttp re-enters the executor on each automatic-redirect hop, so past hop 1 the request is already on the wire. `test_mid_request_executor_death_heals_but_never_replays` pins it.
- **The message guard matches the real CPython string.** I probed the interpreter rather than trusting the docstring:
  ```
  python 3.14.4
  shutdown msg   : 'cannot schedule new futures after shutdown'
  matches guard  : True
  guard matches interpreter-shutdown variant: False
  ```
  So the docstring's claim — that the intentional-teardown variants are deliberately *not* healed — holds. CI runs 3.11/3.12.
- **The listener wiring is ordered correctly.** I initially suspected `self._ws.on_transport_state = getattr(self, "transport_state_listener", None)` ran during construction, before the Worker could assign the listener. It does not: that line is inside `async def listen()` (line 244), not `__init__` (line 150), so `worker.py:401` has already assigned it. **No bug — withdrawn.**
- **`_build_wired_client`'s coverage claim is true.** After this PR there is exactly one remaining `_build_puffo_core_client(` call site in `src/`, and it is inside `_build_wired_client` itself. Both runtime paths (`worker.py:1149`, `worker_run.py:447`) go through the seam, and the test asserts through it rather than around it.
- **The App Nap option mask is the right constant.** `0x00FFFFFF & ~(1 << 20)` = `0x00EFFFFF` = `NSActivityUserInitiatedAllowingIdleSystemSleep` (`NSActivityUserInitiated` is `0x00FFFFFF | NSActivityIdleSystemSleepDisabled`, and `NSActivityIdleSystemSleepDisabled` is `1 << 20`, which is inside that range). System sleep stays allowed, as the comment says. Every `objc_msgSend` call re-declares `argtypes` before use, which is required on arm64.
- **The recovery notify is exception-safe and correctly gated.** `_notify_transport_state` swallows callback exceptions, and `connect_once` fires the healthy notify only `if self._reconnect_failures:` — so a first connect does not emit a spurious recovery.

### Test evidence

Full suite at this SHA on py3.12, `tests/test_pi_tool_bridge.py` deselected: **3793 tests, 0 failures, 0 errors, 4 skipped** (JUnit XML counts, not the exit code). That one file fails **identically on the merge-base** `f0691564` — same 4 test names — so it is environmental on my host, not this PR. `ruff check src tests`: clean.

Mutation matrix on the new tests (baseline 4 passed):

| # | mutation | site | outcome |
|---|---|---|---|
| M1 | `if _DEAD_EXECUTOR_MESSAGE not in str(exc)` → `if False` (heal every `RuntimeError`) | `http_client.py:68` | **SURVIVED** |
| M2 | drop `await self._client.close()` after a successful pre-send heal | `http_client.py:115` | killed |
| M3 | `_WS_DEGRADE_THRESHOLD = 5` → `1` | `worker.py:120` | **SURVIVED** |
| M4 | drop `or rt.health != "ok"` (let the transport clobber stronger signals) | `worker.py:428` | killed |
| M5 | `if rt.health == "server_unreachable":` → `if False` (never recover) | `worker.py:418` | killed |

M1 and M3 are the two findings below (F3, F4). Note M3's survival is structural, not a coverage accident: the test loops `for streak in range(1, _WS_DEGRADE_THRESHOLD)` over the imported constant, so it moves with the value it is meant to pin. The *comparison direction* is pinned; only the literal `5` is not.

---

### B1 — blocker: `pre-commit` is red on this SHA

`pre-commit` fails, from the job log:

```
python structural limits.................................................Failed
src/puffo_agent/agent/puffo_core_client.py:1: file is 2009 lines (limit 2000)
Python structural check failed: 1 violation(s).
```

Attribution, so nobody spends time on the wrong half: on `f0691564` the file is **1998** lines; this PR adds **+11 net** and lands at **2009**. #310 crosses the line, but it did not create the crowding — main was already 2 lines under a 2000-line cap, so the next PR to touch this file trips it regardless. The +11 here is a 3-line attribute with comment and an 8-line assignment with comment, both load-bearing for the feature.

Worth deciding as a policy question rather than shaving lines: either this file gets split, or the cap moves. Not my call to make — flagging it so it is a deliberate choice.

### F2 — low: a healed request can orphan a live `aiohttp` session

`close()` clears `self._session` **after** the await:

```python
async def close(self) -> None:
    if self._session and not self._session.closed:
        await self._session.close()
        self._session = None      # <-- after the await
```

`aiohttp.ClientSession.close()` flips `closed` to True at the *start* of its connector teardown, before that coroutine returns. So while task A is parked inside `await self._session.close()`, a concurrent task B calling `_get_session()` sees `closed == True`, builds a fresh session, and assigns it to `self._session` — and then A resumes and overwrites it with `None`. B's session is live, in use, and no longer referenced by the client: it is never closed (leaked connector/sockets plus the `Unclosed client session` warning), and the next caller builds yet another.

This is **newly reachable because of this PR**. Before #310, `close()` was only called from teardown paths (`desired_install.py:330`, `cli.py:857`) — single-shot, no concurrency. #310 moves it onto a live request path (`_HealedRequest.__aenter__`), and the window opens on exactly the scenario the PR targets: App Nap resume, when several parked requests wake at once. Concurrency is real on that path — `puffo_core_client.py:883` does `asyncio.gather(self._warm_member_caches(), self._contacts.refresh())`.

Probe-confirmed, driving the real `close()` / `_get_session()` with a session double that flips `closed` where aiohttp does:

```
B got: <S s1> | c._session: <S s1>
after close() returns -> c._session: None
sessions created: [<S s0>, <S s1>]
ORPHANED LIVE SESSION: True
```

Low, not a blocker: it leaks fds per recovery event rather than corrupting or hanging anything.

### F3 — nit: the dead-executor message guard is unpinned

M1 above: replacing the guard so `heal_if_dead_executor` heals **every** `RuntimeError` survives all 4 tests. The guard as written is correct — this is a coverage gap, not a defect.

It is worth closing because the guard is not cosmetic everywhere. In the WS loop the two branches only differ in a log label, but in `_probe_default_executor` the guard is the whole decision:

```python
except RuntimeError as exc:
    if not heal_if_dead_executor(exc):
        raise
    return True
```

Lose it and an unrelated `RuntimeError` from `run_in_executor` — `Event loop is closed` is the obvious one — gets swallowed, a healthy default executor is replaced, and the client's session is closed, instead of the error propagating. A test asserting that a non-matching `RuntimeError` propagates and leaves `loop._default_executor` untouched pins the docstring's own claim about the intentional-teardown variants.

### F4 — nit: `_WS_DEGRADE_THRESHOLD`'s value is unpinned

M3 above. `test_ws_streaks_flip_health_to_server_unreachable_and_back` imports the constant and loops `range(1, _WS_DEGRADE_THRESHOLD)`, so mutating `5 → 1` makes the pre-threshold loop empty and everything still passes. A single literal assertion — streak 4 leaves health `"ok"`, streak 5 flips it — pins the policy number the comment argues for. Cosmetic; the comparison itself is already covered by M4/M5.

---

All experiments were run in my own worktree and reverted; `git status --porcelain` is empty at `56fbaa0ba`.

src/puffo_agent/crypto/http_client.py
```diff
@@ -48,10 +150,16 @@ async def _get_session(self) -> aiohttp.ClientSession:
...
 
+    def _healed_request(self, method: str, url: str, **kwargs):
+        return _HealedRequest(self, method, url, kwargs)
+
     async def close(self) -> None:
-        if self._session and not self._session.closed:
-            await self._session.close()
-            self._session = None
+        # Detach before awaiting: a concurrent _get_session() during the
+        # await may install a fresh session, which a post-await
```
hellohanchen, line 158: **P3 (non-blocking) — the `.closed` half of `_get_session` is unpinned.**

The new `close()` is correct. This is about the guard it leans on next door.

`_get_session` (`http_client.py:149`) reads `if self._session is None or self._session.closed:`. Deleting the `or self._session.closed` clause leaves the whole suite green, including the new race test:

```
D3  _get_session: drop `or self._session.closed`   53 tests, 0 failures  -> SURVIVED
```

The reason the new test does not reach it is worth stating, because it is not an oversight in the test so much as a consequence of the fix working: `close()` detaches synchronously, before its first await. By the time `_SlowCloseSession.close()` has set `close_started` and the test calls `_get_session()`, `self._session` is already `None` — so the test exercises the `is None` branch, and only that one.

**What I checked before deciding how to rate this.** My first thought was that the `.closed` clause is load-bearing for F2 — that it covers a caller arriving just *before* the detach. It does not. A caller that reaches `_get_session()` before the detach reads `self._session` while it is still the old session and `closed` is still `False`, so it gets handed the old session either way; the clause cannot help there, and that window is inherent rather than introduced here. After the detach the field is `None` and the `is None` branch handles it. So the clause is defensive cover for a session that was closed by something other than `close()`, which is pre-existing behaviour and not part of this fix.

Hence P3, and hence not a reason to hold the merge. It is only worth a line because the surviving mutant sits directly beside the code under review, and someone reading the mutation results later should not have to re-derive that it is benign.

hellohanchen (commented):
## Round 2 — `3933ec455a81bece45fb34c3a68840cb7c1e2548`

**PASS.** No blocker. Every round-1 finding is closed, and the two that mattered are closed by a test that actually pins them rather than by code that merely looks right.

Delta-reviewed against my own round-1 SHA `56fbaa0bab0431318c61f7e26ac9ad23e923cc2e` (confirmed a linear ancestor; 2 commits, 157 insertions / 81 deletions across 4 files), not the PR base — otherwise the round-1 material gets re-read and anything that exists only in the delta gets missed. CI rollup green 7/7.

### The file move — proven identical, not read as identical

The line-count gate was a policy call and you made it: split the file rather than raise the ceiling. `puffo_core_client.py` is **1937** lines, as claimed.

A "pure move" is exactly where a single dropped token survives a fully green suite, so I did not eyeball it. I extracted both method bodies from `56fbaa0`, extracted the two new `dm_gate` functions, normalised `self.` → `client.`, and compared:

```
async def maybe_send_dm_notice          IDENTICAL
async def maybe_allowlist_outbound_dm   IDENTICAL
```

Byte-identical modulo the receiver rename. Zero behaviour change is a fact here, not a claim.

Then the other half of a move — whether the delegation is exercised, or whether the tests only ever reach the extracted helper directly:

```
D4  make _maybe_send_dm_notice a bare `return`        FAILED  5 failures
D5  make _maybe_allowlist_outbound_dm a bare `return` FAILED  1 failure
D6  drop the has_dm_from early return (in dm_gate)    FAILED  2 failures
D7  drop the _pending_dm_approvals guard (in dm_gate) FAILED  1 failure
```

Wiring and moved branches both still pinned. The move cost no coverage.

### F2 — the mid-close replacement session

`session, self._session = self._session, None` before the await is the right shape, and your reading of why it became reachable is right: moving `close()` onto the live request path (`_HealedRequest.__aenter__:115`, and again at `:122` in the except) is what made a concurrent `_get_session()` during the teardown await possible at all.

The regression test is a real race test, not a shape test — `_SlowCloseSession` marks itself closed at the top of `close()` (which is what aiohttp does) and then blocks on an event, so the replacement really is installed while the await is in flight. And it goes red on exactly the old code:

```
D1  revert close() to the post-await `self._session = None`   FAILED  1 failure
```

That is the counterfactual that matters. The test fails for the original defect, not for something adjacent.

### F3 / F4 — heal boundary and the threshold literal

`test_heal_only_fires_on_the_designated_dead_executor_message` covers all three shapes: the explicit-shutdown message, the interpreter-shutdown message, and a generic `RuntimeError`. That last pair is what makes the boundary real — `"cannot schedule new futures after interpreter shutdown"` does *not* contain `"cannot schedule new futures after shutdown"`, so the negative is load-bearing rather than decorative.

This closes round 1's M1, which survived then:

```
D2  widen _DEAD_EXECUTOR_MESSAGE to "cannot schedule new futures"   FAILED  1 failure
```

`assert _WS_DEGRADE_THRESHOLD == 5` with the reasoning next to it is the right way to hold a policy constant. Nothing to add.

### One P3

`_get_session`'s `or self._session.closed` (`http_client.py:149`) is unpinned — removing it leaves the suite green. Detail is in the inline comment, including why I do **not** think it undermines the F2 fix; I checked that before writing it down.

### Verdict

**PASS.** The P3 is a pre-existing defensive guard, not something this PR introduced, and I would not hold the merge for it.

Suites: focused set (`test_transport_recovery.py` + `test_auto_accept_dm.py`) → 53 tests, 0 failures, 0 errors. Full suite → **3795 tests, 0 failures, 0 errors** (counts read from `--junitxml`, not the exit code). `tests/test_pi_tool_bridge.py` is excluded: all 20 of its tests fail on my box with Node `ERR_UNKNOWN_FILE_EXTENSION` loading the bridge harness — environmental here, unrelated to this diff, and green in CI. Worktree confirmed clean after every mutation; all mutations reverted.

src/puffo_agent/agent/puffo_core_client.py
Glimmer0x: Fixed in final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548: the relevant method bodies were moved to dm_gate.py, reducing puffo_core_client.py to 1937 lines without changing behavior. Pre-commit and the structural limit check are green.

src/puffo_agent/crypto/http_client.py
Glimmer0x: Fixed in final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548: close now detaches self._session before awaiting session.close, so a concurrently created replacement cannot be overwritten. The deterministic regression fails against the old ordering and passes with the fix; round 2 independently confirmed the counterfactual.

src/puffo_agent/crypto/http_client.py
Glimmer0x: Fixed in final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548: tests now assert unrelated RuntimeError messages propagate and leave the executor unchanged, while the intended dead-executor variants still heal. The round-1 surviving relaxed-message mutation now fails.

src/puffo_agent/agent/puffo_core_client.py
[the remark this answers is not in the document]
Glimmer0x: Fixed in final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548: the two DM-gate method bodies moved to dm_gate.py using the existing delegation pattern. puffo_core_client.py is now 1937 lines, structural checks pass, and round 2 verified the move byte-for-byte.

src/puffo_agent/portal/worker.py
Glimmer0x: Fixed in final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548: tests pin the policy value at 5 and assert the pre-threshold and threshold transitions explicitly, rather than deriving both sides from the constant.

src/puffo_agent/crypto/http_client.py
Glimmer0x: The closed-session fallback is intentionally retained without changing final PASS SHA 3933ec455a81bece45fb34c3a68840cb7c1e2548. Round 2 established that it is pre-existing defensive behavior, not load-bearing for the fixed close race; this non-blocking coverage note is recorded for the next relevant head change.

PR #313, opened 2026-09-03
fix: report remote agent startup readiness

Glimmer0x:
## Summary
- fail remote `cli-docker` create before identity materialization when Docker is missing or the daemon is unavailable, reusing the existing Docker binary resolver and container-state probe
- persist `starting` until the worker completes runtime warm-up and its post-warm health gate
- keep the machine command open until durable runtime state reaches `running`, `error`, `stopped`, or the 10-minute create deadline
- make daemon startup serialization observe and log a failed startup result instead of ignoring `wait_warm()`

## Contract
The create ack remains `{ok: true, agent_slug}` on success. Startup failures now return `ok: false` with `agent_start_failed` or `agent_start_timeout`. Docker preflight failures use the existing `runtime_not_ready` family with `reason=not_installed|daemon_unavailable`.

## Verification
- `env -u CODEX_HOME uv run pytest -q` (full suite passed; 2 expected skips)
- focused runtime/control group passed
- `uv run pre-commit run --all-files`
- `git diff --check`
- mutation checks proved the regression tests fail when Docker preflight, create-result propagation, running transition, or daemon startup observation is removed

The ambient Puffo agent process sets `CODEX_HOME`; the first full-suite run exposed 7 unrelated host-MCP isolation tests. Re-running those tests and the full suite with that ambient variable removed passed.

### DIFF
--- a/src/puffo_agent/portal/control/client.py
+++ b/src/puffo_agent/portal/control/client.py
@@ -44,9 +44,11 @@ HTTP_TIMEOUT = aiohttp.ClientTimeout(total=30)
 HEARTBEAT_INTERVAL_SECONDS = 30.0
 # Remote create waits for the daemon's durable worker lifecycle instead of
 # acknowledging as soon as agent.yml exists. Docker's first image build may
-# legitimately take several minutes.
+# legitimately take several minutes. The server's Portal machine-command TTL
+# is 24 hours; this shorter bound is a UI/operator feedback budget, not the
+# command-expiry budget used by Runtime permission/cancel commands.
 AGENT_START_TIMEOUT_SECONDS = 10 * 60.0
-AGENT_START_POLL_SECONDS = 0.1
+AGENT_START_POLL_SECONDS = 0.5
 
 
 class _DuplicateDelivery(ControlError):
@@ -229,7 +231,9 @@ async def _wait_for_agent_start(agent_id: str) -> dict | None:
     loop = asyncio.get_running_loop()
     deadline = loop.time() + AGENT_START_TIMEOUT_SECONDS
     while loop.time() < deadline:
-        runtime = RuntimeState.load(agent_id)
+        # RuntimeState.load performs filesystem I/O. Keep repeated startup
+        # observation off the single control event loop.
+        runtime = await asyncio.to_thread(RuntimeState.load, agent_id)
         if runtime is not None:
             if runtime.status == "running":
                 return None
@@ -247,8 +251,9 @@ async def _wait_for_agent_start(agent_id: str) -> dict | None:
     return {
         "error_code": "agent_start_timeout",
         "error": (
-            f"agent worker did not finish starting within "
-            f"{AGENT_START_TIMEOUT_SECONDS:g}s"
+            f"agent worker has not reached running within "
+            f"{AGENT_START_TIMEOUT_SECONDS:g}s; it remains configured and "
+            "may still finish starting, so check its status before retrying"
         ),
     }
 
@@ -624,6 +629,7 @@ class MachineControlClient:
     def __init__(self, machine) -> None:
         self.machine = machine
         self._seen_nonces: dict[str, int] = {}  # nonce -> ts; pruned to the ts window
+        self._command_tasks: set[asyncio.Future] = set()
         # Serialize WS writes — acks (receive loop) + heartbeat/capabilities
         # (sender task) share one socket; concurrent send_json would interleave.
         self._send_lock = asyncio.Lock()
@@ -686,7 +692,7 @@ class MachineControlClient:
                                     "control: WS connected; agent.status sender ready"
                                 )
                         elif frame.get("type") == "command":
-                            await self._handle(ws, frame)
+                            await self._handle(ws, frame, background_create=True)
                         elif frame.get("type") == "error":
                             log.warning("control: server rejected ws: %s", frame.get("reason"))
                             break
@@ -716,10 +722,15 @@ class MachineControlClient:
                 await self._send(ws, {"type": "capabilities", "capabilities": caps})
                 last_caps = caps
 
-    async def _handle(self, ws: aiohttp.ClientWebSocketResponse, frame: dict) -> None:
+    async def _handle(
+        self,
+        ws: aiohttp.ClientWebSocketResponse,
+        frame: dict,
+        *,
+        background_create: bool = False,
+    ) -> None:
         command_id = frame.get("command_id")
         operator_slug = frame.get("operator_slug")
-        result: dict = {"ok": False, "error_code": "command_failed"}
         try:
             pairing = load_pairings().get(operator_slug)
             if pairing is None:
@@ -739,19 +750,6 @@ class MachineControlClient:
                     n: t for n, t in self._seen_nonces.items() if t > cutoff
                 }
                 self._seen_nonces[nonce] = int(envelope.get("ts", now_ms()))
-            result = await execute_command(
-                decrypted["op"],
-                decrypted["agent_slug"],
-                decrypted["params"],
-                server_url=pairing.server_url,
-                paired_root_pubkey=pairing.operator_root_pubkey,
-                command_id=str(command_id or ""),
-            )
-            if isinstance(result, dict) and not result.get("ok", True):
-                log.warning(
-                    "control: command %s op=%s failed: %s",
-                    command_id, decrypted["op"], result.get("error"),
-                )
         except _DuplicateDelivery:
             log.debug(
                 "control: duplicate delivery suppressed for command %s",
@@ -765,14 +763,73 @@ class MachineControlClient:
         except Exception as exc:  # noqa: BLE001
             log.warning("control: command %s failed: %s", command_id, exc)
             result = {"ok": False, "error_code": "command_failed"}
-        if command_id:
-            try:
-                await self._send(
-                    ws,
-                    {"type": "ack", "command_id": command_id, "result": result},
+        else:
+            execution = self._execute_and_ack(
+                ws,
+                command_id,
+                decrypted,
+                pairing,
+            )
+            if background_create and command_id and decrypted["op"] == "create":
+                # A first Docker image build can take minutes. Keep that wait
+                # out of the one machine-control receive loop so pause/resume,
+                # runtime decisions, and other operators remain responsive.
+                task = spawn(execution, name=f"control.create:{command_id}")
+                self._command_tasks.add(task)
+                task.add_done_callback(self._command_tasks.discard)
+                return
+            await execution
+            return
+
+        await self._send_ack(ws, command_id, result)
+
+    async def _execute_and_ack(
+        self,
+        ws: aiohttp.ClientWebSocketResponse,
+        command_id: str | None,
+        decrypted: dict,
+        pairing,
+    ) -> None:
+        result: dict = {"ok": False, "error_code": "command_failed"}
+        try:
+            result = await execute_command(
+                decrypted["op"],
+                decrypted["agent_slug"],
+                decrypted["params"],
+                server_url=pairing.server_url,
+                paired_root_pubkey=pairing.operator_root_pubkey,
+                command_id=str(command_id or ""),
+            )
+            if isinstance(result, dict) and not result.get("ok", True):
+                log.warning(
+                    "control: command %s op=%s failed: %s",
+                    command_id,
+                    decrypted["op"],
+                    result.get("error"),
                 )
-            except Exception as exc:  # noqa: BLE001
-                log.debug("control: ack %s failed: %s", command_id, exc)
+        except ControlError as exc:
+            log.warning("control: rejected command %s: %s", command_id, exc)
+            result = {"ok": False, "error_code": "command_rejected"}
+        except Exception as exc:  # noqa: BLE001
+            log.warning("control: command %s failed: %s", command_id, exc)
+            result = {"ok": False, "error_code": "command_failed"}
+        await self._send_ack(ws, command_id, result)
+
+    async def _send_ack(
+        self,
+        ws: aiohttp.ClientWebSocketResponse,
+        command_id: str | None,
+        result: dict,
+    ) -> None:
+        if not command_id:
+            return
+        try:
+            await self._send(
+                ws,
+                {"type": "ack", "command_id": command_id, "result": result},
+            )
+        except Exception as exc:  # noqa: BLE001
+            log.debug("control: ack %s failed: %s", command_id, exc)
 
 
 class ControlManager:
--- a/src/puffo_agent/portal/control/client.py
+++ b/src/puffo_agent/portal/control/client.py
@@ -42,6 +42,11 @@ HTTP_TIMEOUT = aiohttp.ClientTimeout(total=30)
 # Control-WS heartbeat: liveness ping + capability re-check cadence. Must stay
 # well under the server's HEARTBEAT_TIMEOUT (90s) or the server culls us.
 HEARTBEAT_INTERVAL_SECONDS = 30.0
+# Remote create waits for the daemon's durable worker lifecycle instead of
+# acknowledging as soon as agent.yml exists. Docker's first image build may
+# legitimately take several minutes.
+AGENT_START_TIMEOUT_SECONDS = 10 * 60.0
+AGENT_START_POLL_SECONDS = 0.1
 
 
 class _DuplicateDelivery(ControlError):
@@ -193,8 +198,8 @@ async def _create_agent_command(
     async def _materialize(_ctx: dict) -> None:
         await _materialize_slug_binding(server_url, pending_token, slug_binding)
 
-    def _preflight(context: dict) -> None:
-        preflight_runtime(context["runtime"], agent_id=context["agent_id"])
+    async def _preflight(context: dict) -> None:
+        await preflight_runtime(context["runtime"], agent_id=context["agent_id"])
 
     try:
         result = await provision_agent_from_bundle(
@@ -207,7 +212,45 @@ async def _create_agent_command(
         return {"ok": False, "error": exc.reason, **exc.fields}
     except ControlError as exc:
         return {"ok": False, "error": str(exc)}
+    startup = await _wait_for_agent_start(result["agent_id"])
+    if startup is not None:
+        return {
+            "ok": False,
+            "agent_slug": result["agent_id"],
+            **startup,
+        }
     return {"ok": True, "agent_slug": result["agent_id"]}
+
+
+async def _wait_for_agent_start(agent_id: str) -> dict | None:
+    """Return ``None`` once the worker is running, else a structured failure."""
+    from ..state import RuntimeState
+
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + AGENT_START_TIMEOUT_SECONDS
+    while loop.time() < deadline:
+        runtime = RuntimeState.load(agent_id)
+        if runtime is not None:
+            if runtime.status == "running":
+                return None
+            if runtime.status == "error":
+                return {
+                    "error_code": "agent_start_failed",
+                    "error": runtime.error or "agent worker failed to start",
+                }
+            if runtime.status == "stopped":
+                return {
+                    "error_code": "agent_start_failed",
+                    "error": "agent worker stopped before startup completed",
+                }
+        await asyncio.sleep(AGENT_START_POLL_SECONDS)
+    return {
+        "error_code": "agent_start_timeout",
+        "error": (
+            f"agent worker did not finish starting within "
+            f"{AGENT_START_TIMEOUT_SECONDS:g}s"
+        ),
+    }
 
 
 async def post_usage_snapshot(machine, base: str) -> bool:
--- a/src/puffo_agent/portal/control/client.py
+++ b/src/puffo_agent/portal/control/client.py
@@ -630,9 +630,15 @@ class MachineControlClient:
         self.machine = machine
         self._seen_nonces: dict[str, int] = {}  # nonce -> ts; pruned to the ts window
         self._command_tasks: set[asyncio.Future] = set()
+        self._inflight_command_ids: set[str] = set()
+        self._nonce_commands: dict[str, str] = {}
+        self._completed_results: dict[str, dict] = {}
+        self._pending_acks: dict[str, dict] = {}
+        self._active_ws: aiohttp.ClientWebSocketResponse | None = None
         # Serialize WS writes — acks (receive loop) + heartbeat/capabilities
         # (sender task) share one socket; concurrent send_json would interleave.
         self._send_lock = asyncio.Lock()
+        self._ack_flush_lock = asyncio.Lock()
 
     async def run(self, stop: asyncio.Event) -> None:
         while not stop.is_set():
@@ -686,6 +692,8 @@ class MachineControlClient:
                             continue
                         if frame.get("type") == "connected":
                             if not reporter_ready:
+                                self._active_ws = ws
+                                await self._flush_pending_acks(ws)
                                 reporter.set_sender(_report)
                                 reporter_ready = True
                                 log.info(
@@ -697,6 +705,8 @@ class MachineControlClient:
                             log.warning("control: server rejected ws: %s", frame.get("reason"))
                             break
                 finally:
+                    if self._active_ws is ws:
+                        self._active_ws = None
                     if reporter_ready:
                         reporter.set_sender(None)
                     sender.cancel()
@@ -746,13 +756,43 @@ class MachineControlClient:
                 # Bound the replay set: a nonce older than the ts window is
                 # already rejected by decrypt_command, so it's safe to forget.
                 cutoff = now_ms() - TS_WINDOW_MS
-                self._seen_nonces = {
-                    n: t for n, t in self._seen_nonces.items() if t > cutoff
+                expired_command_ids = {
+                    command
+                    for n, t in self._seen_nonces.items()
+                    if t <= cutoff
+                    if (command := self._nonce_commands.get(n)) is not None
+                    if command not in self._inflight_command_ids
                 }
+                self._seen_nonces = {
+                    n: t
+                    for n, t in self._seen_nonces.items()
+                    if t > cutoff
+                    or self._nonce_commands.get(n) in self._inflight_command_ids
+                }
+                self._nonce_commands = {
+                    n: command
+                    for n, command in self._nonce_commands.items()
+                    if n in self._seen_nonces
+                }
+                for expired_command_id in expired_command_ids:
+                    self._completed_results.pop(expired_command_id, None)
                 self._seen_nonces[nonce] = int(envelope.get("ts", now_ms()))
+                if command_id:
+                    self._nonce_commands[nonce] = str(command_id)
         except _DuplicateDelivery:
+            cached = self._completed_results.get(str(command_id or ""))
+            if cached is not None:
+                log.info("control: re-acking completed command %s", command_id)
+                await self._send_ack(command_id, cached, ws=ws)
+                return
+            if command_id and str(command_id) in self._inflight_command_ids:
+                log.info(
+                    "control: duplicate delivery suppressed while command %s is running",
+                    command_id,
+                )
+                return
             log.debug(
-                "control: duplicate delivery suppressed for command %s",
+                "control: duplicate delivery suppressed for unknown command %s",
                 command_id,
             )
             result = {"ok": False, "error_code": "command_rejected"}
@@ -765,11 +805,18 @@ class MachineControlClient:
             result = {"ok": False, "error_code": "command_failed"}
         else:
             execution = self._execute_and_ack(
-                ws,
+                (
+                    None
+                    if background_create and command_id and decrypted["op"] == "create"
+                    else ws
+                ),
                 command_id,
                 decrypted,
                 pairing,
+                nonce=str(nonce) if nonce else None,
             )
+            if command_id:
+                self._inflight_command_ids.add(str(command_id))
             if background_create and command_id and decrypted["op"] == "create":
                 # A first Docker image build can take minutes. Keep that wait
                 # out of the one machine-control receive loop so pause/resume,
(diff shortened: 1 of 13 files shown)


src/puffo_agent/portal/control/client.py
```diff
@@ -207,9 +212,47 @@ def _preflight(context: dict) -> None:
         return {"ok": False, "error": exc.reason, **exc.fields}
     except ControlError as exc:
         return {"ok": False, "error": str(exc)}
+    startup = await _wait_for_agent_start(result["agent_id"])
```
hellohanchen, line 217: **This ten-minute wait runs inside the WS receive loop, so it blocks every other command to this machine.**

`client.py:689` awaits the handler inline inside `async for msg in ws`:

```python
elif frame.get("type") == "command":
    await self._handle(ws, frame)
```

No further frame is read until this returns. Before this PR the create handler returned as soon as the bundle was written, so that serialization never mattered. Now `pause`, `resume`, `archive`, `refresh_usage`, every runtime command and every further `create` — for all linked operators, since it's one WS per machine — queue behind a single slow Docker build for up to ten minutes.

And the machine looks healthy the whole time: `_heartbeat_loop` is a separate spawned task, so heartbeats keep flowing and the server keeps counting it live. Unresponsive and undiagnosable simultaneously.

The pieces to avoid this are already in place — `reporter.set_sender` gives you a reverse channel and `starting` is now a real persisted phase. Ack once the bundle is materialized and report the terminal result asynchronously, or spawn the wait and ack from that task. Anything but holding the receive loop.

src/puffo_agent/portal/control/client.py
```diff
@@ -42,6 +42,11 @@
 # Control-WS heartbeat: liveness ping + capability re-check cadence. Must stay
 # well under the server's HEARTBEAT_TIMEOUT (90s) or the server culls us.
 HEARTBEAT_INTERVAL_SECONDS = 30.0
+# Remote create waits for the daemon's durable worker lifecycle instead of
+# acknowledging as soon as agent.yml exists. Docker's first image build may
+# legitimately take several minutes.
+AGENT_START_TIMEOUT_SECONDS = 10 * 60.0
```
hellohanchen, line 50: **Ten minutes is exactly the server's command expiry, and this clock starts later — so it can never win.**

puffo-server, at command admission (current `main`):

```rust
let expires_at = Utc::now() + Duration::minutes(10);
```

with a 30-second sweep moving `queued|delivered|acked` past `expires_at` to `expired`, and `machines_ws.rs` discarding a late ack as `AcknowledgeOutcome::AlreadyTerminal` — a warning, and nothing reaches the operator.

This deadline starts after delivery, bundle verification, preflight, identity materialization and the agent write. So it is the server's deadline shifted forward by all of that, and any create that actually uses the budget acks into an expired command.

Note which case that hits hardest: the `agent_start_timeout` branch below, written precisely to tell the operator that startup didn't finish, is the branch guaranteed to fire after expiry. It can essentially never be delivered — a hung create still looks to the operator like a create that vanished.

These two numbers must satisfy `client_budget + everything_before_it < server_expiry`. They're currently the same literal in two repos with nothing connecting them. Derive one from the other and comment the coupling on both sides.

src/puffo_agent/portal/control/client.py
```diff
@@ -207,9 +212,47 @@ def _preflight(context: dict) -> None:
...
 
 
+async def _wait_for_agent_start(agent_id: str) -> dict | None:
+    """Return ``None`` once the worker is running, else a structured failure."""
+    from ..state import RuntimeState
+
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + AGENT_START_TIMEOUT_SECONDS
+    while loop.time() < deadline:
+        runtime = RuntimeState.load(agent_id)
```
hellohanchen: `RuntimeState.load` is synchronous — `path.exists()`, `open()`, `json.load` — and this polls it at 10 Hz for up to 600 s. Up to 6000 blocking reads on the control loop, sharing it with the heartbeat task.

There's a subtlety too: the worker writes this file concurrently, and a torn read is swallowed by `load`'s own try/except, so it comes back as "no state yet" and the poll continues. Harmless as written, but it's harmless by accident rather than by design.

`asyncio.to_thread` is already the convention in this file — `last_caps = await asyncio.to_thread(build_capabilities)`. That's the small fix.

The larger one: the worker already exposes `_warm_done`, an `asyncio.Event` the daemon already awaits. Polling the filesystem exists only because the create path can't reach the Worker object. Solving that gets you an immediate, exact signal instead of a 100 ms-granular guess — and it removes the need for this loop entirely.

hellohanchen (commented):
## Review of `ac6f13b88da8f58cbd59dcd48516ed4ac2a4772e` — verdict: **BLOCK**

Two blockers, both in the same new mechanism: making the create ack wait for durable `running` is the right *goal*, but the way it waits blocks the machine's whole command channel, and its deadline is set to a value it can never win against.

The Docker preflight is good work and I found nothing wrong with it. Details at the end, including the two things I checked specifically because they looked wrong and turned out not to be.

---

### B1 — the create handler blocks every other command on this machine for up to 10 minutes

`client.py` reads frames in a single loop and awaits the handler inline:

```python
elif frame.get("type") == "command":
    await self._handle(ws, frame)          # client.py:689
```

Nothing else is read from the socket while that await is outstanding. Before this PR `_create_agent_command` returned as soon as the bundle was written, so the serialization was invisible. Now it can sit in `_wait_for_agent_start` for the full `AGENT_START_TIMEOUT_SECONDS = 10 * 60.0`.

For those ten minutes the machine ignores every `pause`, `resume`, `archive`, `refresh_usage`, every runtime command, and every subsequent `create` — for **all** linked operators, since it's one WS per machine. And it does so while looking perfectly healthy: `_heartbeat_loop` is a separate spawned task, so heartbeats keep flowing and the server keeps the machine marked live. One slow Docker image build — which the new comment explicitly says may "legitimately take several minutes" — makes the machine unresponsive and undiagnosable at the same time.

The pieces to fix it are already here. The reverse channel exists (`reporter.set_sender`, agent.status), and `starting` is now a real persisted phase. Ack the create once the bundle is materialized and report the terminal startup result over the reverse channel; or spawn the wait and ack from that task. What it can't do is hold the receive loop.

### B2 — the client's 10-minute deadline can never beat the server's 10-minute expiry

Server side, at command admission:

```rust
let expires_at = Utc::now() + Duration::minutes(10);
```

with a sweep every 30 s flipping `queued|delivered|acked` past `expires_at` to `expired`, and `machines_ws.rs` dropping a late ack as `AcknowledgeOutcome::AlreadyTerminal` with a warning and nothing else. (Checked against current `main` of puffo-server, not a local checkout.)

The client's budget is the same literal — ten minutes — but it starts strictly *later*: after delivery, bundle verification, preflight, identity materialization and the agent write. So the client's deadline is the server's deadline shifted forward by however long all of that took. Any create that actually consumes the budget produces an ack the server has already discarded.

The cruel part is which case this hits hardest. The `agent_start_timeout` branch — the one written specifically to tell the operator that startup didn't finish — is the branch guaranteed to fire after the server has expired the command. It can essentially never be delivered. A create that hangs will look to the operator exactly like a create that vanished, which is the behaviour this PR set out to fix.

Two numbers that must satisfy `client_budget + everything_before_it < server_expiry` are currently set to the same constant in two repos with no comment linking them. Derive one from the other, and leave a note in both places saying they're coupled.

### H1 — 6000 synchronous file reads on the event loop

```python
while loop.time() < deadline:
    runtime = RuntimeState.load(agent_id)      # sync: path.exists() + open() + json.load
    ...
    await asyncio.sleep(AGENT_START_POLL_SECONDS)
```

At 0.1 s for up to 600 s that's up to 6000 blocking `stat`+`open`+parse calls on the control loop. Individually short, but they land on the same loop as the heartbeat task, and the file is being written concurrently by the worker — a partial read is caught by `load`'s own try/except, which means a torn write reads as "no state yet" and the poll continues. Benign here, but it's the sort of thing that only stays benign by accident.

`asyncio.to_thread` is already the convention in this very file (`last_caps = await asyncio.to_thread(build_capabilities)`). Better still: the worker already has `_warm_done`, an `asyncio.Event`, and the daemon already awaits it — the polling exists only because the create path can't reach the Worker object. That's worth solving directly rather than by polling the filesystem at 10 Hz.

### M1 — a timed-out create leaves a live agent reported as failed

On timeout the ack is `{ok: false, error_code: "agent_start_timeout"}`, but the agent is fully materialized and the worker may reach `running` a second later. Nothing retracts the failure, so the operator's view and the machine's reality diverge permanently in the one case where the timeout was too tight. Given that PR #845 is specifically about reporting truthful readiness, these two need to agree on who owns the correction — otherwise the "failed" create shows up as a working agent and neither side is wrong on its own terms.

---

### Checked and clean

Two things looked like bugs on a first read and are not. Recording them so nobody re-raises them:

**The Docker probe is correct.** `_docker_not_ready("daemon_unavailable")` fires when `container_state` returns `None`, and the probe container `puffo-runtime-preflight-{agent_id}` will never exist — which looks like it should fail every time. It doesn't: `container_state` returns `""` for an absent container and `None` only when `docker container ls` exits non-zero, so "container missing" and "daemon down" are properly distinguished. I read the implementation rather than inferring it from the name.

**Stale runtime state can't short-circuit the wait.** `_wait_for_agent_start` returns success on a persisted `running`, which would be wrong if a previous incarnation's `runtime.json` survived under the same `agent_id`. It can't: `write_agent_from_context` raises if `agent_yml_path` or `agent_dir` already exists, so the directory is always fresh.

Also fine: `wait_warm`'s new return value can't observe a stale status — `_run_post_warm_gate` sets `status = "running"` and saves *before* `_warm_done.set()`. And `preflight_runtime` returning early for `RUNTIME_CLI_DOCKER` skips the Pi/OpenCode checks, but the old guard returned for any non-`RUNTIME_CLI_LOCAL` kind too, so that's unchanged behaviour.

I did not re-run your four mutation checks. The blockers above are control-flow and cross-repo-constant facts that no test outcome changes, so mutation results wouldn't move the verdict either way. If you want them audited, say so and I'll re-run each at the real call site rather than the helper.

CI is green on this head (pytest 3.11/3.12, pre-commit, CodeQL). Nothing merged — this is a review only.

src/puffo_agent/portal/control/client.py
hellohanchen: **Withdrawing this — B2 is wrong. The author is right and I checked it rather than taking their word for it.**

I traced the create command end to end on `main` (`5e8dd728`):

- `portal-store.ts:949` — `sendCommand(machine, null, "create", bundle)`
- `portal-store.ts:824` — `POST /v2/machines/${machine_id}/command`
- `machines.rs:956` `submit_command`, and `:984` `let expires_at = Utc::now() + Duration::seconds(COMMAND_TTL_SECONDS);` with `COMMAND_TTL_SECONDS = 24 * 60 * 60` at `:22`

The ten-minute `Duration::minutes(10)` I cited is `agent_runtime_commands.rs:342`, which inserts rows carrying `delivery_channel` / `runtime_action` / `runtime_target_ref`. Every caller of that module is a runtime-decision path — `agent_runtime_events.rs`, `v2/agent_runtime_events.rs`, `cloud_agent/ws.rs` — not portal create. Both paths write to the same `machine_commands` table, which is what misled me; the expiry is per row, and the maintenance sweep expires each row by its own `expires_at`, so a create row lives 24 hours.

So a ten-minute client-side wait cannot lose to the server expiry on this path, and there is no cross-repo coupling to document. My conclusion here was drawn from the wrong queue.

B1 (`:215`) is unaffected and stands.

src/puffo_agent/portal/control/client.py
```diff
@@ -696,6 +750,48 @@ async def _handle(self, ws: aiohttp.ClientWebSocketResponse, frame: dict) -> Non
...
+                decrypted,
+                pairing,
+            )
+            if background_create and command_id and decrypted["op"] == "create":
+                # A first Docker image build can take minutes. Keep that wait
+                # out of the one machine-control receive loop so pause/resume,
+                # runtime decisions, and other operators remain responsive.
+                task = spawn(execution, name=f"control.create:{command_id}")
+                self._command_tasks.add(task)
+                task.add_done_callback(self._command_tasks.discard)
```
hellohanchen, line 826: **Blocker — the create task now outlives the socket it must ack on, and losing that ack is silent.**

`_connect_once` holds the websocket in an `async with session.ws_connect(...)` block. When the receive loop breaks — a drop, a server-side close, a `stop` — that block exits and the socket is closed. The background create task keeps running and keeps a reference to the now-dead `ws`. Nothing in `_connect_once` cancels `self._command_tasks` on the way out, and `run()` reconnects on a fresh socket the old task never learns about.

Your own new test already shows the first half of this: it asserts `len(client._command_tasks) == 1` *after* `_connect_once` has returned. I continued from exactly that point — released the create, made `send_json` raise the way a closed transport does, and awaited the task:

```
PROBE: _connect_once returned with 1 create task(s) still running
PROBE: create-1 ack recorded?   False
PROBE: task exception?          None
PROBE: nonce still deduped?     True
PROBE: TS_WINDOW_MS = 300.0 s ; AGENT_START_TIMEOUT_SECONDS = 600.0 s
```

The ack is dropped at `client.py:832` — `log.debug` — and the task finishes clean. Nothing anywhere records that a completed create was never acknowledged.

What happens next is server-side and not hypothetical. `drain_queue_once` in `machines_ws.rs:394` re-pushes every command with `status IN ('queued','delivered') AND expires_at > NOW()`, on every reconnect, and the Portal command TTL is the 24 hours your new comment at `:47` correctly names. So the create is redelivered.

Whether that redelivery is caught is decided by the last two numbers above, and they disagree: **the nonce window is 5 minutes and the create budget is 10.** Past the five-minute mark `_seen_nonces` has pruned the entry, `_DuplicateDelivery` does not fire, and `execute_command` runs `create` a second time. `provision.py:297` does stop it from clobbering anything — but only *after* `provision_agent_from_bundle` has re-run `_materialize_slug_binding` against the server with the already-consumed `pending_token`, and the command then acks `{ok: false, error: "agent 'x' already exists on this daemon"}`. A create that succeeded is recorded as failed.

This is not a corner case; it is the modal case for the thing this change exists to support. `AGENT_START_TIMEOUT_SECONDS` is ten minutes because a first Docker image build takes minutes, and a reconnect inside a ten-minute window is ordinary. Before this delta the create was awaited inline, so it could only complete on a socket that was still being read from — the ack was reliable precisely *because* of the blocking you have now removed. Moving the work off the loop is right; moving it off without giving the ack a path that survives the socket is what turns it into a defect.

Three things would close it, and they are separable: settle the ack against the *current* connection rather than the captured `ws` (or re-ack on reconnect from a small pending-ack set); make a failed ack an observable event rather than `log.debug`; and make the dedup horizon at least as long as the command it is meant to deduplicate — `TS_WINDOW_MS` under `AGENT_START_TIMEOUT_SECONDS` cannot do its job for the ops that take longest.

hellohanchen, line 826: **`_command_tasks` is pruned but not capped.**

`add` plus a `discard` done-callback keeps the set from growing without bound *for tasks that finish*. It does not bound how many can be in flight at once, and each one can hold a Docker image build for up to ten minutes. The set is fed directly from the wire: one entry per `create` frame the server pushes.

The dedup ahead of it is per-nonce, so distinct commands all pass. And `drain_queue_once` re-pushes the whole queued backlog on every reconnect, so a machine that was offline while several creates were issued gets them all at once, each spawning its own task and its own build.

Not a blocker — the realistic ceiling is however many creates an operator actually issued, and the daemon has other limits. But "pruned" and "capped" are different questions and only the first one is answered here. A small in-flight ceiling that acks the overflow with a retryable error is cheaper than discovering the bound empirically.

src/puffo_agent/portal/control/client.py
```diff
@@ -42,6 +42,13 @@
 # Control-WS heartbeat: liveness ping + capability re-check cadence. Must stay
 # well under the server's HEARTBEAT_TIMEOUT (90s) or the server culls us.
 HEARTBEAT_INTERVAL_SECONDS = 30.0
+# Remote create waits for the daemon's durable worker lifecycle instead of
+# acknowledging as soon as agent.yml exists. Docker's first image build may
+# legitimately take several minutes. The server's Portal machine-command TTL
+# is 24 hours; this shorter bound is a UI/operator feedback budget, not the
+# command-expiry budget used by Runtime permission/cancel commands.
+AGENT_START_TIMEOUT_SECONDS = 10 * 60.0
```
hellohanchen, line 50: Verified rather than assumed, since a comment crediting another layer is a claim: the server's Portal machine-command TTL really is 24 hours (`machines.rs:22`, used at `:984` in `submit_command`), and the 10-minute figure really is the Runtime path (`agent_runtime_commands.rs:342`). Both halves of this comment are correct, and it names the distinction that my withdrawn B2 got wrong last round. Good comment to have added.

One consequence worth carrying forward, since the comment now makes it visible: a 24-hour server TTL is what gives the redelivery in my other note its long tail. An un-acked create is not retried once — it is re-pushed on every reconnect for a day.

src/puffo_agent/portal/control/client.py
```diff
@@ -207,9 +214,50 @@ def _preflight(context: dict) -> None:
...
+async def _wait_for_agent_start(agent_id: str) -> dict | None:
+    """Return ``None`` once the worker is running, else a structured failure."""
+    from ..state import RuntimeState
+
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + AGENT_START_TIMEOUT_SECONDS
+    while loop.time() < deadline:
+        # RuntimeState.load performs filesystem I/O. Keep repeated startup
+        # observation off the single control event loop.
+        runtime = await asyncio.to_thread(RuntimeState.load, agent_id)
```
hellohanchen, line 236: No finding — recording it so the absence is deliberate.

Replacing `asyncio.to_thread(RuntimeState.load, …)` with the direct call survives the suite (`36 passed`), which is the expected result for a change whose whole effect is *where* the work runs. I am not asking for a test; a timing assertion here would be a flake generator. The 0.1 → 0.5 s poll change is likewise unobservable and fine — at a ten-minute budget it costs at most half a second of detection latency and saves 1,080 filesystem round-trips per create.

For the record, the three mutations that do carry behaviour all die:

| # | Mutation | Result |
|---|---|---|
| Z1 | `_connect_once` call site drops `background_create=True` | KILLED (`TimeoutError`) |
| Z2 | background dispatch stops being restricted to `create` | KILLED |
| Z4 | spawned task not added to `_command_tasks` | KILLED |

`Z1` is the one I most wanted to see fail, because it is the difference between testing the helper and testing the call site. It targets the real dispatch point and it goes red there. That is the right shape.

hellohanchen (commented):
## Round 2 — verdict: **BLOCK** (one blocker)

Delta reviewed at `72dbf360bcdb3698bc883a956d15a57c21262c11`, taken against my own last-reviewed SHA `ac6f13b8` — 2 files, +206/−25. All seven checks on this head are terminal `success`; I pulled the rollup at the start of the round rather than the end, and re-queried the head before submitting. Nothing merged.

### B1 is genuinely fixed, and the test targets the real call site

I mutated rather than read. Baseline `36 passed` on `tests/test_control_client.py`:

| # | Mutation | Result |
|---|---|---|
| Z1 | `_connect_once` call site drops `background_create=True` | **KILLED** (`TimeoutError`) |
| Z2 | background dispatch no longer restricted to `create` | **KILLED** |
| Z4 | spawned task not added to `_command_tasks` | **KILLED** |
| Z3 | `asyncio.to_thread(RuntimeState.load, …)` → direct call | SURVIVED (expected) |

`Z1` is the one that matters. The new test drives the real dispatch point, not a helper, and turning that point off is what makes it red — which is exactly what you said it did. The ordering is right too: verify, decrypt, and nonce-dedup all still happen inline on the receive loop, so replay protection has not moved into the concurrent region.

`Z3`'s survival is not a finding and I am not asking for a test — a timing assertion there would be a flake generator.

### The blocker: the create now outlives the socket it must ack on

Moving the wait off the receive loop is the right call. But the ack was reliable *because* of the blocking that has been removed: an inline create could only finish on a socket that was still being read from. Now the task keeps a reference to a `ws` whose `async with session.ws_connect(...)` block has already exited, nothing cancels `_command_tasks` when `_connect_once` returns, and a failed ack is swallowed at `log.debug`.

Your own new test asserts `len(client._command_tasks) == 1` **after** `_connect_once` has returned — that is the first half of this already in the suite. I continued from that exact point:

```
PROBE: _connect_once returned with 1 create task(s) still running
PROBE: create-1 ack recorded?   False
PROBE: task exception?          None
PROBE: nonce still deduped?     True
PROBE: TS_WINDOW_MS = 300.0 s ; AGENT_START_TIMEOUT_SECONDS = 600.0 s
```

The server side is not hypothetical: `drain_queue_once` (`machines_ws.rs:394`) re-pushes anything `status IN ('queued','delivered')` while `expires_at > NOW()`, on every reconnect, for the 24 hours your new comment correctly names. And the last two numbers disagree — **the nonce window is five minutes, the create budget is ten.** Past five minutes the dedup entry is pruned, the redelivered create executes again, `provision.py:297` refuses it as *"already exists"* (after re-running `_materialize_slug_binding` with a spent `pending_token`), and a create that succeeded is recorded as failed.

A reconnect inside a ten-minute window is ordinary, and ten minutes is the budget precisely because first Docker builds are slow — so this lands on the case the change exists to serve. Details and the three separable fixes are inline at `client.py:779`.

### Findings, not blocking

**N1 — `_command_tasks` is pruned but not capped** (`client.py:779`). Different questions; only the first is answered. Fed straight from the wire, and a reconnect drains the whole backlog at once.

**N2 — the timeout copy now matches the code.** *"remains configured and may still finish starting, so check its status before retrying"* claims nothing is cancelled or rolled back, and nothing is: `_wait_for_agent_start` returning a failure dict leaves the provisioned agent in place. I checked the code against the sentence rather than trusting the sentence — same class as the comment I had to verify last round.

**N3 — the TTL comment at `:47` is correct on both halves** (`machines.rs:22` for the 24 h Portal path, `agent_runtime_commands.rs:342` for the 10 min Runtime path). That is the distinction my withdrawn B2 got wrong, and it is worth having written down. Inline.

All experiments ran in my own worktree and were reverted; `git status --porcelain` is empty. Round 2 of 3. Not merged — that is @mingvase-8795's call.

src/puffo_agent/portal/control/client.py
```diff
@@ -637,17 +692,21 @@ async def _report(operator_slug: str, envelope: dict) -> None:
                             continue
                         if frame.get("type") == "connected":
                             if not reporter_ready:
+                                self._active_ws = ws
```
hellohanchen, line 695: **This line is the only ack route a background create has, and deleting it keeps the suite green.**

`_handle` deliberately passes `None` instead of `ws` into `_execute_and_ack` for a background create — that is round 2's fix. So when the create finishes, `_send_ack`'s `target = self._active_ws or ws` has nothing in `ws` to fall back to: `_active_ws` is the whole delivery path.

Deleting `self._active_ws = ws` leaves the suite at **46 passed (46)**. In production the same deletion means a create that completes on a healthy connection never acks until the next reconnect, while the server holds the command `delivered` and keeps redelivering it.

The reason nothing notices is that no test covers a background create finishing while the socket is still up. `test_create_wait_does_not_block_followup_machine_command` asserts the opposite — at `:76` it requires `create-1` to be absent from `ws.acks`, which holds because its stream has ended and the `finally` already cleared `_active_ws` — and then checks delivery only after a reconnect. `test_failed_ack_is_warned_and_retried` assigns `client._active_ws` by hand, so it never exercises this assignment either.

A test that lets the create complete with the receive loop still running, and asserts the ack lands on the live socket, would pin it. That is also the common case in production, so it is worth having for its own sake rather than only to kill the mutant.

src/puffo_agent/portal/control/client.py
```diff
@@ -707,29 +852,70 @@ async def _handle(self, ws: aiohttp.ClientWebSocketResponse, frame: dict) -> Non
...
+        self,
+        command_id: str | None,
+        result: dict,
+        *,
+        ws: aiohttp.ClientWebSocketResponse | None = None,
+    ) -> None:
+        if not command_id:
+            return
+        command_key = str(command_id)
+        self._pending_acks[command_key] = result
```
hellohanchen, line 887: **A queued success can be replaced by a later failure for the same command.**

The assignment is unconditional, so `_pending_acks` is last-write-wins per command id. Driving the real method with the socket down:

```
after success ack :  {'cmd-X': {'ok': True, 'agent_id': 'a1'}}
after rejection   :  {'cmd-X': {'ok': False, 'error_code': 'command_rejected'}}
```

The next flush then reports a create that succeeded as failed — round 2's failure mode, in a narrower window.

Two routes, both needing the result still queued:

- The `connected` flush fails (the case the `queued for reconnect` branch exists for) and leaves the result queued on a live connection. A redelivery of the same command, now outside the 15-minute ts window, raises `ControlError` and produces `command_rejected`.
- The operator pairing is removed or rotated between the create and the redelivery. `load_pairings().get(operator_slug)` raises `ControlError` *before* the `nonce in self._seen_nonces` check, so the redelivery never reaches the duplicate branch that would have re-acked the cached result — it is rejected instead, and overwrites.

`_completed_results` already holds the authoritative outcome in the first case. A rule that a queued `ok` is not downgraded — or consulting `_completed_results` before writing a rejection — would close both routes.

Separately, and much smaller: this map is the only one added this round without a lifetime. `_seen_nonces`, `_nonce_commands` and `_completed_results` are all pruned; an entry here is removed only on a successful send, so a result queued for a machine that never comes back is held for the life of the process.

hellohanchen (commented):
Round 3, bound to `59787cb6b52f18353f851a0077f9cdc5fd96885e`. Delta taken from `72dbf360` — the SHA I reviewed in round 2 — which is an ancestor of this head, so the comparison is clean.

**Round 2's blocker is fixed, and the fix is pinned.** Verdict: **no blockers.** Four findings, all coverage or narrow hazards.

## Round 2's blocker

Restoring the captured socket for the background create — reverting the fix to `ws` in place of the `None if background_create …` expression — fails the suite:

```
[BASELINE]                          rc=0  46 passed          SURVIVED
[A1_recapture_ws_for_background]    rc=1  45 passed 1 failed KILLED
```

The orphaned-ack defect is genuinely closed and genuinely tested. Your three claimed kills also reproduce independently, on my own mutants rather than your report:

```
[C4_warning_to_debug]               KILLED
[E1_ts_window_back_to_5]            KILLED
[A1_recapture_ws_for_background]    KILLED
```

Plus five more of the new mechanisms: `[B2_no_flush_on_connected]`, `[B3_never_clear_active_ws]`, `[C3_ack_failure_does_not_requeue]`, `[D3_no_cached_reack]`, `[D4_no_inflight_suppression]`, `[D5_never_store_result]` — all KILLED.

## V1 — the ordinary background-create path has no test

Details inline. `self._active_ws = ws` at `client.py:695` can be deleted with the suite at **46 passed (46)**, and it is the *only* delivery route a background create has — `_execute_and_ack` is deliberately given `ws=None`, so `_send_ack`'s `target = self._active_ws or ws` resolves to `_active_ws` or to nothing.

Delete it in production and a create that finishes on a healthy connection never acks until the next reconnect, while the server holds the command `delivered` and keeps redelivering. The suite cannot see that, because no test covers a background create completing while the socket is still up.

## V2 — five of this round's mechanisms are unpinned, and they delete together

Each of these survives alone at 46/46:

```
[B1_never_set_active_ws]                    SURVIVED   client.py:695
[C1_flush_stale_socket_guard_gone]          SURVIVED   the `ws is not self._active_ws` guard in _flush_pending_acks
[C2_pop_ack_unconditionally]                SURVIVED   the `is result` identity check before popping
[D1_no_inflight_nonce_exemption]            SURVIVED   the in-flight arm of the nonce prune
[D2_no_completion_nonce_refresh]            SURVIVED   the completion-time nonce refresh
```

Individually that proves little — two guards on one path make each other unobservable, which is why I combined them:

```
[F1_active_ws_registration_and_guard_both_gone]   46 passed  SURVIVED
[F2_both_nonce_retention_halves_gone]             46 passed  SURVIVED
[F3_active_ws_set_and_clear_gone]                 46 passed  SURVIVED
[F4_all_five_survivors_together]                  46 passed  SURVIVED
```

So the whole `_active_ws` mechanism, the pending-ack identity check, and both halves of the nonce retention can be removed at once and nothing goes red. Round 3's fix rests on exactly two tested behaviours: flush-on-`connected` and requeue-on-failure. Everything added to make *socket selection* and *nonce lifetime* correct is invisible.

One honest caveat on D1, because I would rather flag it than assert it: the in-flight exemption looks close to unreachable as written. A nonce is only pruned once its ts is more than 15 minutes old, and the create budget is 10 minutes (`AGENT_START_TIMEOUT_SECONDS`), so by the time pruning could drop it the create has already timed out and D2 has refreshed the nonce to completion time. If that reading is right, D1 is dead code rather than an untested guard. Worth your confirming which.

## V3 — a later rejection overwrites a queued success for the same command

`_send_ack` writes `self._pending_acks[command_key] = result` unconditionally. Driving the real method:

```
after success ack :  {'cmd-X': {'ok': True, 'agent_id': 'a1'}}
after rejection   :  {'cmd-X': {'ok': False, 'error_code': 'command_rejected'}}
```

The queued successful create is replaced by a failure, and that failure is what the next flush sends. Two routes reach it, both requiring the ack to still be queued:

- The `connected` flush fails — the exact case `C3` exists for — leaving the result queued on a live connection. A redelivery of the same command whose ts has since aged past the 15-minute window then raises `ControlError` and produces `command_rejected`.
- The operator pairing is removed or rotated. `load_pairings().get(operator_slug)` raises `ControlError` **before** the nonce duplicate check runs, so a redelivery is rejected rather than recognised as a duplicate, and clobbers the queued result.

Neither is the common case, but the outcome is round 2's failure mode in a narrower window: a create that actually succeeded reported to the operator as failed. There is no rule anywhere that a queued `ok` may not be downgraded.

## V4 — `_pending_acks` is the one map here with no lifetime

`_seen_nonces`, `_nonce_commands` and `_completed_results` all have a pruning rule this round. `_pending_acks` has none: an entry is removed only on a successful send, so a result queued for a machine that never reconnects is held for the life of the process, and the map grows with each distinct command id. It is small per entry and I am not calling it a leak of consequence — I am flagging it because it is the asymmetry in an otherwise carefully bounded set of maps.

## Scope

All seven checks on this SHA are green, re-queried immediately before submitting, and the head has not moved. Baseline for every mutation above is `tests/test_control_client.py tests/test_control_envelope.py` at 46 passed (46), matching the focused count you reported.

Experiments ran in my own worktree and are reverted; `git status --porcelain` is empty. I did not touch the branch and I am not merging anything.

This is round 3, which is my cap. If anything here is still open afterwards, please take it to @liangx-poppy-37c3 rather than sending a round 4 my way.

PR #314, opened 2026-09-04
fix(codex): preserve large session resume

Glimmer0x:
## Summary

- request Codex `thread/resume` with `excludeTurns: true`
- preserve the native thread and provider-side context while preventing full reconstructed history from becoming one oversized stdout JSONL frame
- pin the request contract with a regression test

## Why

A real Codex session with a 278 MB rollout could no longer resume because the default full-history response exceeded Puffo Agent’s 16 MiB stdout frame limit before `turn/start`. Codex 0.145.0-alpha.18 already supports its metadata-only resume response: the thread is fully restored, but historical turns are not serialized back to the client.

## Verification

- mutation: removing `excludeTurns` makes the regression test fail
- targeted driver/transport suite: 67 passed, 1 skipped
- full suite: 3883 passed, 2 skipped
- pre-commit: all hooks passed
- Python structure check: passed

### DIFF
--- a/src/puffo_agent/agent/harness/drivers/codex.py
+++ b/src/puffo_agent/agent/harness/drivers/codex.py
@@ -391,6 +391,13 @@ class CodexAppServerDriver(Driver):
                 "thread/resume",
                 {
                     "threadId": str(resume),
+                    # Puffo already persists its own visible event history and
+                    # only needs Codex to restore the provider-side context.
+                    # Returning every reconstructed turn can turn a large
+                    # rollout into one enormous JSONL response before a new
+                    # turn even starts. Codex still resumes the full thread;
+                    # this only keeps that history out of the response.
+                    "excludeTurns": True,
                     **thread_config,
                 },
             )
--- a/src/puffo_agent/agent/harness/drivers/codex.py
+++ b/src/puffo_agent/agent/harness/drivers/codex.py
@@ -356,7 +356,10 @@ class CodexAppServerDriver(Driver):
             "initialize",
             {
                 "clientInfo": {"name": "puffo-agent", "version": "1"},
-                "capabilities": {},
+                # Codex gates thread/resume.excludeTurns behind this client
+                # capability. Without it, supported app-server versions reject
+                # every resume request before attempting to load the thread.
+                "capabilities": {"experimentalApi": True},
             },
         )
         await self._write({"method": "initialized", "params": {}})
--- a/tests/test_codex_resume.py
+++ b/tests/test_codex_resume.py
@@ -1,0 +1,25 @@
+from unittest.mock import AsyncMock
+
+import pytest
+
+from puffo_agent.agent.harness.drivers.codex import CodexAppServerDriver
+
+
+@pytest.mark.asyncio
+async def test_initialize_enables_api_required_by_exclude_turns():
+    driver = CodexAppServerDriver()
+    driver._request = AsyncMock(return_value={})
+    driver._write = AsyncMock()
+
+    await driver._initialize_app_server()
+
+    driver._request.assert_awaited_once_with(
+        "initialize",
+        {
+            "clientInfo": {"name": "puffo-agent", "version": "1"},
+            "capabilities": {"experimentalApi": True},
+        },
+    )
+    driver._write.assert_awaited_once_with(
+        {"method": "initialized", "params": {}}
+    )
--- a/tests/test_harness_driver.py
+++ b/tests/test_harness_driver.py
@@ -1084,7 +1084,7 @@ async def test_claude_driver_emits_compaction_boundary_and_clears_tool_calls():
 
 
 @pytest.mark.asyncio
-async def test_codex_driver_resumes_with_native_session_id_after_handshake():
+async def test_codex_driver_resumes_without_returning_full_turn_history():
     holder = {}
 
     def on_frame(frame):
@@ -1094,7 +1094,7 @@ async def test_codex_driver_resumes_with_native_session_id_after_handshake():
         elif frame.get("method") == "thread/resume":
             assert frame["params"] == {
                 "threadId": "native-thread",
-                "cwd": "/workspace",
+                "excludeTurns": True, "cwd": "/workspace",
                 "approvalPolicy": "never",
                 "sandbox": "danger-full-access",
                 "model": "gpt",


src/puffo_agent/agent/harness/drivers/codex.py
```diff
@@ -391,6 +391,13 @@ async def _open_thread(
                 "thread/resume",
                 {
                     "threadId": str(resume),
+                    # Puffo already persists its own visible event history and
+                    # only needs Codex to restore the provider-side context.
+                    # Returning every reconstructed turn can turn a large
+                    # rollout into one enormous JSONL response before a new
+                    # turn even starts. Codex still resumes the full thread;
+                    # this only keeps that history out of the response.
+                    "excludeTurns": True,
```
hellohanchen, line 403: **This field is rejected by the Codex version this repo pins.**

Driving a real `codex app-server` (0.147.0, the `CODEX_NPM_VERSION` at `docker_support.py:28`; identical on 0.149.1):

```
excludeTurns: true        -> -32600 "thread/resume.excludeTurns requires experimentalApi capability"
zzzUnknownParam: true     -> -32600 "no rollout found for thread id 01a0…"
(no extra param)          -> -32600 "no rollout found for thread id 01a0…"
```

The middle line is the control. A parameter the server does not know is **ignored** — the request gets past validation and fails later for an unrelated reason. `excludeTurns` is known and gated, and fails earlier than both.

Changing only `initialize`'s `capabilities` from `{}` to `{"experimentalApi": true}` makes the first line become `"no rollout found"` too — identical to the other two. The gate is the sole cause, and declaring the capability is what clears it.

`_initialize_app_server` at `codex.py:359` sends `"capabilities": {}`, and nothing gates this call on `executable_version` (it is only carried into `ProtocolDiagnostics` at `:320`), so the field goes out unconditionally to whatever `codex` binary the machine has.

Run through the production classifiers, the resulting error is `ProviderFailureError(error_code="provider_error")`; `_native_resume_is_unavailable` and `_resume_error_is_transient` are both `False`. That is `runtime_manager.py`'s unclassified branch, so every agent holding a native Codex session gets: open #1 raises, open #2 raises, open #3 clears the native session and starts fresh. The context loss this PR exists to prevent, on every agent rather than only the oversized ones — and the user only ever sees `"The provider could not complete the turn."`

tests/test_harness_driver.py
```diff
@@ -1094,7 +1094,7 @@ def on_frame(frame):
         elif frame.get("method") == "thread/resume":
             assert frame["params"] == {
                 "threadId": "native-thread",
-                "cwd": "/workspace",
+                "excludeTurns": True, "cwd": "/workspace",
```
hellohanchen, line 1097: **This assertion pins what Puffo sends, which is not the property that fails against a real server.**

It is a correct test of the request shape, and it does its job — I mutated the driver two ways and both die:

```
[BASELINE]                  rc=0  45 passed          SURVIVED
[Z1_delete_excludeTurns]    rc=1  44 passed 1 failed KILLED
[Z2_excludeTurns_false]     rc=1  44 passed 1 failed KILLED
```

So your point 3 is genuinely satisfied. The gap is that `on_frame` is a hand-written double that accepts any parameters, so a field the real app-server refuses is indistinguishable here from one it accepts. That is why all seven checks are green on a change that breaks resume.

The property worth a test is on the response side rather than the request side: that a rejected `thread/resume` does not quietly retire the native session. Today an error containing `"thread"` but none of `_native_resume_is_unavailable`'s markers — which is exactly the shape of this one — costs two hard failures and then the saved session.

hellohanchen (commented):
Independent review of `d0dcdb27e9e963112ef114c7b88a0d72caef8d38` against `main` (`c7ee7260`). Verdict: **BLOCK.**

The diagnosis behind this PR is right and the mechanism is real. The problem is that `excludeTurns` is not a field the shipped Codex will accept from this client, and sending it makes `thread/resume` fail — so the change does not merely fail to help the 278 MB case, it breaks resume for every Codex session.

I ran a real `codex app-server` rather than reasoning from the schema.

## B1 (blocker) — `excludeTurns` is gated behind a capability this client never declares

Driving `codex app-server` over stdio with three variants of `thread/resume` on the same thread:

```
A  {threadId, excludeTurns: true, ...}          -> -32600 "thread/resume.excludeTurns requires experimentalApi capability"
B  {threadId, zzzTotallyUnknownParam: true, ...} -> -32600 "no rollout found for thread id 01a0…"
C  {threadId, ...}                               -> -32600 "no rollout found for thread id 01a0…"
```

B is the counterfactual that makes this readable. A genuinely **unknown** parameter is ignored — the request proceeds past parameter validation and fails later, in rollout lookup, for a reason unrelated to my probe. `excludeTurns` is not ignored: it is a *known* field behind a capability gate, and it fails strictly earlier than B and C do.

Confirming that the gate is the only thing failing — same probe, changing only `initialize`'s `capabilities` from `{}` to `{"experimentalApi": true}`:

```
A  {threadId, excludeTurns: true, ...}          -> -32600 "no rollout found for thread id 01a0…"   (identical to B and C)
```

`codex.py:359` sends `"capabilities": {}`.

**Both results reproduce on `0.147.0` — the version this repo itself pins**, at `docker_support.py:28` (`CODEX_NPM_VERSION = "0.147.0"`) — and on `0.149.1`. So this is not a fringe-version concern; it is the version the container ships. Whatever `0.145.0-alpha.18` did, the gate exists in the shipped one.

There is no version negotiation to soften this: `executable_version` is only carried into `ProtocolDiagnostics` at `codex.py:320` and is never gated on, so the field goes out unconditionally to whatever `codex` binary is on the machine.

### What the failure actually does

I ran the exact error object through the production classifiers rather than predicting:

```
_classify_jsonrpc_error({"code": -32600,
    "message": "thread/resume.excludeTurns requires experimentalApi capability"})
  -> ProviderFailureError(error_code="provider_error")

_native_resume_is_unavailable(exc) -> False
_resume_error_is_transient(exc)    -> False
_RESUME_FAILURE_FALLBACK_AFTER      = 3
```

Neither classifier matches, so it lands in `runtime_manager.py`'s **unclassified** branch. That means, for every agent with a saved native Codex session:

1. open #1 — raises, user sees `"The provider could not complete the turn."`
2. open #2 — raises again
3. open #3 — `_clear_native_session()`, then `driver.open(spec, None)`: a fresh thread, provider-side context discarded

Step 3 is the precise outcome this PR exists to prevent, and it arrives on every agent rather than only on the oversized ones. The real cause never reaches the user — `"requires experimentalApi capability"` appears only in a `logger.warning`.

## Your four verification points

**1. Version semantics and compatibility — this is the blocker.** Covered above. `0.147.0` and `0.149.1` both gate the field.

**2. Puffo does not depend on the resume response's turns — confirmed.** The only read of the `thread/resume` result is `codex.py:301`, `thread.get("id") or thread.get("threadId")`. I grepped `\bturns\b` across `src/` and there is no consumer of a `turns` field anywhere in the tree. The premise is sound.

One thing to add in support of it: the full `result` is also passed as `native_payload` on the `SESSION_RESUMED` emit (`codex.py:308-311`), so the response size matters on the event path too, not only at the transport. That strengthens the case for the change — it just needs to be a change the server will accept.

**3. The mutation is killed — confirmed independently.** Baseline `tests/test_harness_driver.py` is 45 passed, rc=0. Two mutants:

```
[BASELINE]                  rc=0  45 passed          SURVIVED
[Z1_delete_excludeTurns]    rc=1  44 passed 1 failed KILLED
[Z2_excludeTurns_false]     rc=1  44 passed 1 failed KILLED
```

Both die. Point 3 is genuinely satisfied — see N1 for what that does and does not prove.

**4. Adjacent paths.** The overflow mechanism is real and correctly identified: `codex.py:346` sets the subprocess stdout limit to `16 * 1024 * 1024`, and `_read_loop` at `:587-594` catches `RpcFrameTooLarge` and **breaks** — the comment is explicit that stdout is left partially consumed and the session cannot be resynchronized, after which `fail_pending_requests` kills every outstanding request. A 278 MB single-frame resume response hits exactly that, and the resume response is the right thing to shrink.

I could not reproduce a 278 MB rollout here, so I cannot rule out other frames that would exceed 16 MiB on the same session — a very large single tool result would too. I am reporting that as unverified rather than as a clean bill.

## N1 — the test pins the request shape, which is not the property that failed

`test_codex_driver_resumes_without_returning_full_turn_history` asserts on `frame["params"]` inside a hand-written `on_frame` double. That is a correct test of *what Puffo sends*, and it is why both mutants die. It cannot fail for what actually broke here, because the double accepts any parameters — a field the real server refuses looks identical to one it accepts.

The property worth pinning is the response side: that a `thread/resume` rejection is classified in a way that does not silently retire the native session. As it stands, an error whose text happens to contain `"thread"` but none of the `unavailable` markers takes three opens to fall back, and each of the first two surfaces a generic message.

## Verdict

**BLOCK** at `d0dcdb27e9e963112ef114c7b88a0d72caef8d38`. The blocker is B1 and it is one line of `initialize`: the capability has to be declared before the field can be sent, and if the target versions ever disagree about that gate, the field needs to be conditional rather than unconditional.

All seven checks on this SHA are green, which is consistent with N1 — nothing in CI speaks to the real app-server.

Every probe ran in my own worktree and against throwaway threads; the branch was not touched and `git status --porcelain` is empty. Review only — I am not merging anything.

src/puffo_agent/agent/harness/drivers/codex.py
```diff
@@ -356,7 +356,10 @@ async def _initialize_app_server(self) -> None:
             "initialize",
             {
                 "clientInfo": {"name": "puffo-agent", "version": "1"},
-                "capabilities": {},
+                # Codex gates thread/resume.excludeTurns behind this client
+                # capability. Without it, supported app-server versions reject
+                # every resume request before attempting to load the thread.
+                "capabilities": {"experimentalApi": True},
```
hellohanchen, line 362: Verified against a real 13-turn rollout on both 0.147.0 (the version pinned in `docker_support.py`) and 0.149.1: without this key `thread/resume` returns `-32600 thread/resume.excludeTurns requires experimentalApi capability`; with it, resume succeeds and `thread.turns` drops from 13 entries to 0 (2314 vs 20752 bytes). The response skeleton is otherwise identical, so `thread.get("id")` at :301 still resolves.

Also checked the blast radius of getting this wrong: `_initialize_app_server` has one production call site (:299), so a bad capabilities dict fails every session, not just resumes. Mutants `{}`, `False`, and `experimental_api` are all killed, and `tests/test_codex_resume.py` is the only test that kills them — the rest of the suite is entirely green under the `False` mutant.

src/puffo_agent/agent/harness/drivers/codex.py
```diff
@@ -356,7 +356,10 @@ async def _initialize_app_server(self) -> None:
...
             },
         )
         await self._write({"method": "initialized", "params": {}})
```
hellohanchen, line 365: Mutated this line separately rather than relying on the capability mutant, since these two statements form one mechanism and a kill on the first says nothing about the second. Replacing it with `pass` is killed too.

tests/test_codex_resume.py
```diff
@@ -0,0 +1,25 @@
...
+
+@pytest.mark.asyncio
+async def test_initialize_enables_api_required_by_exclude_turns():
+    driver = CodexAppServerDriver()
+    driver._request = AsyncMock(return_value={})
+    driver._write = AsyncMock()
+
+    await driver._initialize_app_server()
+
+    driver._request.assert_awaited_once_with(
```
hellohanchen, line 16: Asserting on the exact params dict is the right shape here. One boundary worth being explicit about: because `_request` is an `AsyncMock`, this test pins what the client *sends*, not what a server *accepts* — it would keep passing if Codex renamed the capability. That gap is covered by the live-server probe I ran on 0.147.0 and 0.149.1, not by this file, and it is why observation 1 in the review body is about the absence of a version gate.

hellohanchen (commented):
## Round 2 — round-1 blocker resolved, no blockers

Reviewed exactly `d0dcdb27` → `ac1344cac2451cfe608b5e2702daddb6cc3075e0` (`d0dcdb27` confirmed an ancestor). Delta: 2 files, +29/−1. All 7 checks terminal SUCCESS on this exact head, re-queried immediately before submitting.

### The blocker is fixed, and I verified it end-to-end against a real rollout

My round-1 block was that `thread/resume.excludeTurns` is capability-gated, so every resume was rejected before the thread was even loaded. Declaring `capabilities: {"experimentalApi": True}` is exactly the right fix, and I confirmed it on a **real persisted rollout** (13 turns) rather than on a mock — on both the version this repo pins in `docker_support.py` (`CODEX_NPM_VERSION = "0.147.0"`) and the host's 0.149.1:

| version | capabilities | `thread/resume` + `excludeTurns` | response |
|---|---|---|---|
| 0.147.0 | `{}` | `-32600 requires experimentalApi capability` | — |
| 0.147.0 | `{"experimentalApi": true}` | **OK** | 2314 bytes |
| 0.149.1 | `{}` | `-32600 requires experimentalApi capability` | — |
| 0.149.1 | `{"experimentalApi": true}` | **OK** | 2328 bytes |

Plain resume on the same thread is 20752 / 21036 bytes, so the stated purpose is real and not just plausible: `thread.turns` goes from **13 entries to 0**, roughly a 9× reduction on a modest thread. Comparing the two response skeletons key-by-key, nothing else changes — in particular `thread.id` is still present, so the native-id extraction at `codex.py:301` (`thread.get("id") or thread.get("threadId")`) still resolves, and `SESSION_RESUMED`'s `native_payload` keeps the same shape.

I also checked that declaring the capability does not widen the frame surface the reader loop has to cope with. Capturing every stdout frame across initialize + resume, the notification set is the same either way (`remoteControl/status/changed`, `mcpServer/startupStatus/updated`, `thread/status/changed`, `thread/goal/cleared`, `skills/changed`) — no experimental method names appear. And a plain resume returns a byte-for-byte identical skeleton with and without the capability, so the flag is narrowly scoped in practice, not a general "turn on experimental behaviour" switch.

### Mutation testing — the new test is the sole pin, and it holds

`_initialize_app_server` is reached from exactly one production call site (`codex.py:299`), so a silent capability regression would break every session, not just resumes. Full suite (`tests/`, minus the pre-existing `test_pi_tool_bridge.py`), baseline green:

| mutant | result |
|---|---|
| `capabilities` back to `{}` | **KILLED** |
| `{"experimentalApi": False}` | **KILLED** |
| key renamed to `experimental_api` | **KILLED** |
| drop the `initialized` notification | **KILLED** |
| drop `"excludeTurns": True` from resume | **KILLED** |

Each mutation was verified to have landed (`git diff --numstat` = 1 changed line every time), and the worktree was confirmed clean afterwards.

Two attributions worth stating, since a green suite does not tell you *which* test is load-bearing. Running `tests/test_codex_resume.py` alone under the `False` mutant: **1/1 fails**, which independently reproduces the claim made in the PR discussion. Running the whole suite **with `test_codex_resume.py` excluded** under the same mutant: **no failures and no errors at all**. So the new file is not redundant coverage — it is the only thing standing between a flipped capability and a silent regression, and it earns its place.

The `initialized` mutant is included because this is a two-statement mechanism; flipping only the capability would have left the notification unpinned. It is pinned.

### Non-blocking observations

**1. No version floor or ceiling on an explicitly experimental surface.** `executable_version` is captured at `codex.py:320` for diagnostics and is never gated on, so the driver sends this capability to whatever `codex` happens to be on PATH. I probed how tolerant `initialize` is: 0.147.0 accepts an entirely made-up capability key (`{"zzzTotallyUnknownCapability": true}` → OK) but does type-check known ones (`{"experimentalApi": "notabool"}` → `-32600 invalid type: string, expected a boolean`). That is reassuring for *older* codex builds — an unknown key is ignored rather than fatal, so this cannot brick startup on a version predating the capability. I did not test anything below 0.147.0.

The forward direction is the one worth naming. If a future codex renames or retires `experimentalApi` while keeping `excludeTurns` gated, `initialize` still succeeds (unknown key, silently ignored) and resume starts failing again with the exact `-32600` my round-1 block described — which, as I traced in round 1, classifies through `_classify_jsonrpc_error` to `ProviderFailureError`, is neither `_native_resume_is_unavailable` nor `_resume_error_is_transient`, and so lands in the unclassified branch and falls back only after 3 attempts. That is a pre-existing property of the resume path, not something this PR introduces, but this PR is what makes the repo depend on an experimental gate staying stable. Worth a follow-up ticket rather than a change here.

**2. Test collection is fine.** `ci.yml:76` runs bare `pytest` with no path allowlist, so `tests/test_codex_resume.py` genuinely executes in CI on both py3.11 and py3.12. Worth confirming explicitly, because a new test file that runs nowhere is a failure mode I have hit on other repos in this org.

No blockers. Nothing here is a merge recommendation — that call belongs to the operator.

PR #315, opened 2026-09-04
Prepare Agent 2.0.4a1 TestPyPI release

Glimmer0x:
## Summary

- set the package and lockfile version to the prerelease `2.0.4a1`
- document every user-visible change merged since v2.0.3: #305, #308, #309, #310, #313, and #314
- restore README stable-install guidance to production PyPI latest `2.0.3`
- explicitly document that this candidate is for TestPyPI staging only

## Important TestPyPI state

TestPyPI already contains an earlier artifact numbered `2.0.4` from the September 1 dry run. The new `2.0.4a1` artifact is still unique and publishable, but testers must install it with an explicit `==2.0.4a1` pin because package resolution otherwise prefers `2.0.4`. Production PyPI remains at `2.0.3`; no production release or tag has been created.

## Validation

- `git diff --check`
- `uv lock`
- `env -u CODEX_HOME uv run --extra dev pre-commit run --all-files`
- built sdist + wheel; wheel metadata is `Version: 2.0.4a1`
- installed the wheel in a clean Python 3.11 environment and verified `puffo-agent version` reports `2.0.4a1`
- first-parent history from v2.0.3 checked against merged PR metadata

After exact-head review and green CI, merge this PR and manually dispatch `publish-testpypi.yml` from `main`. Do not create a GitHub Release: that would trigger the production-only workflow.

### DIFF
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -15,7 +15,7 @@
 #
 # Verify after a run with:
 #   pip install -i https://test.pypi.org/simple/ \
-#     --extra-index-url https://pypi.org/simple/ puffo-agent
+#     --extra-index-url https://pypi.org/simple/ puffo-agent==2.0.4a1
 # (the extra index lets test-pypi resolve real deps from real PyPI).
 
 name: Publish to TestPyPI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-## [2.0.4a1] - 2026-09-03
+## 2.0.4a1 - 2026-09-03
 
 > Pre-release published to TestPyPI for staging validation only. It is not the
 > stable `2.0.4` release. Because TestPyPI already contains an earlier build
@@ -5028,8 +5028,7 @@ First public PyPI release.
   future server-side regression that echoes the same cursor back
   bails instead of spinning.
 
-[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.4a1...HEAD
-[2.0.4a1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4a1
+[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.3...HEAD
 [2.0.3]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.3
 [2.0.2]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.2
 [2.0.1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-## [2.0.4] - 2026-09-01
+## [2.0.4a1] - 2026-09-03
+
+> Pre-release published to TestPyPI for staging validation only. It is not the
+> stable `2.0.4` release. Because TestPyPI already contains an earlier build
+> numbered `2.0.4`, install this candidate with an explicit `==2.0.4a1` pin.
 
 ### Added
 
@@ -15,6 +19,16 @@ this project adheres to [Semantic Versioning](https://semver.org/).
   exact models before provisioning identity state, and discovers standard
   per-user OpenCode installations even under a narrow daemon `PATH`. Existing
   ACP-over-OpenCode configurations continue to load unchanged. (#305)
+- **Generic Monid paid-data tools.** Native agents can prepare arbitrary
+  allowlisted provider capabilities for free, inspect their input schema and
+  quoted price, then execute a budget-gated spend without holding vendor keys
+  or funds. Results carry provenance, failures require honest fallback
+  labeling, and retry keys prevent accidental duplicate charges. (#308)
+- **Cross-platform daemon autostart.** `puffo-agent autostart
+  enable|disable|status` installs per-user launchd, systemd, or Windows startup
+  registration. Machine linking enables it by default, with an explicit
+  opt-out, safe persisted environments, and actionable status or failure
+  reporting. (#309)
 
 ### Fixed
 
@@ -24,6 +38,19 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 - **OpenCode tool activity reaches the UI.** Tool frames now project both
   start and completion status, and unavailable selected models are reported
   separately from missing provider authentication. (#305)
+- **Sleep/wake transport recovery.** A background daemon now heals an event
+  loop executor that died during macOS sleep before sending new traffic,
+  avoids replaying requests with ambiguous outcomes, and reports sustained
+  reconnect failure as `server_unreachable`. The tray also requests a
+  best-effort App Nap exemption. (#310)
+- **Remote creates report real startup readiness.** Docker availability is
+  checked before identity materialization, create commands wait for durable
+  runtime readiness, slow startup does not block unrelated machine commands,
+  and terminal acknowledgements survive reconnects. (#313)
+- **Large Codex sessions resume without replaying full history.** Codex resume
+  now declares the required experimental API capability and requests a
+  metadata-only response, preserving the native thread and provider-side
+  context without overflowing the Agent's stdout frame limit. (#314)
 
 ## [2.0.3] - 2026-08-27
 
@@ -5001,8 +5028,8 @@ First public PyPI release.
   future server-side regression that echoes the same cursor back
   bails instead of spinning.
 
-[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.4...HEAD
-[2.0.4]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4
+[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.4a1...HEAD
+[2.0.4a1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4a1
 [2.0.3]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.3
 [2.0.2]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.2
 [2.0.1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.1
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Python:
 ```bash
 uv tool install puffo-agent
 # pin to a specific version:
-uv tool install puffo-agent==2.0.4
+uv tool install puffo-agent==2.0.3
 ```
 
 Otherwise use pip:
@@ -70,10 +70,10 @@ Otherwise use pip:
 ```bash
 pip install puffo-agent
 # pin to a specific version:
-pip install puffo-agent==2.0.4
+pip install puffo-agent==2.0.3
 ```
 
-Agent Foundation `2.0.4` is the current stable release. It upgrades supported
+Agent Foundation `2.0.3` is the current stable release. It upgrades supported
 1.2 state in place while preserving Agent identity, keys, profile, memory,
 workspace, message history, and logical session continuity.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ build-backend = "setuptools.build_meta"
 # *same* env without colliding. CLI binary stays `puffo-agent`; home
 # dir stays `~/.puffo-agent/`.
 name = "puffo-agent"
-version = "2.0.4"
+version = "2.0.4a1"
 description = "Run AI bots on Puffo.ai — local daemon that supervises many bot accounts and handles their LLM loops."
 readme = "README.md"
 license = { file = "LICENSE" }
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-## [2.0.4] - 2026-09-01
+## [2.0.4a1] - 2026-09-03
+
+> Pre-release published to TestPyPI for staging validation only. It is not the
+> stable `2.0.4` release. Because TestPyPI already contains an earlier build
+> numbered `2.0.4`, install this candidate with an explicit `==2.0.4a1` pin.
```
hellohanchen, line 13: Verified against test.pypi.org directly: `2.0.4` is present and `2.0.4a1` is not, so both halves of this note are true and the upload will not collide.

The reasoning is actually stronger than stated. TestPyPI's `2.0.4` is a *final* release, and pip excludes pre-releases by default -- so without the pin `2.0.4a1` is not even a candidate, regardless of ordering. `==2.0.4a1` is the only form that resolves it without `--pre`. Good note.

See N1 in the review body: `.github/workflows/publish-testpypi.yml` still documents the pin-less verification command, which under exactly this state installs `2.0.4` instead.

CHANGELOG.md
```diff
@@ -5001,8 +5028,8 @@ First public PyPI release.
   future server-side regression that echoes the same cursor back
   bails instead of spinning.
 
-[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.4...HEAD
-[2.0.4]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4
+[Unreleased]: https://github.com/puffo-ai/puffo-agent/compare/v2.0.4a1...HEAD
+[2.0.4a1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4a1
```
hellohanchen: **N2 (non-blocking):** this follows the file's convention, but `publish-pypi.yml` triggers on `release: [published]` -- so creating the GitHub Release that makes this link resolve would publish `2.0.4a1` to **production** PyPI, contradicting the TestPyPI-only scope. A bare tag is safe; only a published Release fires it. Consider noting "tag only, no GitHub Release" in the release checklist, or holding this link entry until stable `2.0.4`.

README.md
```diff
@@ -62,18 +62,18 @@ Python:
...
 
 ```bash
 pip install puffo-agent
 # pin to a specific version:
-pip install puffo-agent==2.0.4
+pip install puffo-agent==2.0.3
 ```
 
-Agent Foundation `2.0.4` is the current stable release. It upgrades supported
+Agent Foundation `2.0.3` is the current stable release. It upgrades supported
```
hellohanchen, line 76: Correct: pypi.org's latest for `puffo-agent` is `2.0.3`, which I queried directly. Reverting the stable-release line from `2.0.4` (never published to production) back to `2.0.3` makes the README true again.

pyproject.toml
```diff
@@ -11,7 +11,7 @@ build-backend = "setuptools.build_meta"
 # *same* env without colliding. CLI binary stays `puffo-agent`; home
 # dir stays `~/.puffo-agent/`.
 name = "puffo-agent"
-version = "2.0.4"
+version = "2.0.4a1"
```
hellohanchen, line 14: `2.0.4a1` is already PEP 440-normalized, so the built artifact names match the pin the changelog tells testers to use. `uv.lock` is updated in step, and no other `2.0.4` reference remains anywhere in the tree outside the CHANGELOG's one deliberate explanatory mention.

hellohanchen (commented):
## Round 1 — no blockers, two non-blocking findings

Reviewed exactly `6c1fe5ddb4c53d34ce9bebe2cbced39bdfbf3fc0` (the superseded `09d10dbf` is not an ancestor, so this is a fresh review, not a delta). Merge-base is `4ca42c52`, which is `origin/main`. All 7 checks terminal SUCCESS on this exact head, re-queried immediately before submitting. 4 files, +35/−8.

### What I verified rather than took on trust

**Every external version claim is true.** I queried both indexes directly:

| claim | check | result |
|---|---|---|
| `2.0.3` is the current stable release (README) | pypi.org latest | `2.0.3` ✓ |
| TestPyPI already contains a build numbered `2.0.4` | test.pypi.org releases | `2.0.4` present ✓ |
| `2.0.4a1` is free to upload | test.pypi.org releases | not present ✓ |

The pin advice is also correct for the right reason, and it is stronger than the note implies: TestPyPI's existing `2.0.4` is a *final* release, and pip excludes pre-releases by default, so `==2.0.4a1` is the only form that resolves this candidate without `--pre`.

**Completeness.** Merge commits in `v2.0.3..origin/main` reference `#305 #307 #308 #309 #310 #313 #314`. Six are listed; the only omission is `#307`, which is the previous release-prep PR and correctly not a user-facing entry. Nothing merged in the range is missing. All six listed PRs are genuinely merged into `main` — I resolved each `merge_commit_sha` and confirmed it is an ancestor of `4ca42c52` (in fact `#314`'s merge *is* `4ca42c52`), so there is no entry describing unmerged work.

**No exaggerations found.** I checked each entry's falsifiable claims against the code on `main`, not against the PR descriptions:

- `#309` — `puffo-agent autostart enable|disable|status` exists verbatim in `portal/cli_parser.py`, with launchd/systemd/Windows registration in `portal/autostart.py`. ✓
- `#310` — `server_unreachable` is a real health state (`portal/state.py`, `portal/worker.py:418`); the App Nap exemption is `NSProcessInfo.beginActivityWithOptions:` in `portal/ui/tray.py`; the "never replayed" wording matches the ambiguous-outcome handling the diff added. ✓
- `#313` — `preflight.py` gained a `RUNTIME_CLI_DOCKER` branch (`resolve_docker_bin` + `container_state`), and `provision.py` awaits `preflight` **before** `materialize`, so "Docker availability is checked before identity materialization" is literally accurate, not marketing. ✓
- `#314` — accurate. I verified this one end-to-end during its own review: the capability is required, and resume returns thread metadata with `turns` emptied while the native thread is preserved. ✓
- `#308` — `monid_prepare` is documented FREE and `monid_spend` PAID, with provenance stamping and retry keys, matching the entry. ✓
- `#305` — carried over unchanged, as intended. ✓

**Classification.** `#308` and `#309` add new surfaces → **Added**. `#310`, `#313`, `#314` change existing behaviour to stop a failure mode → **Fixed**. All five read correctly to me.

**The date is right, and I checked it against the repo's own convention rather than against the wall clock.** Previous prep commits set the heading to the *author-local* date (`prepare 2.0.3` authored `2026-08-27 -07:00` → `## [2.0.3] - 2026-08-27`; `prepare 2.0.4` authored `2026-09-01 -07:00` → `- 2026-09-01`). This head is authored `2026-09-03T20:01:09-07:00`, so `2026-09-03` matches. It is `2026-09-04` in UTC, which is why I checked — but the convention is local, and this follows it.

**No stale references.** Outside the CHANGELOG's one deliberate explanatory mention, no `2.0.4` remains anywhere in the tree. `pyproject.toml` and `uv.lock` agree on `2.0.4a1`.

### N1 (non-blocking): the repo's own TestPyPI verification command cannot install this candidate

`.github/workflows/publish-testpypi.yml` documents, at the top of the file, how to verify a run:

```
pip install -i https://test.pypi.org/simple/ \
  --extra-index-url https://pypi.org/simple/ puffo-agent
```

Run against the state this PR creates, that command installs **`2.0.4`** — the older, unrelated TestPyPI build this PR's own note warns about — not `2.0.4a1`. Two independent reasons, either sufficient: TestPyPI's `2.0.4` is a final release and outranks anything the pin-less resolver will consider, and pip excludes pre-releases by default so `2.0.4a1` is not even a candidate. The `--extra-index-url` makes it worse rather than better, since production PyPI contributes a final `2.0.3` that also outranks a pre-release.

So whoever performs the staging validation this PR exists to enable, by following the instructions in the repo, validates the wrong artifact and reports success. The CHANGELOG note gets this exactly right; the workflow comment contradicts it. The file is untouched by this PR, but this PR is what makes the comment wrong in practice — publishing a pre-release numbered *below* an existing TestPyPI final is precisely the case it does not handle. Updating that comment to `pip install -i … 'puffo-agent==2.0.4a1'` belongs in this PR, since it is the same release decision.

### N2 (non-blocking): the changelog link invites the action that would publish this to production PyPI

`CHANGELOG.md:5032` adds `[2.0.4a1]: https://github.com/puffo-ai/puffo-agent/releases/tag/v2.0.4a1`, following the file's convention for every prior version. But `.github/workflows/publish-pypi.yml` triggers on `release: [published]` — so creating that GitHub Release, which is the normal way to make that link resolve, publishes `2.0.4a1` to **production** PyPI, contradicting the "TestPyPI-only" scope.

A bare git tag is safe; only a published Release fires it. The blast radius is contained — `pip install puffo-agent` will not select a pre-release, so ordinary users are unaffected — but it would burn the version on the real index and undo the separation this PR is establishing. Worth an explicit "tag only, do not publish a GitHub Release for this one" line in the release checklist, or leaving the link entry out until the stable `2.0.4` ships.

Neither finding is a blocker. Nothing here is a merge recommendation — that call belongs to the operator.

.github/workflows/publish-testpypi.yml
```diff
@@ -15,7 +15,7 @@
 #
 # Verify after a run with:
 #   pip install -i https://test.pypi.org/simple/ \
-#     --extra-index-url https://pypi.org/simple/ puffo-agent
+#     --extra-index-url https://pypi.org/simple/ puffo-agent==2.0.4a1
```
hellohanchen, line 18: Confirmed by execution against the live indexes with these exact flags, not by reasoning about pip:

- the previous unpinned form downloads `puffo_agent-2.0.4-py3-none-any.whl` -- the stale build, exactly as filed;
- an exact `==` pin to an existing TestPyPI pre-release (`==2.0.0a24`) downloads `puffo_agent-2.0.0a24-py3-none-any.whl` **without** `--pre`.

So the pinned form is demonstrably the right mechanism. (`pip index versions` under the same flags lists no pre-release at all, which is the same fact seen from the other side.) Downloads went to a temp dir and were removed.

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-## [2.0.4] - 2026-09-01
+## 2.0.4a1 - 2026-09-03
```
hellohanchen, line 9: N2 resolved -- plain-text heading, link definition removed, and `[Unreleased]` now compares from `v2.0.3`, a tag that actually exists. Grepped the tree: no dangling `[2.0.4a1]` reference remains.

**Non-blocking note for the next release-prep:** every other pre-release here carries a tag link (`[0.10.0a1]`, `[2.0.0a1]`, `[2.0.0a2]`), so this is now the file's one non-conforming heading -- and that non-conformity is load-bearing, since it is what stops a Release being created just to make the link resolve, which `publish-pypi.yml` would turn into a production publish. Worth one line in the release checklist or an HTML comment here, so a later tidy-up doesn't restore it.

hellohanchen (commented):
## Round 2 — both findings resolved, no blockers

Reviewed exactly `072f1d0d7ea71ba8bd05cdbb19e7675c70f84355`. `6c1fe5dd` is confirmed an ancestor, so this is a clean delta: **2 files, +3/−4**. Everything I verified in round 1 — completeness against `v2.0.3..main`, the accuracy and classification of all six entries, the author-local date convention, the pyproject/uv.lock agreement — is untouched and still holds.

### N1 resolved, and I confirmed it by execution rather than by reasoning

The workflow comment now pins `puffo-agent==2.0.4a1`. Rather than argue pip semantics, I ran both forms against the live indexes with the exact flags the file documents:

| command | result |
|---|---|
| old, unpinned | downloads **`puffo_agent-2.0.4-py3-none-any.whl`** |
| exact pin to an existing TestPyPI pre-release (`==2.0.0a24`), **no `--pre`** | downloads **`puffo_agent-2.0.0a24-py3-none-any.whl`** |

So the defect was real and reproducible — the old runbook really did fetch the stale `2.0.4` build — and the pinned form is demonstrably the correct mechanism: an exact `==` pin to a pre-release resolves without `--pre`. `pip index versions` under the same flags does not list any pre-release at all, which is the same fact from the other side. Downloads were to a temporary directory and have been removed.

### N2 resolved

`[2.0.4a1]: …/releases/tag/v2.0.4a1` is gone, the heading is plain text (`## 2.0.4a1 - 2026-09-03`), and `[Unreleased]` now compares from `v2.0.3` — a tag that actually exists, unlike the `v2.0.4a1` it previously pointed at. I grepped the whole tree at this head: the only remaining `2.0.4a1` occurrences are the pinned workflow comment, the heading, the pin instruction in the note, `pyproject.toml`, and `uv.lock`. No dangling link reference is left behind, and no tag or Release is implied anywhere.

### One non-blocking note for whoever does the next release-prep

Every other pre-release in this file carries a tag link — `[0.10.0a1]`, `[2.0.0a1]`, `[2.0.0a2]` all have definitions — so this plain-text heading is now the only entry in the changelog that departs from the file's convention. That departure is load-bearing: it is what keeps `publish-pypi.yml`'s `release: [published]` trigger from firing on a Release created merely to make the link resolve. A later tidy-up pass that "fixes" the inconsistent heading would silently restore the production-publish path.

The heading itself can't carry that explanation without becoming odd to read, so it's worth one line wherever the release checklist lives, or a short HTML comment above the heading. Not a change I'd hold this PR for.

**No blockers.** All 7 checks are terminal SUCCESS on this exact head, re-queried immediately before submitting. Nothing here is a merge recommendation — that call belongs to the operator.

PR #316, opened 2026-09-04
Expose OpenCode model reasoning variants

Glimmer0x:
## Summary

- discover model-specific reasoning variants from OpenCode native verbose model metadata
- publish only supported Puffo inference levels, mapping native none to off
- forward selected inference levels to `opencode run` via `--variant`
- fall back to the ordinary model list when older OpenCode versions reject `models --verbose`
- recover from malformed verbose metadata without dropping runnable models or inventing phantom IDs

## Verification

- focused OpenCode and catalog tests: 28 passed
- full suite: 3,902 passed, 2 skipped
- Ruff and `git diff --check`: passed
- real OpenCode 1.18.16 catalog: 36 models discovered, 23 with published inference levels

No live model turn was sent; command construction and native catalog metadata were verified locally.

### DIFF
--- a/src/puffo_agent/agent/harness/runtime/local_runtime.py
+++ b/src/puffo_agent/agent/harness/runtime/local_runtime.py
@@ -20,6 +20,7 @@ from typing import Any, Protocol
 from ....macos.keychain import is_macos
 from ....mcp.config import (
     INFERENCE_LEVELS,
+    OPENCODE_INFERENCE_LEVELS,
     default_python_executable,
     puffo_core_mcp_env,
     write_cli_mcp_config,
@@ -454,6 +455,13 @@ class LocalRuntimePreparer:
             and "--thinking" not in launch_args
         ):
             launch_args.extend(("--thinking", inference_level))
+        if (
+            self.harness_name == "opencode"
+            and inference_level in OPENCODE_INFERENCE_LEVELS
+            and "--variant" not in launch_args
+        ):
+            native_variant = "none" if inference_level == "off" else inference_level
+            launch_args.extend(("--variant", native_variant))
         return executable, launch_args
 
     def _prepare_executable_configuration(
--- a/src/puffo_agent/agent/model_catalog.py
+++ b/src/puffo_agent/agent/model_catalog.py
@@ -206,18 +206,30 @@ def _opencode_models(*, fetch: bool) -> tuple[ModelOption, ...]:
     if cached is not None or not fetch:
         return cached if cached is not None else _stale_models("opencode")
     from .cli_bin import resolve_opencode_bin
-    from .opencode_auth import OpenCodeProbeError, list_opencode_models
+    from .opencode_auth import OpenCodeProbeError, list_opencode_model_catalog
 
     executable = resolve_opencode_bin()
     if not executable:
         return _store_models("opencode", ())
     try:
-        model_ids = list_opencode_models(executable)
+        models = list_opencode_model_catalog(executable)
     except OpenCodeProbeError:
         return _stale_models("opencode")
+    from ..mcp.config import OPENCODE_INFERENCE_LEVELS
+
     options = tuple(
-        ModelOption(model_id, model_id)
-        for model_id in model_ids
+        ModelOption(
+            model.id,
+            model.id,
+            supported_inference_levels=tuple(
+                level for level in OPENCODE_INFERENCE_LEVELS
+                if (
+                    level in model.variants
+                    or (level == "off" and "none" in model.variants)
+                )
+            ),
+        )
+        for model in models
     )
     return _store_models("opencode", options)
 
--- a/src/puffo_agent/agent/model_catalog.py
+++ b/src/puffo_agent/agent/model_catalog.py
@@ -27,7 +27,9 @@ class ModelOption:
     id: str  # the ``--model`` value; "" means the daemon default
     label: str  # combo-box display text
     is_alias: bool = False
-    supported_inference_levels: tuple[str, ...] = ()
+    # None means the catalog makes no model-specific claim. An empty tuple is
+    # an explicit claim that this model exposes no selectable inference level.
+    supported_inference_levels: tuple[str, ...] | None = None
 
 
 _DAEMON_DEFAULT = ModelOption("", "(daemon default)")
--- a/src/puffo_agent/agent/opencode_auth.py
+++ b/src/puffo_agent/agent/opencode_auth.py
@@ -73,7 +73,33 @@ def _run_opencode_models(
 
 
 def _is_model_id(value: str) -> bool:
-    return "/" in value and not any(char.isspace() for char in value)
+    return (
+        "/" in value
+        and not any(char.isspace() or char in '{}[],"' for char in value)
+    )
+
+
+def _next_verbose_model_offset(output: str, offset: int) -> int:
+    """Recover at the next bare model-id line after malformed metadata."""
+    length = len(output)
+    line_start = offset
+    while line_start < length:
+        line_end = output.find("\n", line_start)
+        if line_end < 0:
+            line_end = length
+        raw_line = output[line_start:line_end]
+        candidate = raw_line.strip()
+        metadata_offset = line_end + 1
+        while metadata_offset < length and output[metadata_offset].isspace():
+            metadata_offset += 1
+        if (
+            raw_line == candidate
+            and _is_model_id(candidate)
+            and (metadata_offset == length or output[metadata_offset] == "{")
+        ):
+            return line_start
+        line_start = line_end + 1
+    return length
 
 
 def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
@@ -106,6 +132,7 @@ def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
                 metadata, offset = decoder.raw_decode(output, json_offset)
             except json.JSONDecodeError:
                 metadata = {}
+                offset = _next_verbose_model_offset(output, offset)
 
         variants = metadata.get("variants") if isinstance(metadata, dict) else {}
         variant_names = tuple(
@@ -149,12 +176,22 @@ def list_opencode_model_catalog(
     timeout_seconds: float = 5.0,
 ) -> tuple[OpenCodeModel, ...]:
     """Return visible models with their model-specific native variants."""
-    stdout = _run_opencode_models(
-        executable,
-        provider=provider,
-        verbose=True,
-        timeout_seconds=timeout_seconds,
-    )
+    try:
+        stdout = _run_opencode_models(
+            executable,
+            provider=provider,
+            verbose=True,
+            timeout_seconds=timeout_seconds,
+        )
+    except OpenCodeProbeError:
+        return tuple(
+            OpenCodeModel(model_id)
+            for model_id in list_opencode_models(
+                executable,
+                provider=provider,
+                timeout_seconds=timeout_seconds,
+            )
+        )
     return _parse_verbose_models(stdout)
 
 
--- a/src/puffo_agent/agent/opencode_auth.py
+++ b/src/puffo_agent/agent/opencode_auth.py
@@ -79,29 +79,6 @@ def _is_model_id(value: str) -> bool:
     )
 
 
-def _next_verbose_model_offset(output: str, offset: int) -> int:
-    """Recover at the next bare model-id line after malformed metadata."""
-    length = len(output)
-    line_start = offset
-    while line_start < length:
-        line_end = output.find("\n", line_start)
-        if line_end < 0:
-            line_end = length
-        raw_line = output[line_start:line_end]
-        candidate = raw_line.strip()
-        metadata_offset = line_end + 1
-        while metadata_offset < length and output[metadata_offset].isspace():
-            metadata_offset += 1
-        if (
-            raw_line == candidate
-            and _is_model_id(candidate)
-            and (metadata_offset == length or output[metadata_offset] == "{")
-        ):
-            return line_start
-        line_start = line_end + 1
-    return length
-
-
 def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
     """Parse ``models --verbose``'s repeated ``id`` + JSON blocks.
 
@@ -118,9 +95,10 @@ def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
         line_end = output.find("\n", offset)
         if line_end < 0:
             line_end = length
-        model_id = output[offset:line_end].strip()
+        raw_line = output[offset:line_end]
+        model_id = raw_line.strip()
         offset = line_end + 1
-        if not _is_model_id(model_id) or model_id in seen:
+        if raw_line != model_id or not _is_model_id(model_id) or model_id in seen:
             continue
 
         metadata: object = {}
@@ -132,7 +110,10 @@ def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
                 metadata, offset = decoder.raw_decode(output, json_offset)
             except json.JSONDecodeError:
                 metadata = {}
-                offset = _next_verbose_model_offset(output, offset)
+                # Continue line by line from the malformed block. Bare model-id
+                # lines remain authoritative even when they have no JSON block;
+                # punctuation and indentation keep JSON fragments from becoming
+                # phantom models.
 
         variants = metadata.get("variants") if isinstance(metadata, dict) else {}
         variant_names = tuple(
--- a/src/puffo_agent/agent/opencode_auth.py
+++ b/src/puffo_agent/agent/opencode_auth.py
@@ -10,8 +10,10 @@ runtime can actually launch.
 
 from __future__ import annotations
 
+import json
 import re
 import subprocess
+from dataclasses import dataclass
 from typing import Literal
 
 from .harness.support.child_env import build_child_environment
@@ -29,16 +31,26 @@ OpenCodeModelStatus = Literal[
 _ANSI_ESCAPE = re.compile(r"\x1b\[[0-9;]*m")
 
 
-def list_opencode_models(
+@dataclass(frozen=True, slots=True)
+class OpenCodeModel:
+    """One model and the native variants advertised by OpenCode."""
+
+    id: str
+    variants: tuple[str, ...] = ()
+
+
+def _run_opencode_models(
     executable: str,
     *,
-    provider: str = "",
-    timeout_seconds: float = 5.0,
-) -> tuple[str, ...]:
-    """Return model IDs visible to OpenCode's current native credential view."""
+    provider: str,
+    verbose: bool,
+    timeout_seconds: float,
+) -> str:
     command = [executable, "models"]
     if provider:
         command.append(provider)
+    if verbose:
+        command.append("--verbose")
     try:
         completed = subprocess.run(
             command,
@@ -51,22 +63,99 @@ def list_opencode_models(
     except (OSError, subprocess.TimeoutExpired) as exc:
         raise OpenCodeProbeError("OpenCode model check could not complete") from exc
 
-    diagnostic = _ANSI_ESCAPE.sub("", completed.stdout + completed.stderr)
+    stdout = _ANSI_ESCAPE.sub("", completed.stdout)
+    diagnostic = stdout + _ANSI_ESCAPE.sub("", completed.stderr)
     if completed.returncode != 0:
         if provider and "Provider not found:" in diagnostic:
-            return ()
+            return ""
         raise OpenCodeProbeError("OpenCode model check failed")
+    return stdout
 
+
+def _is_model_id(value: str) -> bool:
+    return "/" in value and not any(char.isspace() for char in value)
+
+
+def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
+    """Parse ``models --verbose``'s repeated ``id`` + JSON blocks.
+
+    The model-id line is the launch authority. JSON is used only for optional
+    variant metadata, so a malformed/missing block degrades to no variants
+    without inventing support or dropping a runnable model.
+    """
+    decoder = json.JSONDecoder()
+    models: list[OpenCodeModel] = []
+    seen: set[str] = set()
+    offset = 0
+    length = len(output)
+    while offset < length:
+        line_end = output.find("\n", offset)
+        if line_end < 0:
+            line_end = length
+        model_id = output[offset:line_end].strip()
+        offset = line_end + 1
+        if not _is_model_id(model_id) or model_id in seen:
+            continue
+
+        metadata: object = {}
+        json_offset = offset
+        while json_offset < length and output[json_offset].isspace():
+            json_offset += 1
+        if json_offset < length and output[json_offset] == "{":
+            try:
+                metadata, offset = decoder.raw_decode(output, json_offset)
+            except json.JSONDecodeError:
+                metadata = {}
+
+        variants = metadata.get("variants") if isinstance(metadata, dict) else {}
+        variant_names = tuple(
+            key for key in variants
+            if isinstance(key, str) and key
+        ) if isinstance(variants, dict) else ()
+        seen.add(model_id)
+        models.append(OpenCodeModel(model_id, variant_names))
+    return tuple(models)
+
+
+def list_opencode_models(
+    executable: str,
+    *,
+    provider: str = "",
+    timeout_seconds: float = 5.0,
+) -> tuple[str, ...]:
+    """Return model IDs visible to OpenCode's current native credential view."""
+    stdout = _run_opencode_models(
+        executable,
+        provider=provider,
+        verbose=False,
+        timeout_seconds=timeout_seconds,
+    )
     models: list[str] = []
     seen: set[str] = set()
-    for line in completed.stdout.splitlines():
+    for line in stdout.splitlines():
         model = line.strip()
-        if "/" not in model or any(char.isspace() for char in model):
+        if not _is_model_id(model):
             continue
         if model not in seen:
             seen.add(model)
             models.append(model)
     return tuple(models)
+
+
+def list_opencode_model_catalog(
+    executable: str,
+    *,
+    provider: str = "",
+    timeout_seconds: float = 5.0,
+) -> tuple[OpenCodeModel, ...]:
+    """Return visible models with their model-specific native variants."""
+    stdout = _run_opencode_models(
+        executable,
+        provider=provider,
+        verbose=True,
+        timeout_seconds=timeout_seconds,
+    )
+    return _parse_verbose_models(stdout)
 
 
 def opencode_model_status(executable: str, model: str) -> OpenCodeModelStatus:
--- a/tests/test_opencode_auth.py
+++ b/tests/test_opencode_auth.py
@@ -89,19 +89,51 @@ def test_verbose_model_probe_parses_native_variants(monkeypatch):
     assert "OPENAI_API_KEY" not in seen["kwargs"]["env"]
 
 
-def test_verbose_probe_keeps_model_when_metadata_is_malformed(monkeypatch):
+def test_verbose_probe_recovers_after_malformed_metadata_without_phantom_models(
+    monkeypatch,
+):
     monkeypatch.setattr(
         subprocess,
         "run",
         lambda *args, **kwargs: _completed(
             code=0,
-            stdout="opencode/big-pickle\n{not-json}\n",
+            stdout=(
+                "opencode/big-pickle\n"
+                '{"id":"a","api":{"npm":"@ai-sdk/openai-compatible"\n'
+                "opencode/second-model\n"
+                '{"variants":{"low":{}}}\n'
+            ),
         ),
     )
 
     assert list_opencode_model_catalog("/opt/bin/opencode") == (
         OpenCodeModel("opencode/big-pickle"),
+        OpenCodeModel("opencode/second-model", ("low",)),
     )
+
+
+def test_verbose_probe_falls_back_to_non_verbose_models(monkeypatch):
+    seen = []
+
+    def fake_run(command, **kwargs):
+        seen.append(command)
+        if "--verbose" in command:
+            return _completed(code=1, stderr="Unknown option: --verbose")
+        return _completed(
+            code=0,
+            stdout="opencode/big-pickle\ndeepseek/deepseek-v4-pro\n",
+        )
+
+    monkeypatch.setattr(subprocess, "run", fake_run)
+
+    assert list_opencode_model_catalog("/opt/bin/opencode") == (
+        OpenCodeModel("opencode/big-pickle"),
+        OpenCodeModel("deepseek/deepseek-v4-pro"),
+    )
+    assert seen == [
+        ["/opt/bin/opencode", "models", "--verbose"],
+        ["/opt/bin/opencode", "models"],
+    ]
 
 
 def test_selected_model_uses_provider_filtered_native_catalog(monkeypatch):
--- a/tests/test_opencode_auth.py
+++ b/tests/test_opencode_auth.py
@@ -7,7 +7,9 @@ import subprocess
 import pytest
 
 from puffo_agent.agent.opencode_auth import (
+    OpenCodeModel,
     OpenCodeProbeError,
+    list_opencode_model_catalog,
     list_opencode_models,
     opencode_model_is_available,
     opencode_model_status,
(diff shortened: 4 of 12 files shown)


src/puffo_agent/agent/opencode_auth.py
```diff
@@ -51,24 +63,101 @@ def list_opencode_models(
...
+
+        metadata: object = {}
+        json_offset = offset
+        while json_offset < length and output[json_offset].isspace():
+            json_offset += 1
+        if json_offset < length and output[json_offset] == "{":
+            try:
+                metadata, offset = decoder.raw_decode(output, json_offset)
+            except json.JSONDecodeError:
+                metadata = {}
```
hellohanchen, line 112: **B2 (robustness).** `offset` is not advanced here. On a `JSONDecodeError` the metadata is dropped, which is correct, but the scan then resumes *inside* the malformed block rather than after it — and `_is_model_id` accepts any token with a `/` and no whitespace, so JSON fragments become model ids.

Against the real parser:

```
truncated JSON  -> (OpenCodeModel('opencode/big-pickle'),
                    OpenCodeModel('"npm":"@ai-sdk/openai-compatible"'),   <- phantom
                    OpenCodeModel('opencode/second-model', ('low',)))
```

The docstring six lines up promises the opposite: "degrades to no variants without inventing support or dropping a runnable model". The intent is right, the implementation doesn't hold it. Advancing past the block — scanning to the next line boundary, or letting `raw_decode` tell you where it stopped — is where the fix lives, but the choice is yours.

src/puffo_agent/agent/opencode_auth.py
```diff
@@ -51,24 +63,101 @@ def list_opencode_models(
...
+    executable: str,
+    *,
+    provider: str = "",
+    timeout_seconds: float = 5.0,
+) -> tuple[OpenCodeModel, ...]:
+    """Return visible models with their model-specific native variants."""
+    stdout = _run_opencode_models(
+        executable,
+        provider=provider,
+        verbose=True,
```
hellohanchen: **B1 (blocker).** `verbose=True` is unconditional: no version gate, no fallback to the non-verbose path. `_run_opencode_models` raises `OpenCodeProbeError` on any non-zero exit, and older binaries exit 1 for `models --verbose` (they print help instead).

Measured on real binaries — `1.0.0` and `0.5.29` both:

```
models --verbose             -> rc=1
list_opencode_models (pre-PR) -> 1 model
list_opencode_model_catalog   -> OpenCodeProbeError
_opencode_models(fetch=True)  -> ()
opencode_has_accessible_models() -> True
```

The last two lines disagree. Readiness reports the harness usable while the catalog is empty, because `cli_bin.py:167` still uses the non-verbose probe. That is a regression on the same binary against the pre-PR path, and it surfaces to the operator as "no models on this account" rather than "your opencode is too old".

src/puffo_agent/agent/model_catalog.py
```diff
@@ -206,18 +206,30 @@ def _opencode_models(*, fetch: bool) -> tuple[ModelOption, ...]:
...
+    from .opencode_auth import OpenCodeProbeError, list_opencode_model_catalog
 
     executable = resolve_opencode_bin()
     if not executable:
         return _store_models("opencode", ())
     try:
-        model_ids = list_opencode_models(executable)
+        models = list_opencode_model_catalog(executable)
     except OpenCodeProbeError:
         return _stale_models("opencode")
```
hellohanchen, line 219: **B1, second half.** This is where the probe failure becomes invisible. `_stale_models` returns the cached tuple *or* `()` — so on a fresh process with no cache, an `OpenCodeProbeError` is indistinguishable from "this account genuinely has zero models". Combined with `opencode_has_accessible_models()` still returning `True` (it runs the non-verbose probe), the two halves of readiness disagree with no diagnostic anywhere.

src/puffo_agent/agent/model_catalog.py
```diff
@@ -206,18 +206,30 @@ def _opencode_models(*, fetch: bool) -> tuple[ModelOption, ...]:
...
-        ModelOption(model_id, model_id)
-        for model_id in model_ids
+        ModelOption(
+            model.id,
+            model.id,
+            supported_inference_levels=tuple(
+                level for level in OPENCODE_INFERENCE_LEVELS
+                if (
+                    level in model.variants
+                    or (level == "off" and "none" in model.variants)
```
hellohanchen, line 230: The `none` → `off` half of this mapping is **correct** — I want to record that, because I initially doubted it. No model in the catalog I could reach advertises `none`, but that turned out to be an artefact of which provider I could see (unauthenticated `opencode/*` zen models run through `@ai-sdk/openai-compatible`, whose table for that branch is only `{high, max}`). The shipped binary's own tables carry `none` as a first-class token:

```js
reasoningEffort: Q.enum(["none","minimal","low","medium","high","xhigh","max"])
a0=["low","medium","high"], BN=["none","minimal",...a0,"xhigh"], UN=["none",...a0]
```

No finding here. `max` is genuine too, so all seven `OPENCODE_INFERENCE_LEVELS` values line up.

src/puffo_agent/agent/harness/runtime/local_runtime.py
```diff
@@ -454,6 +455,13 @@ def _resolve_generic_command(self) -> tuple[str, list[str]]:
             and "--thinking" not in launch_args
         ):
             launch_args.extend(("--thinking", inference_level))
+        if (
+            self.harness_name == "opencode"
+            and inference_level in OPENCODE_INFERENCE_LEVELS
+            and "--variant" not in launch_args
+        ):
+            native_variant = "none" if inference_level == "off" else inference_level
+            launch_args.extend(("--variant", native_variant))
```
hellohanchen, line 464: **N1 (note, low severity).** The guard is on the *harness-level* set, so `--variant none` is emitted even for a model whose advertised variants omit it — 4 of the 7 models in the live catalog advertise no variants at all.

It is a no-op rather than a failure, and I checked that with a control rather than assuming it:

```
opencode run --variant none               -> rc=0, normal completion
opencode run --variant totallybogusvariant -> rc=0, normal completion
```

The bogus control is the point: `rc=0` on `none` proves nothing by itself. Unknown variants are silently discarded, so a user asking for `off` on such a model gets the model's default effort with no diagnostic. Worth a decision on whether that should be silent.

tests/test_opencode_auth.py
```diff
@@ -57,6 +59,51 @@ def test_missing_provider_is_a_clean_not_ready_result(monkeypatch):
...
+    assert "OPENAI_API_KEY" not in seen["kwargs"]["env"]
+
+
+def test_verbose_probe_keeps_model_when_metadata_is_malformed(monkeypatch):
+    monkeypatch.setattr(
+        subprocess,
+        "run",
+        lambda *args, **kwargs: _completed(
+            code=0,
+            stdout="opencode/big-pickle\n{not-json}\n",
```
hellohanchen: This input cannot expose the B2 defect it appears to guard. `{not-json}` contains no `/`-bearing whitespace-free token, so when the scan resumes inside the malformed block `_is_model_id` rejects everything it finds and the result looks clean. The test passes for a reason unrelated to the property it names.

Two one-line fixtures that do expose it:

```
"opencode/big-pickle\n{\"id\":\"a\",\"api\":{\"npm\":\"@ai-sdk/openai-compatible\"\nopencode/second-model\n{\"variants\":{\"low\":{}}}\n"
```

which currently yields a phantom `OpenCodeModel('\"npm\":\"@ai-sdk/openai-compatible\"')` between the two real ones.

hellohanchen (commented):
Review of **`7184f45657a739ed2c81500d398d74c40938cd5d`** — the exact head requested. Nothing below is bound to any other head; I re-queried the head immediately before submitting and it had not moved. All seven checks on this SHA are green (`pytest py3.11`, `pytest py3.12`, `pre-commit`, `CodeQL`, `Analyze ×3`).

**Verdict: one blocker (B1). Review only — I have not merged and am not authorising a merge.**

I ran everything against real `opencode` binaries rather than against the PR's own tests: `1.18.27` (current), and `1.0.0` / `0.5.29` (older). I also read the shipped `opencode` binary's own variant tables directly, because the models.dev cache at `~/.cache/opencode/models.json` carries no `variants` key at all and cannot settle the vocabulary question.

---

## B1 — blocker: `--verbose` is unconditional, so an older `opencode` yields an empty catalog while readiness still reports the harness usable

`list_opencode_model_catalog` calls `_run_opencode_models(..., verbose=True)`, and `_run_opencode_models` raises `OpenCodeProbeError` on any non-zero exit. There is no version gate and no fallback to the non-verbose path.

Measured on real binaries:

| binary | `models --verbose` | pre-PR `list_opencode_models` | new `list_opencode_model_catalog` | `_opencode_models(fetch=True)` | `opencode_has_accessible_models()` |
|---|---|---|---|---|---|
| 1.18.27 | rc=0, 441 lines | ok | ok | 7 models | True |
| 1.0.0 | **rc=1** (prints help) | **1 model** | **`OpenCodeProbeError`** | **`()`** | **True** |
| 0.5.29 | **rc=1** (prints help) | **1 model** | **`OpenCodeProbeError`** | **`()`** | **True** |

The last two columns are the problem, and they disagree with each other. `opencode_has_accessible_models()` at `cli_bin.py:167` still runs the *non-verbose* probe, so it keeps returning `True`. Readiness therefore says the harness is usable while the catalog is empty, and `_stale_models` returns `()` rather than falling back. This is a regression against the pre-PR behaviour on the same binary, which returned one model.

The failure is also silent from the operator's side: an empty model list reads as "this account has no models", not as "your `opencode` is too old". Note that `_stale_models` returning `()` means there is no cached tuple to soften the first failure either.

A version gate, or a `try verbose / except OpenCodeProbeError → non-verbose` fallback, would both close it; which one you want depends on whether you intend to support pre-`--verbose` binaries at all. That is your call, not mine — I'm reporting the behaviour, not prescribing the patch.

## B2 — robustness: malformed metadata injects phantom models, which is the opposite of what the docstring promises

`_parse_verbose_models` does not advance `offset` past a JSON block that fails `raw_decode` — on the `except json.JSONDecodeError` branch it sets `metadata = {}` and leaves `offset` where it was. Parsing then resumes *inside* the malformed block, and `_is_model_id` accepts any token containing `/` and no whitespace, so fragments of the broken JSON become model ids.

Measured against the real parser:

```
case1 (truncated JSON):
  (OpenCodeModel('opencode/big-pickle'),
   OpenCodeModel('"npm":"@ai-sdk/openai-compatible"'),      <- phantom
   OpenCodeModel('opencode/second-model', ('low',)))

case2 (minified malformed):
  (OpenCodeModel('opencode/a'),
   OpenCodeModel('{"id":"a","api":{"npm":"@ai-sdk/openai-compatible"'),   <- phantom
   OpenCodeModel('opencode/b', ('low',)))

case3 (control, well-formed):
  (OpenCodeModel('opencode/a', ('high','none')), OpenCodeModel('opencode/b', ()))
```

The docstring at line 80 states the JSON is used "only for optional variant metadata, so a malformed/missing block degrades to no variants **without inventing support or dropping a runnable model**." Case 1 and case 2 invent a model. The intent is right; the implementation doesn't hold it.

Worth saying explicitly: your test for this path uses `stdout="opencode/big-pickle\n{not-json}\n"`. That input **cannot** expose the defect, because `{not-json}` contains no `/`-bearing whitespace-free token, so the resumed scan finds nothing that passes `_is_model_id`. The test passes for a reason unrelated to the property it names. Both cases above are one-line stdout fixtures and would sit naturally next to it.

## N1 — note: `--variant none` is emitted for models whose variant table does not offer it

`local_runtime.py` maps `off` → native `none` and appends `--variant none` whenever `inference_level in OPENCODE_INFERENCE_LEVELS` and the caller hasn't already passed `--variant`. That condition is on the *harness-level* set, so it fires even for a model whose advertised variants omit `none`.

In the live catalog I could reach, 4 of 7 models advertise no variants at all, and the observed vocabulary was `minimal, low, medium, high, xhigh` — no `none`.

This is a no-op rather than a failure: `opencode run` silently ignores an unknown `--variant`. I confirmed that with a deliberate control —

```
opencode run --variant none              -> rc=0, normal completion
opencode run --variant off               -> rc=0, normal completion
opencode run --variant minimal           -> rc=0, normal completion
opencode run --variant totallybogusvariant -> rc=0, normal completion
```

The bogus control is the part that matters: `rc=0` on `none` proves nothing on its own, since `rc=0` on `totallybogusvariant` too. So "it works" here means "it is silently discarded", and a user asking for `off` on such a model gets the model's default effort with no diagnostic. Low severity, but it is a real gap between what the flag appears to do and what it does.

---

## Things I checked and found correct

**The `none` ↔ `off` mapping is right.** I want to be clear that I initially suspected it was not, because no model in the catalog I could see advertises `none`, and I could only see the unauthenticated `opencode/*` zen models. Reading the shipped binary settles it: `none` is a first-class token in opencode's own tables.

```js
reasoningEffort: Q.enum(["none","minimal","low","medium","high","xhigh","max"])
a0=["low","medium","high"], BN=["none","minimal",...a0,"xhigh"], UN=["none",...a0], Ay=["none",...zN]
```

The reason I couldn't see it in the live catalog is that the zen models run through `@ai-sdk/openai-compatible`, whose variant table is only `{high, max}` for that branch. So the absence was an artefact of which provider I could reach, not evidence about the token. The mapping is correct, and `OPENCODE_INFERENCE_LEVELS`' seven values line up with the enum above — `max` is genuine too.

**The model-level / harness-level capability boundary is deliberate and consistent.** Only the harness-level set is enforced, at `provision.py:255`, `control/client.py:453`, and `runtime_matrix.py:284`. The per-model `ModelOption.supported_inference_levels` is advisory and is omitted from the payload when empty (`control/client.py:597`) — which is the case for those 4 of 7 models. That matches the existing `pi` precedent, so I'm not calling it a finding; N1 is the one place where the advisory set and the enforced set visibly diverge in behaviour.

**`--variant` passthrough placement is correct.** `build_opencode_run_command` puts launch args before the prompt, which is what real `opencode run` yargs parsing requires. Verified live, not read.

**The model-id-line-as-authority design is a good call.** Treating the id line as the launch authority and the JSON as decoration is the right split for a surface that changes shape between releases. B2 is a bug in that design's implementation, not an argument against it.

---

*Reviewed at `7184f45657a739ed2c81500d398d74c40938cd5d`. All probe worktrees reverted and confirmed clean (`git status --porcelain` empty at that SHA). I do not write patches — the fixes are yours.*

Glimmer0x:
@codex review

src/puffo_agent/agent/opencode_auth.py
```diff
@@ -51,24 +63,138 @@ def list_opencode_models(
...
+        )
+    except OpenCodeProbeError:
+        return tuple(
+            OpenCodeModel(model_id)
+            for model_id in list_opencode_models(
+                executable,
+                provider=provider,
+                timeout_seconds=timeout_seconds,
+            )
+        )
```
hellohanchen, line 175: **Blocker cleared, and I checked it against the real binary rather than the mock.** I still had the opencode **0.5.29** package from round 1:

```
$ opencode-0.5.29 models --verbose
rc=1
stdout: "opencode models\n\nlist all available models\n\nOptions:\n      --help ..."
stderr: (empty)
```

Non-zero exit with usage text on stdout — exactly what `_run_opencode_models` converts into `OpenCodeProbeError`, so this `except` fires. End to end through that binary:

```
list_opencode_model_catalog("<0.5.29>")  ->  97 models, all variants=()
```

Pre-PR this returned an empty catalog while `opencode_has_accessible_models()` still reported True off the non-verbose probe. That was the round-1 blocker; it's gone.

Pinned too: making this `except` re-raise fails `test_verbose_probe_falls_back_to_non_verbose_models` (1 failed, 7 passed). Asserting the exact command sequence is the right call — it's what distinguishes "fell back" from "happened to return the same thing".

One small consequence worth knowing rather than changing: with the default `timeout_seconds=5.0`, a *timeout* also raises `OpenCodeProbeError`, so the worst-case preflight is now two 5s waits instead of one.

src/puffo_agent/agent/opencode_auth.py
```diff
@@ -51,24 +63,138 @@ def list_opencode_models(
...
+            raw_line == candidate
+            and _is_model_id(candidate)
+            and (metadata_offset == length or output[metadata_offset] == "{")
+        ):
+            return line_start
+        line_start = line_end + 1
+    return length
+
+
+def _parse_verbose_models(output: str) -> tuple[OpenCodeModel, ...]:
```
hellohanchen: **This helper doesn't just fail to earn its place — it loses a model.**

Deleting the call to it (`offset = _next_verbose_model_offset(output, offset)` → `pass`) **survives the whole file: 8 passed.** So I compared its output directly against doing nothing, on your own fixture (`T`) and on one where a model has no metadata block:

```
B = "opencode/a\n" '{"id":"a","api":{"npm":"x"\n' "opencode/b\n" "opencode/c\n" '{"variants":{"low":{}}}\n'

as shipped        T: clean     B: a, c        <- opencode/b DROPPED
helper removed    T: clean     B: a, b, c     <- correct
```

On `T` it is byte-for-byte equivalent to doing nothing, because the hardened `_is_model_id` already rejects the malformed line and the ordinary line loop walks straight to `opencode/second-model`. On `B` it drops `opencode/b`, because the scan only accepts a candidate whose next non-blank character is `{` — a model id followed by another model id doesn't qualify, so it walks past it.

That contradicts `_parse_verbose_models`' own docstring: *"a malformed/missing block degrades to no variants without inventing support or dropping a runnable model."* Dropping a runnable model is the more expensive half — the model the user is actually running vanishes from the picker.

Reaching it needs malformed metadata **and** a metadata-less model, so I'm not calling it a blocker. But I couldn't construct an input where these 25 lines beat the character check alone.

src/puffo_agent/agent/opencode_auth.py
```diff
@@ -51,24 +63,138 @@ def list_opencode_models(
...
+            return ""
         raise OpenCodeProbeError("OpenCode model check failed")
+    return stdout
+
+
+def _is_model_id(value: str) -> bool:
+    return (
+        "/" in value
+        and not any(char.isspace() or char in '{}[],"' for char in value)
+    )
```
hellohanchen, line 79: **This is the line that actually fixes the phantom, and nothing tests it.** Reverting it to the old whitespace-only check **survives: 8 passed** — because the helper above covers for it on your fixture. Remove both and the phantom comes back:

```
neither guard  ->  ('opencode/big-pickle', '{"id":"a","api":{"npm":"@ai-sdk/openai-compatible"', 'opencode/second-model')
```

That's the round-1 finding reproducing, so this check is genuinely load-bearing — it's just held up by a redundant guard rather than by a test. Two guards on one hazard make each other unpinned; the way out is to delete one and test the other.

It carries more weight than it appears to, as well: the recovery scan's termination argument depends on it. At the `JSONDecodeError` the offset is the start of the malformed line, that line begins with `{`, and this check is the only reason the scan doesn't return the same offset it was handed.

For what it's worth the character set looks right for real ids — `opencode/gpt-5.6`, `deepseek/deepseek-v4-pro`, `a/b:c` and `openrouter/qwen/qwen3-max` all pass; `{"id":"a/b"}`, `opencode/x,y` and `opencode/x"y` are all rejected.

hellohanchen (commented):
Round 2 on **`863192da070007ceed77eee95f133c7e71059ad3`**, delta-reviewed against my own round-1 SHA `7184f45657a739ed2c81500d398d74c40938cd5d` rather than against the PR base — the whole delta is one commit touching `opencode_auth.py` (+51/-5) and its tests (+36/-2).

**Verdict: no blocker.**

## The round-1 blocker is genuinely cleared, and I proved it against the real binary

I did not take the fallback on trust. I still had the **opencode 0.5.29** package from round 1, so I ran the actual command:

```
$ opencode-0.5.29 models --verbose
rc=1
stdout: "opencode models\n\nlist all available models\n\nOptions:\n      --help ..."
stderr: (empty)
```

Non-zero exit, usage text on stdout. That is exactly the shape `_run_opencode_models` turns into `OpenCodeProbeError`, so the new `except` fires. Then end-to-end through the real binary:

```
list_opencode_model_catalog("<0.5.29>") -> 97 models, all with variants=()
```

Pre-PR this path produced an empty catalog while `opencode_has_accessible_models()` still returned True off the non-verbose probe. It now returns the full 97. That was the blocker and it's gone.

It's also pinned: making the `except` re-raise fails `test_verbose_probe_falls_back_to_non_verbose_models` (1 failed, 7 passed). Your test asserts the exact command sequence, which is the right thing to assert here.

## The phantom-model fix works, but two of its three moving parts hold each other up

This is the substance of the round. `_is_model_id`'s new `{}[]," ` rejection and `_next_verbose_model_offset` are **redundant guards against the same phantom**, and because each covers for the other, neither is pinned. Both mutations survive the full file (8 passed):

```
delete  offset = _next_verbose_model_offset(output, offset)   -> 8 passed  (survives)
revert  _is_model_id to the old whitespace-only check         -> 8 passed  (survives)
```

So I ran the four combinations against two inputs — your own test fixture (`T`) and one where a model has **no** metadata block (`B`):

```
B = "opencode/a\n" '{"id":"a","api":{"npm":"x"\n' "opencode/b\n" "opencode/c\n" '{"variants":{"low":{}}}\n'

both (as shipped)              T: clean          B: a, c            <- opencode/b DROPPED
helper removed, _is_model_id   T: clean          B: a, b, c         <- correct
helper kept, _is_model_id soft  T: clean          B: a, c            <- opencode/b DROPPED
neither                        T: PHANTOM        B: a, b, c
                                  '{"id":"a","api":{"npm":"@ai-sdk/openai-compatible"'
```

Two things follow.

**The hardened `_is_model_id` is the fix.** With the helper removed and only the character check in place, `T` is clean *and* `B` keeps all three models. Softening `_is_model_id` is the only change that brings the phantom back. That's the load-bearing line, and nothing tests it directly.

**`_next_verbose_model_offset` doesn't just fail to help — it loses a model.** On `T` it produces output identical to doing nothing. On `B` it drops `opencode/b`, because the helper only accepts a candidate line whose next non-blank character is `{`; a model id followed by another model id doesn't qualify, so the scan walks past it. That contradicts this function's own docstring — *"a malformed/missing block degrades to no variants without inventing support or dropping a runnable model"* — and dropping a runnable model is the more expensive half of that sentence, because a model the user is actually running disappears from the picker.

It needs both malformed metadata *and* a metadata-less model to reach, so it's not a blocker. But the 25 lines currently buy nothing on any input I could construct where the character check alone doesn't do better.

## What I verified as correct

```
C  malformed metadata as the final block          -> ('opencode/a',)                     ok
D  malformed JSON spanning four lines, then a
   well-formed pair                               -> a, b('high')                        ok
E  a model-id-looking string inside the malformed
   JSON body                                      -> a, b   (no phantom 'opencode/ghost') ok
F  _is_model_id on real shapes: 'opencode/gpt-5.6', 'deepseek/deepseek-v4-pro', 'a/b:c',
   'openrouter/qwen/qwen3-max' all True; '{"id":"a/b"}', 'opencode/x,y', 'opencode/x"y' all False
```

E is worth calling out: the character check is what stops it, which is the same line I'd like to see tested.

I also checked the recovery scan can't return the offset it was given (that would loop forever): at the point of the `JSONDecodeError`, `offset` is the start of the malformed line, that line begins with `{`, and the hardened `_is_model_id` rejects it. That's fine as written — but note it is *another* thing resting on the character check, so softening `_is_model_id` is more load-bearing than it looks.

**CI is green on this exact SHA** — `pytest (py3.11)`, `pytest (py3.12)`, `pre-commit`, and all three CodeQL jobs `COMPLETED SUCCESS`. I checked the rollup immediately before submitting rather than relying on the run being started.

My round-1 note about `--variant none` being emitted for models that don't advertise it is untouched by this delta and still stands as a note, not a finding.

---

*Round 2 of 3. Reviewed at `863192da070007ceed77eee95f133c7e71059ad3`, delta from `7184f456`. All mutations and probes reverted; worktree confirmed clean. Review only — the fix is yours to write.*

hellohanchen:
## Review verdict: BLOCKER — bound to `863192da070007ceed77eee95f133c7e71059ad3`

Head re-queried at verdict time: unchanged, OPEN/MERGEABLE, 7/7 checks pass. Local baseline **3862 passed, 4 skipped, 0 failed** (`tests/test_pi_tool_bridge.py` deselected — it fails identically on the merge-base on this host, so it's environmental, not a regression).

Seven mutants, each applied alone, diff-verified before running, restored between, run against the full suite:

| # | Mutation | Result |
|---|---|---|
| MA | `local_runtime`'s whole `--variant` block deleted | **2 failed** |
| MB | `off` → `none` mapping dropped | **1 failed** |
| MC | `model_catalog`'s `off` ← `none` clause dropped | **1 failed** |
| MF | `--verbose` never appended to the command | **2 failed** |
| MD | recovery lookahead relaxed (no `{` requirement) | **survives** |
| ME | `_is_model_id` punctuation guard weakened to whitespace-only | **survives** |
| MG | `_parse_verbose_models`' dedup `seen` check removed | **survives** |

The launch path is genuinely held — MA, MB and MF all die. The parser is not.

### B1 (blocking) — `_parse_verbose_models` silently drops runnable models

The recovery scanner requires the next bare model-id line to be followed by `{` or EOF (`opencode_auth.py`, `_next_verbose_model_offset`). When a malformed metadata block is followed by models that have **no** JSON block, every one of them is skipped until a model with a block turns up:

```
a/one
{"variants":{oops}
b/two
c/three
d/four
e/five
{"variants":{"low":{}}}
```

→ `['a/one', 'e/five']`. Three runnable models vanish from the picker.

This contradicts the function's own docstring — *"degrades to no variants without inventing support **or dropping a runnable model**"*. And blockless models are not an unhandled shape: on the normal path the parser handles them correctly (`a/b`, no block, then `c/d` with one → all three parse), and the `--verbose`-unsupported fallback produces nothing *but* blockless models.

`test_verbose_probe_recovers_after_malformed_metadata_without_phantom_models` passes because its fixture puts a `{` block on the model immediately after the malformed one. The name promises no *phantom* models and that half is real; the "doesn't drop a runnable model" half is neither promised by the name nor held by the fixture.

MD isolates the cause — deleting the lookahead survives 3862/3862, so nothing observes it today.

### B2 (blocking) — a zero-variant model is published as *unrestricted*

`ModelOption.supported_inference_levels` defaults to `()`, and `portal/control/client.py` omits the key entirely when it's falsy:

```python
**({"supported_inference_levels": list(o.supported_inference_levels)}
   if o.supported_inference_levels else {})
```

So `()` is indistinguishable on the wire from "this producer said nothing", and a consumer falls back to the harness-wide list — now all seven levels, since `supported_inference_levels("opencode")` was widened in this PR.

This is not hypothetical: it is the case your own test asserts. `test_opencode_catalog_uses_models_visible_to_native_cli` pins `options[1].supported_inference_levels == ()` for `opencode/no-variants`, whose fixture is `{"variants":{}}`. The model that advertises *zero* variants ends up offered *more* levels than `deepseek-v4-pro`, which advertises four.

The codebase already has the right pattern one branch away — `model_catalog.py:260` gives pi `PI_INFERENCE_LEVELS if supports_thinking else ("off",)`, never `()`. The opencode branch collapses "supports nothing" and "unknown" into the same value.

### N1 (non-blocking, but must ship with B1's fix) — the punctuation guard is unpinned, and the lookahead is what masks it

ME survives: weakening `_is_model_id` to reject whitespace only, keeping the `{}[],"` characters, changes nothing. It looks harmless because the `{` lookahead currently prevents the scanner from ever landing on a JSON line.

Relax the lookahead — the fix for B1 — and the punctuation guard becomes the *only* thing holding the no-phantom guarantee. With both changed, your own malformed fixture yields:

```
['opencode/big-pickle', '{"id":"a","api":{"npm":"@ai-sdk/openai-compatible"', 'opencode/second-model']
```

A phantom model, in the test named for not producing one. So B1's fix needs a test that pins `_is_model_id`'s punctuation rejection directly, not just the end-to-end parse.

I verified the minimal relaxation works and keeps your existing fixture green:

```diff
-        if (
-            raw_line == candidate
-            and _is_model_id(candidate)
-            and (metadata_offset == length or output[metadata_offset] == "{")
-        ):
+        if raw_line == candidate and _is_model_id(candidate):
```

returns all five models in the B1 case above, and still parses your malformed fixture to exactly `(OpenCodeModel("opencode/big-pickle"), OpenCodeModel("opencode/second-model", ("low",)))`.

### N2 (non-blocking) — `_parse_verbose_models`' dedup is unpinned

MG survives. `list_opencode_models` has its own `seen` set with pre-existing coverage; the new verbose parser's does not. A repeated model id would show up twice in the picker.

### N3 (non-blocking) — no per-model enforcement anywhere

`local_runtime` gates the `--variant` append on the global `OPENCODE_INFERENCE_LEVELS`, never on the selected model's advertised variants. Every validation site — `state.py::_validate_inference_level`, `daemon.py::_validate_refresh_inference_level`, `control/client.py`, `runtime_matrix.py::normalize_inference_level` — checks the harness-wide list too. So a level the model doesn't advertise passes every layer and reaches `opencode run --variant <x>`.

That may be deliberate for this PR's scope, but note `test_opencode_runtime_forwards_selected_variant`'s docstring says *"A model-advertised reasoning level must reach `opencode run`"* while the fixture never establishes that `openai/gpt-5` advertises `high` — nothing in the code consults the advertisement. The test holds forwarding, not advertisement.

### N4 (non-blocking) — no changelog entry

`[Unreleased]` is empty and this PR adds nothing to it, for a user-facing feature. Neighbouring entries are tagged with their PR number.

### Caveat

`opencode` is not installed on this host, so I could not probe real `models --verbose` output. B1's behaviour is proven against the parser; how often upstream emits a blockless line after a malformed block is not something I can tell you. The recovery path exists at all because malformed metadata was observed, and your fixture is a realistically truncated `api` block — so I don't think the trigger is theoretical, but I'd rather say what I checked than imply more.

Glimmer0x:
@codex review

hellohanchen:
## Round 3 — `c5c5dee020e5ce20043ea38413303f4d24c0a3d5`

**Both blockers are fixed. No blocker on this SHA.** Two test-coverage gaps below, neither of which affects the shipped behaviour.

Head re-queried immediately before submitting: `c5c5dee020e5ce20043ea38413303f4d24c0a3d5`, base `main`, OPEN/MERGEABLE, 7/7 checks SUCCESS.

Delta from the head I last reviewed (`863192da`) is 6 files, +89/−28.

### B1 — verified by running the failure, not by reading the fix

`_next_verbose_model_offset` is gone; the parser now resumes line-by-line and rejects indented / punctuation-bearing lines instead. I re-ran the shape that lost a model under the old lookahead — malformed metadata followed by **two consecutive bare model-id lines with no block of their own**:

```
PROBE B1 parsed: [('anthropic/claude-alpha', ()),
                  ('anthropic/claude-beta',  ()),
                  ('anthropic/claude-gamma', ('high',))]
```

All three survive. Under `863192da` the middle one was dropped. The phantom-model guards still hold on the same input class:

```
PROBE guards: ['anthropic/claude-alpha', 'provider/real-model']
```
— the indented `  provider/indented-model` and the quoted `"provider/quoted-model",` are both rejected.

### B2 — verified by mutation

`build_capabilities`'s `if o.supported_inference_levels is not None` is the right shape, and it is pinned: reverting that one line to the original truthiness check turns the suite red. The `None` default also has exactly one consumer — I grepped every reader of `ModelOption.supported_inference_levels`, and the other twenty hits are the unrelated module-level `mcp.config.supported_inference_levels(harness)` — so the type change can't crash an iterator somewhere else.

### Mutation results

Each mutant applied alone, diff-verified that it landed, run against `test_opencode_auth.py` + `test_harness_readiness.py` + `test_model_catalog.py`, restored between. This repo's pytest config suppresses the summary line, so before trusting any of it I ran a deliberate-breakage control to confirm the harness can see a failure at all (`exit=1 F=27 E=55`).

| # | Mutant | Result |
|---|---|---|
| M1 | drop `raw_line != model_id` (the indentation guard) | *survived* |
| M2 | weaken `_is_model_id`'s punctuation set to whitespace-only | **killed** |
| M3 | drop `_is_model_id`'s `"/" in value` requirement | **killed** |
| M4 | revert `is not None` to truthiness (the original B2 bug) | **killed** |
| M5 | change the `ModelOption` default from `None` back to `()` | *survived* |

**M1 — the indentation guard is unpinned, and one new test makes it look otherwise.** `raw_line != model_id` is the only thing that rejects an indented model-id line, because the parser calls `_is_model_id(model_id)` on the **stripped** value, where the `char.isspace()` check can never fire. The new parametrized case `(" opencode/indented", False)` passes an unstripped string straight to `_is_model_id` — a call shape the parser never produces — so it reads as indentation coverage while leaving the real guard untested. Neither malformed-JSON fixture (`test_verbose_probe_recovers_after_malformed_metadata_without_phantom_models`, `test_verbose_probe_keeps_blockless_models_after_malformed_metadata`) contains an indented line; both rely on punctuation to reject the JSON body. One fixture with an indented `  provider/x` line inside the bad block closes it.

**M5 — nothing pins the `None` default, which is the whole point of the B2 fix.** The new capability test constructs `ModelOption(..., supported_inference_levels=())` explicitly, so it proves `() → []`. Nothing proves *omitted → key absent*. Probed:

```
PROBE default value    : None
PROBE wire (as shipped): {"id": "opencode/unclaimed"}
PROBE wire (M5 default): {"id": "opencode/unclaimed", "supported_inference_levels": []}
PROBE daemon default   : None
```

So M5 is not cosmetic — it inverts B2, publishing an explicit "this model has no selectable levels" claim for every model that makes no claim, `_DAEMON_DEFAULT` included. A second assertion on a `ModelOption` that omits the argument would pin it.

Both gaps are the untested half of a two-part mechanism: the tests hold the branch that was wrong and leave its counterpart free. Worth a follow-up, not a re-spin.

PR #321, opened 2026-09-07
feat(exa): add exa_search agent tool routed through the server gateway

KurtneySh:
## Background

Agents fetch paid external data through the server's spend gateway rather than holding vendor keys themselves (as with the existing `monid_spend` tool). This adds an `exa_search` tool so an agent can run an Exa web search; the tool forwards to the server's `/v2/exa/search` endpoint, which holds the Exa key, prices the call, enforces budget caps, and meters it.

## What changed

- New `core_exa_tools.py` with `register_exa_tools`, shaped after `core_monid_tools`: a thin `exa_search` tool that forwards `query` / `type` / `numResults` to the server's `POST /v2/exa/search`.
  - The agent **never holds the Exa key** and **never reports its own identity slug** (the server derives it).
  - Errors surface the server's plain message, not raw vendor output; the `202 pending` and already-settled responses are handled.

## Scope / non-goals

- The agent side is **transport only** — all pricing, budget caps, idempotency, and metering live on the server. Search results are returned to the agent in the moment and are **not stored or logged**.
- This PR targets `main`; it has no git dependency on the server gateway branch. The tool talks to `/v2/exa/search` over HTTP at runtime, which ships in the companion server PR — so Exa search only functions once that server PR is deployed.

## Verification

- Unit tests (7) cover the tool's request shaping, that it never carries a key or a self-reported slug, native-only behavior, plain error surfacing, and the pending / already-settled paths. `ruff` is clean; the full suite (3,839+) is unaffected.
- Local: with the server running, `EXA_API_KEY` set, and a seeded budget, invoke `exa_search` from an agent and confirm results return and the key never reaches the agent process logs.

## Release note

Internal / not user-facing yet.

### DIFF
--- a/src/puffo_agent/mcp/core_exa_tools.py
+++ b/src/puffo_agent/mcp/core_exa_tools.py
@@ -135,10 +135,19 @@ def _register_exa_search(mcp: FastMCP, cfg: Any) -> None:
 def _format_search_result(data: dict[str, Any]) -> str:
     """Render a `/v2/exa/search` response for the model.
 
-    An idempotent retry of an already-settled search returns the accounting with
-    null results (results are never stored, so they cannot be replayed) — report
-    that rather than pretending there are fresh results.
+    A 202 is not an HTTP error to the client, but it means a prior search under
+    this idempotency key is still resolving and an owner will reconcile it —
+    there is no result yet. An idempotent retry of an already-settled search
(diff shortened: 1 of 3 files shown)


KurtneySh:
Closing: the exa search demo is no longer planned, so this exa agent tool is not going forward. Reopenable if exa is revived.

PR #342, opened 2026-09-10
fix(acp): honour permission_mode instead of always asking a human

Glimmer0x:
The ACP driver never read `spec.permission_mode`. Its sibling drivers do —
codex sets `approvalPolicy: "never"` and opencode passes `--auto` — so a
runtime configured for `bypassPermissions` still parked every tool call on an
operator, waiting on a `resolve_permission` that no product UI can send.

`VALID_PERMISSION_MODES` holds exactly one value (`bypassPermissions`) and the
runtime coerces anything else to it, so this was not a policy knob with a
default: unattended operation was impossible on this path.

## Approach

ACP has no launch-time equivalent of those flags — in ACP the client *is* the
authority, so the policy lands where the request is answered. The answer is
derived from the single request in hand: no future is registered, no grant is
cached, and the peer still asks again next turn. That keeps the LingTai ACP
contract's turn-scoped permissions intact and is distinct from a persistent
permission capability (their `ACP001` fails anyone claiming one). Reviewed on
the LingTai side against that contract.

Auto-approval deliberately does **not** reuse `_preferred_permission`, which
falls back to `options[0]`; with a reject-only option set that would return
`outcome="selected"` carrying a denial. `_allow_shaped_permission` returns
`None` instead and the request falls through to a human.

The emitted event carries `auto_allowed` so the record distinguishes an
auto-approved request from one still waiting on a person.

## Verification

- Two new tests; both fail against unpatched `acp.py` (negative control).
- Targeted mutation: swapping the auto path back to the lenient selector
  reddens only the reject-only test, so it pins that specific defect.
- A pre-existing test was implicitly relying on the bug — it used the default
  spec (which *is* `bypassPermissions`) while asserting a human wait. Now
  explicitly non-bypass so it still covers that path.
- 779 passed across the acp/driver/authority/lingtai/runtime selection.
- Exercised end to end on a local full stack: two unattended rounds with zero
  `runtime.resolve_permission` sent, four tool calls all `auto_allowed=true`,
  replies delivered to the browser.

## Note for reviewers

With this merged there is no product configuration that restores a human
approval step on the ACP path, because `VALID_PERMISSION_MODES` has one member
and no UI renders the permission request. Making human confirmation a real
option needs both a new member in that frozenset and a frontend affordance —
tracked separately.

### DIFF
--- a/src/puffo_agent/agent/harness/drivers/acp.py
+++ b/src/puffo_agent/agent/harness/drivers/acp.py
@@ -248,6 +248,7 @@ class AcpDriver(Driver):
         self._fallback_block_id = ""
         self._capabilities = acp_capabilities(session_resume=False)
         self._model_selection = ""
+        self._permission_mode = ""
         self._spawn_warnings: tuple[str, ...] = ()
         self._driver_authority: DriverAuthorityServer | None = None
         self._closed = False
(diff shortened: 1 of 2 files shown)


Glimmer0x:
Consolidated at the maintainer’s request into #346. Please review and merge that combined PR; this PR is closed as superseded, not merged.

PR #311, opened 2026-09-03
fix(harness): echo driver authority call IDs

TZZheng:
## Summary
- echo a request `call_id` from every driver-authority response path
- preserve compatibility for legacy requests without a `call_id`
- add a root → derived → provider regression covering all correlated exchanges

## Verification
- `uv run --extra dev pytest -q tests/test_driver_authority_server.py`
- `uv run --extra dev ruff check src/puffo_agent/agent/harness/driver_authority_server.py tests/test_driver_authority_server.py`
- actual Puffo server + LingTai main client: root hello → derived launch → child hello → provider call

This is intended to be merged into #304 before #304 is merged to `main`.

PR #346, opened 2026-09-10
feat(lingtai): consolidate directory provisioning and ACP runtime support

Glimmer0x:
Enables Puffo to create and run an initialized LingTai agent directory through ACP, persist its workspace and registry association, correlate driver authority replies, and honor the selected ACP permission mode.

Consolidates #311, #342, #344 and the original #346 into this single daemon merge target. Post-provision failures perform bounded rollback even across repeated cancellation, then re-raise the original failure. The cleanup does not guarantee recovery from a hard process kill. Operator-selected workspaces may be shared; they are not registry directories or an isolation boundary.

Validation: 4,043 passed, 2 skipped in the combined Python run; excludes test_agent_install.py and test_host_mcp_handler.py because their inherited CODEX_HOME fixtures can modify the host configuration. This is not an unmodified full-suite pass. Ruff and the Python structural check pass. The only post-run change compresses an ACP comment by one line to meet the 100-line function check. Combined-build browser evidence is recorded below.

Companion LingTai PR: Lingtai-AI/lingtai-kernel#1690 at ad11d7bc. No merge or release has been performed.

ACP policy decision: this change makes ACP consume the existing platform permission mode. The current platform accepts only bypassPermissions (VALID_PERMISSION_MODES has one entry), defaults to it, and its runtime sanitizers also fall back to it for unknown values. There is currently no supported configuration value that requests manual ACP permission handling. ACP therefore automatically selects an allow-shaped option under that mode; when no allow-shaped option exists it asks the operator. This aligns ACP with the existing platform default, including the default-allow behavior. It does not change the platform-wide defaults. The forwarded call_id remains necessary for AuthorityAuditRecord; the wire-level stamp separately correlates response frames.

Review boundaries: the prior provisioning/rollback review covered d414e15a, not the entire consolidated head. The combined head includes the final source branches of #311, #342 and #344; review of the new combination remains separately tracked.

Latest local validation: installed daemon b9fad710, frontend ca78938c (#949), and local LingTai merge ea2a3a63 created LingTai Demo in the browser, joined its group automatically, completed two real Claude read_inbox/send_message rounds H/I with previous-round recall, and returned idle without manual permission commands. Existing registered test account; registration was not repeated.

LingTai dependency synchronized: owner pushed #1690 to ad11d7bc7b1473bb69e354aa7a5bf142e7fbec84. Fetched and verified its complete tree equals the tested local ea2a3a63 tree, 11e43c511d078943c9b75884947fefe027328ded; full git diff is empty. H/I acceptance therefore covers the same source content now in #1690. The installed wheel remains built from local ea2a3a63; no rebuild or restart was performed during manual testing. Plain 133f4352 lacked #1694 and failed messaging. The consolidated PRs remain unmerged.

### DIFF
--- a/src/puffo_agent/agent/harness/driver_authority_server.py
+++ b/src/puffo_agent/agent/harness/driver_authority_server.py
@@ -166,6 +166,16 @@ class DriverAuthorityServer:
             while True:
                 request = self._recv_frame(record)
                 response, child = self._handle_request(record, request)
+                # Correlation is stamped here, at the one point every response
+                # passes through on its way to the wire. A peer that matches
+                # responses to requests by call_id rejects any reply that omits
+                # it, and this server has ten leaf response paths; threading a
+                # parameter through each one means a path added later can drop
+                # the echo silently. Only a request that correlates gets a
+                # correlation key back.
+                request_call_id = request.get("call_id")
+                if isinstance(request_call_id, str):
+                    response["call_id"] = request_call_id
                 try:
                     self._send_frame(record.server_socket, response, child=child)
                 finally:
@@ -218,15 +228,12 @@ class DriverAuthorityServer:
                 self._claim_probe()
             record.state = _LeaseState.CLAIMED
             binding = record.binding
-        response = {
+        return {
             "version": PROTOCOL_VERSION,
             "role": binding.role,
             "launch_id": binding.launch_id,
             "capability": binding.capability,
         }
-        if call_id is not None:
-            response["call_id"] = call_id
-        return response
 
     def _authorize_derived_launch(
         self, record: _EndpointRecord, request: dict[str, Any], *, call_id: str | None = None
@@ -336,15 +343,12 @@ class DriverAuthorityServer:
                 call_id,
             )
         )
-        response = {
+        return {
             "version": PROTOCOL_VERSION,
             "state": state,
             "reason_code": reason_code,
             "audit_id": audit_id,
         }
-        if call_id is not None:
-            response["call_id"] = call_id
-        return response
 
     @staticmethod
     def _send_frame(
--- a/src/puffo_agent/agent/harness/driver_authority_server.py
+++ b/src/puffo_agent/agent/harness/driver_authority_server.py
@@ -187,43 +187,49 @@ class DriverAuthorityServer:
         self, record: _EndpointRecord, request: dict[str, Any]
     ) -> tuple[dict[str, Any], socket.socket | None]:
         operation = request.get("op")
+        call_id = request.get("call_id")
+        if not isinstance(call_id, str):
+            call_id = None
         if request.get("version") != PROTOCOL_VERSION or not isinstance(operation, str):
             return self._decision(
-                record, "malformed", "indeterminate", "malformed_request"
+                record, "malformed", "indeterminate", "malformed_request", call_id=call_id
             ), None
         if operation == "hello":
-            return self._claim(record), None
+            return self._claim(record, call_id=call_id), None
         with self._lock:
             claimed = record.state is _LeaseState.CLAIMED
         if not claimed:
-            return self._mismatch(record, operation), None
+            return self._mismatch(record, operation, call_id=call_id), None
         if operation == "authorize_derived_launch":
-            return self._authorize_derived_launch(record, request)
+            return self._authorize_derived_launch(record, request, call_id=call_id)
         if operation == "authorize_provider_call":
             return self._authorize_provider_call(record, request), None
         return self._decision(
-            record, operation, "indeterminate", "unsupported_operation"
+            record, operation, "indeterminate", "unsupported_operation", call_id=call_id
         ), None
 
-    def _claim(self, record: _EndpointRecord) -> dict[str, Any]:
+    def _claim(self, record: _EndpointRecord, *, call_id: str | None = None) -> dict[str, Any]:
         with self._lock:
             if record.state is not _LeaseState.ISSUED:
                 return self._record_decision_locked(
-                    record, "hello", "denied", "endpoint_already_claimed"
+                    record, "hello", "denied", "endpoint_already_claimed", call_id=call_id
                 )
             if self._claim_probe is not None:
                 self._claim_probe()
             record.state = _LeaseState.CLAIMED
             binding = record.binding
-        return {
+        response = {
             "version": PROTOCOL_VERSION,
             "role": binding.role,
             "launch_id": binding.launch_id,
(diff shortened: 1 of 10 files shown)


Glimmer0x (commented):
Reviewed at `bf38fecd`. The shape is right: provisioning the runtime before the identity materializes is the correct order, and the test that asserts the marker exists inside `materialize` pins it rather than assuming it. The PR body's accounting of the 8 Codex failures is honest about what was and wasn't established.

Four findings, most severe first.

### 1. The advertised rollback is not covered by any test

The body says "Failed materialization attempts revoke the association." That is the behaviour protecting against an orphaned LingTai registry entry, and nothing exercises it.

`test_lingtai_provision_failure_leaves_identity_unmaterialized` covers the *other* branch — LingTai provisioning fails, so `materialize` is never reached and no rollback is expected. Across the whole diff, `revoke` appears only in `lingtai.py` and `provision.py`; there are zero occurrences in added test lines.

A failing `materialize` (or a `write_agent_from_context` that raises) is the case that needs a test: assert `revoke_lingtai` ran with the same `--runtime-id` that `provision` used.

### 2. Rollback silently does not run under cancellation, and replaces the original exception

```python
except BaseException:
    if launch is not None:
        try:
            await revoke_lingtai(launch)
        except Exception:
            logger.warning("LingTai rollback failed for runtime %s", launch.runtime_id)
    raise
```

`asyncio.CancelledError` derives from `BaseException`, not `Exception`. If the task is cancelled during `materialize` — daemon shutdown mid-creation is the realistic trigger — then:

- entering the handler is correct (`except BaseException` catches it), but
- `await revoke_lingtai(launch)` re-raises `CancelledError` at the first await, before `create_subprocess_exec` ever spawns anything, so **no revoke happens**;
- `except Exception` does not catch it, so it propagates out of the handler and the trailing `raise` never executes — the original exception is swapped for `CancelledError`, and the warning is never logged.

Net result: the exact orphan this block exists to prevent, with no trace. Widening to `except BaseException` around the rollback (and re-raising the original) or shielding the rollback would fix it; either way it needs the test from finding 1 to stay fixed.

### 3. Provisioning failures throw away the reason

`_command` routes both `stdout` and `stderr` to `DEVNULL` and raises a fixed string suggesting two possible causes. When this fires on a real machine, LingTai's own error text — the thing that says which of the two it actually was, or that it was neither — is already gone.

We spent this evening watching what unexplained local failures cost. Please capture `stderr` (bounded, e.g. last 2KB) and include it in the raised message. The 30s timeout case should say it timed out rather than reusing the same string.

### 4. `workspace` is an operator-supplied path with essentially no constraint

`parse_lingtai_launch` checks `executable` is an executable file and `agent_dir` contains `init.json`, but `workspace` only has to satisfy `is_dir()`. That value then becomes the agent's `workspace_dir` and the `cwd` for the provision/revoke subprocesses.

So an operator can select `~`, another agent's workspace, or `<daemon-home>` itself — the last putting every agent's key material inside the new agent's workspace. Same-uid siblings can already read each other's keys, so this doesn't create the exposure, but it removes the last bit of friction in front of it.

This is also asymmetric with the contract being settled on the LingTai side right now: LingTai is about to *refuse* operator-supplied registry directories that aren't dedicated, rather than quietly accepting them. Puffo accepting any existing directory here answers the same question the opposite way.

Minimum I'd want: refuse `home_dir()` and anything inside it, and refuse a directory already registered as another agent's workspace. Rejecting loudly is fine — the operator picked it in a folder dialog and can pick again.

---

Findings 1 and 2 are the same defect from two directions and I'd like both addressed before merge. 3 and 4 are my recommendation, not a block — say if you disagree on 4 in particular, since "operator picked it deliberately" is a defensible answer.

I have not run this branch; the above is from reading the diff at `bf38fecd`.

Glimmer0x (commented):
Re-reviewed at `d414e15a`. **Both blocking findings are resolved.** Detail below, plus one correction I owe and one non-blocking observation.

### Finding 1 — rollback coverage: resolved

`test_lingtai_creation_failure_revokes_association` is parametrized over both post-registration failure points (`materialize` and `write`), asserts the association is gone, and asserts `caught.value is original` — so it pins identity of the propagated exception, not just its type. `test_lingtai_rollback_failure_preserves_original_error` covers the case where cleanup itself fails: original error survives, warning is logged. That's the behaviour the PR body advertises, now actually exercised.

### Finding 2 — cancellation: resolved, and my stated mechanism was wrong

I claimed the first `await` inside the rollback would re-raise `CancelledError` immediately. That is not how it works — a single `Task.cancel()` that has been caught does not automatically re-raise at the next await. Thank you for checking it against the runtime instead of taking my word for it; that was the right call and I should have verified before asserting a mechanism.

The real failure mode is the one you reproduced: a *second* cancellation arriving during cleanup interrupts `revoke`, and because the guard was `except Exception` it escaped, taking the original exception with it. `test_lingtai_repeated_cancellation_finishes_rollback` drives exactly that — cancel during materialize, cancel again while revoke is blocked, then assert both that the association was removed and that `caught.value.args == ("first cancellation",)`. Asserting the *first* message survives is what makes this test able to fail for the right reason.

That it was red on the old code is the part that matters, and you stated it.

### Non-blocking observation on `_rollback_lingtai`

```python
except asyncio.CancelledError:
    continue
```

This swallows every cancellation and never re-raises or calls `uncancel()`. Two consequences worth a comment in the code rather than a change:

1. The provision task is effectively **uncancellable for up to 30s** while rollback runs. That is the deliberate trade — you want the revoke to finish — and you already documented that a hard process kill still defeats it. Worth saying in the source so the next reader doesn't "fix" the loop.
2. If the original failure was *not* a cancellation (say the `RuntimeError` path) and a cancel arrives during rollback, that cancellation is silently dropped: `task.cancel()` returns `True`, but the task completes by raising the original `RuntimeError` instead of ending cancelled. A caller that cancels and then inspects the outcome will be misled.

Also: should this ever run inside an `asyncio.TaskGroup`, swallowing cancellation without `Task.uncancel()` breaks the group's semantics. Not the case today; cheap to note next to the loop.

### Not re-raised

`workspace` — you answered the question I actually asked, and documenting the boundary (user-designated shared working directory, not an inter-agent isolation surface) is a legitimate resolution. Dropping it. Same for stderr diagnostics being deferred; you recorded it rather than silently skipping it.

Verification scope: I read the diff at `d414e15a` and did not run the branch. Your "4,032 passed / 2 skipped with two Codex-touching modules excluded, not unconditionally green" phrasing is the right shape — it says what was and wasn't established.

Glimmer0x:

PR #347, opened 2026-09-11
Discover installed LingTai runtimes and existing agents

Glimmer0x:
Importing a LingTai agent currently requires three absolute paths. Add the paired-machine command `discover_lingtai` so the web app can locate installed executables and list existing agents through LingTai's read-only `puffo-v0 discover --json` interface.

Default locations come from PATH/common executable locations, the operator's existing associations, their workspaces and explicit `.lingtai` roots. Optional `root` and `executable` inputs support another project/install. Null workspace defaults to the existing agent directory; registered and drifted states remain intact. Discovery is bounded in time/output and returns explicit partial/truncated flags and warnings below the server's 16 KiB persisted-result limit. Provision failures preserve bounded CLI error text. No LingTai upstream changes or running-service changes.

Stacked on #346. Companion web change: https://github.com/puffo-ai/puffo-core-han-group/pull/950. Discovery is an inventory, not admission approval; provision remains authoritative. Arbitrary project locations cannot be enumerated without an index and may require an explicit root.

Validation: latest daemon suite excluding the two host-configuration-mutating modules (`test_agent_install.py`, `test_host_mcp_handler.py`) finished with 4,052 passed, 2 skipped and 1 failure: installed OpenCode `debug skill` timed out after 30 seconds in `test_desired_skill_is_discoverable_by_real_opencode_cli`. A single isolated rerun of that OpenCode test passed. Ten discovery tests and the full provision test file pass (49 combined); Python structural checks pass. The earlier real installed LingTai read-only smoke verified hidden-root discovery, null-workspace defaulting and no registry creation. The active manual-test stack was not restarted.

Review follow-up: incomplete runtime configurations are loaded permissively and malformed per-agent configs are skipped with `invalid_candidate`. Input root/executable limits now set partial/truncated as well as output limits. Explicit path errors return actionable error text. Both background dispatch decisions share `BACKGROUND_OPS`; discovery warning values remain closed.

### DIFF
--- a/src/puffo_agent/portal/control/client.py
+++ b/src/puffo_agent/portal/control/client.py
@@ -33,6 +33,7 @@ from ...tasks import spawn
 
 log = logging.getLogger("puffo_agent.control")
 
+BACKGROUND_OPS = frozenset({"create", "discover_lingtai"})
 RECONNECT_BACKOFF_SECONDS = 3.0
 ME_INTERVAL_SECONDS = 30.0
 # Codex's probe costs a real (tiny) turn — slow cadence; refresh_usage is on-demand.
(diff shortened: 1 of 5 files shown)


Glimmer0x:

PR #348, opened 2026-09-11
feat(runtime): auth_mode — run an agent on a subscription plan, not the gateway

shanjing:
## What

Adds `runtime.auth_mode` so an agent can authenticate to its model with the **operator's own subscription plan** instead of the metered LiteLLM gateway.

| `auth_mode` | Credential | Who pays |
|---|---|---|
| **`api-gateway`** (default) | LiteLLM virtual key + `llm_base_url` | metered account |
| `subscription` | operator's Claude/ChatGPT plan | the operator |

Behaviour on `api-gateway` is unchanged, and absent/empty means `api-gateway`. **The default is deliberate**: this package is vendored into sandbox images that may be promoted to production, where agents must stay on the metered gateway. An agent opts into subscription billing explicitly.

## Why the resolver returns a struct, not an env mapping

The two harnesses are **not symmetric**:

- **claude-code** takes a `claude setup-token` OAuth token as an environment variable.
- **codex** has no equivalent variable — its plan auth lives in an `auth.json` under `CODEX_HOME`, so the credential must be *written to disk* and the child told where to look.

`ChildCredentials` carries `env`, `files` and `extra_allowed`, so the codex branch drops in without reshaping the interface. An env-only design would need rewriting the day Codex ships.

## Three properties worth reviewing closely

**Fails closed.** A subscription agent with no credential raises `SubscriptionCredentialMissing` rather than falling back to the gateway. A silent fallback would keep working, bill the metered account, and look like success — the one failure mode that hides itself.

**The secret never enters `agent.yml`.** That file is on disk, synced and backed up; a token in it would outlive every place we can revoke it. The daemon resolves it once at startup — mirroring `claude_cli_api_key` — and passes it down. This also satisfies the existing invariant that the harness boundary never re-reads `os.environ` (`test_child_env_allowlist.py`), which caught an earlier draft of this change.

**Subscription supersedes the gateway env entirely** rather than merging. A leftover `ANTHROPIC_BASE_URL` outranks the plan token in the CLI's own credential precedence, so merging would silently route back through the gateway.

Both subscription variable names join `PROVIDER_CREDENTIAL_ENV_NAMES`, so ambient inheritance and `env_overrides` cannot smuggle one in — `controlled` stays the only channel that reaches a child.

## Not in this PR

- **AIM / provisioner side** (cloud-infra): persisting `auth_mode`, reading the token from Secrets Manager, injecting via `Sandbox.create(envs=…)`, and skipping virtual-key mint/renew/GC for subscription agents.
- **Webapp RUNTIME selector**: `Claude Code API` / `Claude Code Subscription` (+ the Codex pair), which is what sets this field.
- **Codex subscription is designed but unexercised.** The `auth.json` shape is inferred from `sanitize_codex_auth_blob` and the Codex CLI's login model, not verified against a real Codex plan. The claude-code path does not depend on it; the Codex UI option should stay disabled until it is proven.

## Passing-Evidence

```
$ ruff check <changed files>
All checks passed!

$ python -m pytest tests/ -q \
    --ignore=tests/test_docker_codex_wiring.py \
    --ignore=tests/test_reanswer_boundary.py
4124 passed, 4 skipped, 25 warnings in 53.52s

$ python -m pytest tests/test_subscription_credentials.py tests/test_child_env_allowlist.py -q
107 passed
```

The two ignored modules are untracked strays from another working branch (`ImportError` on `docker_cli` / `_NEW_MESSAGE_BOUNDARY`); they do not exist on `main` and are unrelated to this change.

`ruff format` was **not** run — this tree is not format-clean and pre-commit runs `ruff-check` only (PR #327).

New coverage in `tests/test_subscription_credentials.py` (14 cases): the 2×2 auth-mode × harness matrix, fail-closed on missing and blank credentials, and three negatives proving a plan token cannot arrive via ambient env or `env_overrides` — only via `controlled`.


shanjing:
## Review — runtime side of `auth_mode`

Read the diff end to end, the companion PR (puffo-ai/cloud-infra#173), and the design doc, then checked the load-bearing claims against the code on both branches. Nothing here is a "don't merge". One finding is a real landmine on macOS, and one is a cross-PR wiring gap worth knowing before a template rebuild is spent on this.

### 1. macOS: the Keychain-deletion bug is no longer hypothetical

`macos/keychain.py:36` says verbatim *"We never set that env var"* about `CLAUDE_CODE_OAUTH_TOKEN`, citing anthropics/claude-code#37512 — the CLI `security delete-generic-password "Claude Code-credentials"` on exit via its fallback-combiner cleanup path. This PR starts setting it, and there is no platform guard.

E2B is Linux, so the cloud path is safe. The exposure is a `cli-local` agent on a Mac in subscription mode, where it would delete the operator's own login credential. `CLAUDE_CONFIG_DIR` / `HOME` redirection doesn't help — the Keychain item is user-wide.

Suggest refusing `auth_mode: subscription` under `is_macos()` (the design doc's finding 5 asks for exactly this scoping), and updating that docstring either way — it is load-bearing documentation that this change makes false.

### 2. Codex subscription is unreachable, not merely unexercised

The dispatcher routes `harness == "codex"` to `_prepare_codex_spec` (`local_runtime.py:395` → `:810`), which builds its child environment at `:861` with no subscription branch. The new resolver is called only from `_prepare_claude_spec` (`:747`), so `_codex()` is dead code today — and a codex agent with `auth_mode: subscription` silently runs the gateway path.

That means the fail-closed guarantee holds for **claude-code only**. The PR body is honest that codex is unverified; I'd scope the "fails closed" claim the same way, or raise in `_prepare_codex_spec` (and `_prepare_generic_spec`) when `auth_mode == subscription`, so the unsupported combination is loud rather than quiet. A raise costs three lines and preserves the property the PR is actually selling.

Same shape in `docker_runtime.py:323` — no `auth_mode` branch. Out of scope for the cloud (sandbox agents boot `kind: cli-local`), but the entire local fleet is `cli-docker`, so it would silently use the gateway if Stage 2 ever ran subscription locally.

### 3. The cross-PR wire is not connected yet

`_load_runtime_config` reads `runtime.auth_mode` from `agent.yml` and nowhere else. On the AIM side, `agent_store._agent_yaml` is the sole writer of the sandbox's `agent.yml` (`api_puffo.provision` → `materialise_agent_config` maps `agent.yaml` → `agent.yml`) and it emits no `auth_mode` line; separately, AIM's create path drops `auth_mode` from the request body. Details are in the review on cloud-infra#173 — flagging it here only because it means this PR cannot currently be exercised end to end: the token lands in the sandbox environment, `runtime.auth_mode` defaults to `api-gateway`, and `build_child_environment` strips the token as an uncontrolled provider credential.

Nothing to change in this PR for that; it's an argument for landing both halves together and asserting on the rendered `agent.yaml`.

### 4. Smaller things

**`subscription_token` (`state.py:473`).** `getattr(getattr(daemon_cfg, "subscription", None), "token", "")` — `DaemonConfig` has no `subscription` field, so that half always returns `""`, and the error text's "(daemon config / `CLAUDE_CODE_OAUTH_TOKEN`)" hint names a config key that does not exist. Either add the field or drop the branch and the hint. A getattr chain on a typed dataclass also means a future field rename fails silently rather than at import.

**"resolved once at startup" vs. where it's called.** The docstring says the daemon resolves the token once at startup, but the call is inside `_prepare_claude_spec`, so `os.environ` is read on every spec build (every refresh). It satisfies the AST invariant only because `portal/state.py` sits outside the package `test_child_env_allowlist.py` walks — the same shape as `claude_cli_api_key`, so it is consistent with the codebase; the difference is that one reads *config* and this one reads *ambient env*. Either soften the docstring or genuinely resolve at daemon start.

**`_codex` sets `CODEX_HOME` twice.** It goes into `env` (→ `controlled`) *and* into `extra_allowed`, while the `ChildCredentials` docstring says it "travels as an allowance rather than a controlled value". Redundant and harmless, but comment and code disagree.

**Layering.** `from ....portal.host_assets import _atomic_write_private` reaches across a package boundary for a private symbol, into the harness tree whose whole point is a narrow boundary. The function itself is correct — parent `mkdir(mode=0o700)`, `fchmod(0o600)` on the fd, `fsync`, atomic rename — so this is about promoting it to a public helper, not about the behaviour.

### What checks out

`ChildCredentials{env, files, extra_allowed}` is the right shape given codex's plan auth is file-based, and `extra_allowed` already existed in `build_child_environment`, so the codex branch slots in without widening the boundary — good call not to ship an env-only interface.

Both names in `PROVIDER_CREDENTIAL_ENV_NAMES` with three negatives (ambient, `env_overrides`, `controlled`) is exactly right, and `llm_env = dict(subscription.env)` **replacing** rather than merging is the correct read of the CLI's credential precedence — a leftover `ANTHROPIC_BASE_URL` would otherwise outrank the plan token silently.

Default-safety holds at every layer checked: `RuntimeConfig()` default, `_validate_auth_mode` on absent/empty, and AIM's `_build_install_bundle` `or "api-gateway"`. `agent_home / ".codex"` does equal `agent_codex_user_dir(agent_id)`, so `CODEX_HOME` agrees with the codex spec's own directory.

### What I did not verify

No tests were run. The codex `auth.json` shape I took as inferred, per the PR body. Whether E2B's `envs` survive a pause/resume — and therefore whether the daemon still holds the token on resume — I could not confirm from code.

shanjing:
Thanks — this was a good review. I verified each load-bearing claim against the code before acting; **findings 1, 2 and 4 were all real** and are fixed in the latest push. Finding 3 is fixed on the other PR.

**1. macOS — fixed, and you're right that it wasn't hypothetical.** My own design note asked for exactly this scoping and I never implemented it. `resolve_subscription_credentials` now raises `SubscriptionUnsupported` under `is_macos()` before touching the token, with the `#37512` reasoning inline. Two tests pin it, including that the platform refusal beats a missing credential — on macOS the answer is "never", not "set the variable and retry".

**2. Codex unreachable — confirmed and fixed.** You were right that `_codex()` was dead code: only `_prepare_claude_spec` called the resolver, so a codex agent with `auth_mode: subscription` ran the gateway silently. Rather than narrow the claim in the PR body, I took your alternative and made it true: `_prepare_codex_spec`, `_prepare_generic_spec` and `docker_runtime` all raise now. Three lines each, and the fail-closed property holds again for every path.

**3. Cross-PR wire — you found the blocker.** `agent_store._agent_yaml` emits the runtime block and had no `auth_mode` line, so puffo-agent defaulted every agent to `api-gateway` and `build_child_environment` stripped the token. Adding it to `_build_install_bundle` was not enough — that bundle feeds the *thin* runtime, while the fat agent here gets agent.yml from `_agent_yaml`. Two writers, one missed. Fixed in cloud-infra#173, plus recovery in `_YAML_TO_BUNDLE` so an edit doesn't silently move a subscription agent back onto the gateway. Both pinned by tests.

**4. Smaller things — all fixed.** The `DaemonConfig.subscription` getattr was dead and the error text named a config key that doesn't exist; both gone. The "resolved once at startup" docstring was false, and your read is right — this reads *ambient env* where `claude_cli_api_key` reads *config*; the docstring now says what it does and why it lives in `portal`. `CODEX_HOME` travels in `env` only. `_atomic_write_private` is promoted to a public `atomic_write_private`.

**On your "what I did not verify".** Whether E2B's `envs` survive pause/resume is the one open risk, and I can't settle it from code either. Resume restores the VM's memory state, so the daemon *should* still hold the token — but a paused-then-resumed subscription agent is now on the acceptance-test list rather than assumed.

```
$ ruff check src/ tests/test_subscription_credentials.py     All checks passed!
$ pytest tests/ --ignore=…docker_codex_wiring --ignore=…reanswer_boundary
4126 passed, 4 skipped
```

(`ruff format` deliberately not run — the tree isn't format-clean and pre-commit is check-only, per #327.)

shanjing:
## Post-merge re-review

Read the merged state on `main` (781d20d0). Every finding from the earlier review is addressed, and one of the fixes is better than what I suggested.

- **macOS** — `resolve_subscription_credentials` refuses under `is_macos()` (`subscription_credentials.py:110`), with `SubscriptionUnsupported` as its own error type. The `keychain.py` "we never set that env var" note is true again because the code makes it true, which is the right direction to resolve that contradiction.
- **Unimplemented paths refuse instead of falling back** — `_prepare_generic_spec:411`, `_prepare_codex_spec:825`, `docker_runtime:316`. "Fails closed" now describes the whole surface rather than one branch of it.
- **`subscription_token`** — the dead `daemon_cfg.subscription` chain is gone, replaced by `del daemon_cfg  # reserved` and a docstring that says what it actually does. The overclaim about resolving once at startup is gone too.
- **`atomic_write_private`** — public alias in `host_assets.py`, no private symbol crossing the boundary.

The `daemon.py` `_is_subscription` predicate is the part I'd call out. I flagged the `llm_base_url`-shaped "is this gateway mode?" test as an accidental coupling; the fix names a consequence I had not traced — with #173 correctly omitting `llm_base_url` for subscription agents, the old predicate would have routed them into the OAuth refresh path, where they'd hunt for a host credential view that does not exist in a sandbox and block the turn. The two changes were only safe together, and checking the mode before the side effect is the reason it holds.

Two residual notes, neither blocking and neither needing a follow-up on its own:

**`_codex()` is now unreachable by construction.** Both `_prepare_codex_spec` and `docker_runtime` raise before it is called, so the codex branch and its tests describe an intended contract rather than live behaviour. That is the right shape while the `auth.json` path is unverified — worth one line in the module docstring saying so, since a reader who finds `_codex()` and its passing tests could reasonably conclude codex subscription works today.

**`CODEX_HOME` is still set twice** — in `env` (which becomes `controlled`) and in `extra_allowed`, while the `ChildCredentials` docstring says it travels as an allowance rather than a controlled value. Harmless and redundant; the comment and the code just disagree about which channel carries it.

Nothing else from the earlier pass survives. Good change.

PR #349, opened 2026-09-11
Import LingTai Agents and synchronize authoritative names

Glimmer0x:
Import existing LingTai Agents using authoritative source names, then keep those names synchronized while the connected Puffo daemon runs. Ordinary Puffo creation remains separate.

Discovery and manual validation reject corrupt/quarantined source metadata and preserve unnamed profiles. Executable-folder discovery is bounded and does not execute search candidates. A daemon-owned publisher checks active bindings after each 30-second interval, sends only display_name, retries failed updates, and never writes LingTai source files. Runtime synchronization uses the current .agent.json only: missing or corrupt files and missing name fields preserve the last published name; explicit null clears it. Import discovery retains its init.json fallback compatibility.

Imported profiles remain source-managed. Control edits reject changes; CLI rename/profile and Qt editing now also refuse them before writing. Startup/warm snapshot sync skips LingTai to avoid restoring stale names. The acceptance document lists each entry point, evidence, and remaining integration checks: docs/engineering/lingtai-name-sync-acceptance.md.

Validation: relevant tests and Python structure checks pass, and the full daemon suite exits successfully with inherited CODEX_HOME removed (two environment-specific skips). Qt controls were tested directly for read-only fields, disabled Save, and handler rejection. Prior named/unnamed imports completed two real model/tool/result/reply turns. Live owner and non-owner chat and profile pages automatically receive ordinary name updates; server-null clear/restore passed on both accounts, with explicit null in both raw caches. Requires puffo-ai/puffo-server#370 for explicit name clearing. No merge or deployment.

Rollout requirement: deploy server PR #370 before this daemon change. Older servers silently ignore display_name:null while reporting success, which would let the daemon incorrectly cache a completed clear. See docs/engineering/lingtai-name-sync-acceptance.md. Boris reran the combined server 18d3538 lib suite: 1429 passed, 0 failed, 3 ignored.

CI: all checks passed on c02d080e (pre-commit and Python 3.11/3.12 test suites).

### DIFF
--- a/docs/engineering/lingtai-name-sync-acceptance.md
+++ b/docs/engineering/lingtai-name-sync-acceptance.md
@@ -1,0 +1,36 @@
+# LingTai 名称同步与编辑边界验收
+
+截至 2026-09-12；维护者 @peter-stei-5525-5d5ebaa5，服务端协作者 @boris-cher-7418-1578233d。
+
+## 用户确认的要求
+
+- 灵台修改 NAME 后，Puffo 自动反映；描述可以不显示。
+- Puffo 不提供会被下一次来源同步覆盖的改名入口。这是必需关闭项。
+- 保持来源权威和单向性：无名仍为 null，“未命名”只用于显示。
+
+## 实施与证据清单
+
+| 项目 | 当前证据 | 状态 |
+| --- | --- | --- |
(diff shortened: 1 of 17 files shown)


Glimmer0x:

PR #350, opened 2026-09-11
feat(agent): reintroduce Monid data tools behind a default-off flag

KurtneySh:
## Background
Agents need tools to look up and pay for data via the paid-data service (Monid). These tools existed previously and were reverted; this reintroduces them, now routed through the billing spend gateway.

## What changed
- Reintroduces two tools: one to discover a capability and its price (free), one to run a paid lookup. The paid tool calls the billing-backed spend endpoint; **the agent never holds the payment key or a spend token** — the server holds the key and mints the token internally.
- Registration is gated on a default-off flag (`PUFFO_MONID_TOOLS_ENABLED`); with the flag off the tools are not registered at all, so production is unchanged.

## Scope
Agent tool layer only.

## Not-ready-to-merge (draft)
The paid tool depends on the server-side spend endpoint, which is still in review; until that ships the flag stays off. Merge to production stays gated on that server-side review.

## Verification
Structure check + `ruff` + full pre-commit + full test suite green; the tool's request shape was verified field-by-field against the server's spend request contract.

## Release note
Internal — flag-gated, not user-visible.


hellohanchen (commented):
## Round 1 — independent review (requested by @mingvase-8795)

**Pinned head:** `a2398d30a1b1a3db5ffff17cec08b7e92618c467` (draft), base `main` = `bfa61a59` (current tip; base is an ancestor, additive-only diff, +1361/−0). All 9 hosted checks green. Local: `tests/test_monid_tools.py` 25/25 against this checkout (`PYTHONPATH=src`, py3.12 venv).

**Verdict: 0 blocker / 1 major / 5 minor.** The tool code is careful (fail-closed mint, https-only Bearer sink, namespaced idempotency keys, 4xx-release / 5xx-retain, pending/replay rendering) and the money-safety mechanisms are pinned by tests. The one thing I'd fix before this leaves draft is that the PR text and CHANGELOG describe a different architecture than the code ships (M1).

### Mutation run (one at a time, `tests/test_monid_tools.py` + `tests/test_auto_accept_dm.py`, tree restored + verified clean)
| mutant | result |
|---|---|
| P2 register for keyless agents too | **KILLED** |
| P4 drop the https check on `billing_base_url` | **KILLED** |
| P5 release the automatic key on 5xx too (`ambiguous = False`) | **KILLED** |
| P6 `finish()` retires the key even when the spend is `PENDING_RECONCILE` | **KILLED** |
| P7 explicit `idempotency_key` not namespaced by slug | **KILLED** |
| P1 `monid_tools_enabled()` accepts any non-empty value instead of exact `"true"` | **SURVIVED** |
| P3 `post_bearer` without the https check | **SURVIVED** |
| P3b `post_bearer` also puts the token in the query string | **SURVIVED** |
| P8 mint the token *after* `key_for()` (allocates retry state before the mint can fail) | SURVIVED (behaviour-neutral today; noted) |
| P9 `build_server` passes `monid_tools_enabled=True` regardless of env | **SURVIVED** |

### Findings

**M1 — the PR description and CHANGELOG do not describe the code.** PR body: *"the agent never holds the payment key or a spend token — the server holds the key and mints the token internally"*; CHANGELOG: *"through the server's monid spend gateway, which holds the key…"*. The code (since `ceedfd3`) does the opposite for the paid hop: `monid_spend` mints a spend JWT from puffo-server (`_fetch_spend_token`, `core_monid_tools.py:319-331`, `POST /v2/agent/spend-token`) and then POSTs **directly to billing** at `<billing_base_url>/v2/monid/spend` with `Authorization: Bearer <token>` (`:215-231`, `http_client.py:311-337`). The module docstring (`:11-19`) says this correctly ("calls puffo-billing directly"). For an auth/money review the difference is material: the agent process *does* hold a short-lived money credential in memory and sends it to whatever `billing_base_url` the mint returns. Please make the PR body and the Unreleased changelog entry match the two-hop design (mint at puffo-server → spend at billing), and say the token is ≤ 1 h (server-capped) and never persisted.

**m1 — env-flag parsing is unpinned (P1).** `monid_tools_enabled()` (`config.py:70-72`) is the only place the exact-`"true"` rule lives; every test passes `monid_tools_enabled=` into the config directly. `PUFFO_MONID_TOOLS_ENABLED=1` silently enabling the tools would pass CI. One `monkeypatch.setenv` test each for `"true"`, `"1"`, `""` closes it.

**m2 — `post_bearer` itself is never executed by a test (P3, P3b).** The fake in `test_monid_tools.py:24` replaces it, so the https refusal, "token in `Authorization` only", no-subkey-signature and no-`x-sandbox-token` properties are asserted only by the docstring. Pin them with an `aiohttp` test server (the repo already has that pattern) or by mocking `_healed_request` and asserting the exact `(method, url, headers)` tuple.

**m3 — wiring is unpinned (P9).** `puffo_core_server.py:368` and `ws_local/route.py:208` are the only callers of `monid_tools_enabled()`; hard-coding `True` there passes everything. A `build_server`-level test asserting `monid_*` is absent from `list_tools()` with the env unset (and present with `"true"`) would prove the gate end to end rather than at the helper.

**m4 — non-`HttpError` failures on the billing hop skip the retry guidance.** `post_bearer` is wrapped only for `HttpError` (`:224-228`); an `aiohttp.ClientError` / timeout propagates raw. Key state is retained (safe — the retry reuses it), but the model gets no "retry the same arguments to reuse key …" text and no non-Monid labelling instruction. Catch `Exception` there (or `aiohttp.ClientError`, `asyncio.TimeoutError`) and route through the same ambiguous path.

**m5 — both server dependencies are unmerged, not just the spend one.** puffo-server `main` (`b646a60c`) has no `/v2/monid/prepare` route at all and the spend-token mint is puffo-server #366 (in review). With the flag off nothing is reachable, so this is fine for a draft; just noting the PR body's "the paid tool depends on the server-side spend endpoint" understates it — `monid_prepare` 404s today too.

Cross-reference for @mingvase-8795: this PR is the concrete consumer of puffo-server #366's `/v2/agent/spend-token` — a **native** (subkey-holding) agent; keyless/bridge agents are explicitly excluded here (`register_monid_tools`, `:99-110`). That answers #366's M1 (reachability): the "cloud agent" wording in #366's router comment is what's off, not the design.

Review only — not merging.

PR #351, opened 2026-09-11
feat: seed a cloud agent's memory from the fleet remote on first boot

shanjing:
## What

A cloud agent is created with an identity but an **empty brain** — the Hub mints its keys, its persona and memory live elsewhere. Today that gap is closed by hand after every create, and an agent messaged before the copy answers from nothing, which reads like a memory bug rather than a missing step.

Two halves:

**`agent/memory_seed.py`** — on start, an agent whose config names a `memory_remote` clones `<repo>/<name>/{memory,profile.md}` into itself.

**`tools/seed_cloud_agent.py`** — publishes a local agent into that remote.

```
1. create the cloud agent in the Hub, choosing its billing mode
2. ./tools/seed_cloud_agent.py --from desk --repo <fleet> --push
3. the agent seeds itself on its next boot
```

## It seeds; it never syncs

The clone happens **only when the memory tree is empty**. An agent that has written anything is left alone, unconditionally.

That single rule is what makes this safe to run on every boot: there is no state in which it can overwrite a note the agent wrote, and no flag anyone has to remember to unset. The load-bearing test is the negative one — `test_never_overwrites_an_agent_that_has_written`.

A related subtlety with its own test: `~/memory` **is itself a git repo**, so `.git` must not count as content. Counting it would make every agent look seeded and silently skip the seed.

## Failure is never fatal

An unreachable remote, a missing directory, or a broken key returns a status string and logs a warning. The agent starts with an empty memory instead of not starting. A boot-time convenience must not be able to take an agent down.

## The CLI is not a migration, and says so first

No identity is exported, rotated, or revoked; the local agent is only ever read. `keys/` (identity), `agent.yml` (identity + a gateway credential), `workspace/` (reproducible, up to 109 MB), `Library/` (leaked host state) and `messages.db` (server-held) are deliberately left behind. The tool that *does* migrate is `portal/import_agents.py` — a different, harder operation behind a different decision.

Two behaviours come from measuring a real 62-directory fleet rather than assuming its shape:

- **An agent needs both `profile.md` and `agent.yml`** — half the directories are empty scratch dirs, and a bare glob would publish 31 of them.
- **Two slug shapes exist** (`desk-6332-b73d5e96`, `planner-2d35d73e`), both shortening to the name a human uses — a cloud agent gets a fresh slug on every create, and its memory should follow the agent rather than the identity it happens to be wearing.

## No `--token`

The subscription credential is a property of the **deployment**, not of an agent: it reaches a sandbox as an environment variable set once by the provisioner. A per-agent flag would imply per-agent tokens — more copies of a credential, more places to revoke — and would have this CLI write a secret into a sandbox, which is worse than leaving it where it is.

## Not in this PR

The provisioner must set `memory_remote` on the agent it creates, or nothing seeds. That is a small AIM change in cloud-infra and belongs with the terraform that would configure the remote.

## Passing-Evidence

```
$ pre-commit run --files <changed>        all hooks passed
$ pytest tests/                            4230 passed, 4 skipped
$ seed_cloud_agent.py --all --dry-run      31 agents found, 31 scratch dirs skipped
```

(Two untracked test modules from another working branch are excluded, as on `main`.)


shanjing:
## Review

Read the diff end to end plus the two backend PRs this follows (#348, cloud-infra#173). The seed-never-sync rule is the right invariant and the reason to prefer it — no flag to unset, no state where it can overwrite — is stated where the code lives, which is where it belongs. `.git` not counting as content is the subtle one and it has the test it deserves. Splitting `cloud_verify` into pure checks over collected facts is a good call; the checks are the most valuable artifact here and they outlive the CLI that calls them.

Findings below, roughly in order of how much they'd cost to discover the hard way.

### 1. `bridge_connected` passes in exactly the case it exists to catch

```
echo "bridge_connected=$(ss -tn 2>/dev/null | grep -cE 'ESTAB' )"
```

That counts *any* established TCP connection. A subscription agent with 24 sockets open to Anthropic and a dead relay reports `bridge_connected=1`. The check's stated purpose is "an agent that can think but not speak" — this is the shape of the `BAD_FRAME` failure that ended Option A, and the probe would have called it healthy.

Resolve the relay host and count against it, the way `conns_gateway` already resolves `litellm-staging.puffo.ai` — or key off the log (`bridge error category=` vs a connected event), which is what actually distinguished the two states during the build-out.

### 2. `conns_anthropic` will report zero on a healthy agent

```
conns_anthropic=$(ss -tn | grep -c '160.79.104')
```

A hard-coded IPv4 prefix. Phase 0 in the plan doc resolved `api.anthropic.com` to `2607:6bc0::10` — an IPv6 connection matches nothing here, so `_talks_to_the_right_upstream` fails a correctly-working subscription agent, and the operator's next move is to go looking for a bug that isn't there. Resolve the host like the gateway branch does, and count both families.

### 3. The probe echoes the real virtual key

```
echo "api_key=$(awk ... /^ *api_key:/ ...)"
```

The value, not a boolean. It travels back over the E2B command channel and lands in `facts["api_key"]`, where the only consumer asks `not key`. Nothing prints it today, but a debug print of `facts` — the obvious thing to do when a check reports `????` — dumps a live gateway credential. Emitting `api_key_present=0|1` costs nothing and keeps this consistent with how carefully the rest of this work handles secrets.

### 4. "Failure is never fatal" has two holes

`seed_from_remote` catches `OSError` plus the two subprocess errors. Two paths escape:

- `src_profile.read_text(encoding="utf-8")` raises `UnicodeDecodeError` on a non-UTF-8 profile in the remote. That's a `ValueError`, not an `OSError`, so it propagates out of `_prepare_paths` and the agent doesn't start.
- `tempfile.mkdtemp(...)` is *above* the `try`, so a failure there propagates too.

Both are exactly the "a boot-time convenience must not be able to stop an agent from starting" case. `errors="replace"` on the read, and moving `mkdtemp` inside (or widening to `except Exception`) closes it. Worth noting `shutil.Error` *is* an `OSError` subclass, so `copytree` is genuinely covered — that one's fine.

### 5. A 120-second clone sits on the boot path

`_CLONE_TIMEOUT_S = 120`, synchronous in `_prepare_paths`. Only the first boot pays it (`memory_is_empty` short-circuits after), but the first boot is precisely when someone just created the agent and is waiting to message it. The doc's own measurement for a full clone round-trip in-sandbox was 7.4s. Something in the 20–30s range would still be generous, and would fail fast enough to look like a seeding problem rather than a hung agent.

### 6. The clone-failure log line can carry a credential

The comment says the stderr tail is kept short "so a token never lands in a log line", but truncating to 120 characters doesn't remove one — git commonly echoes the remote in failure messages, and `https://user:token@host/repo` survives truncation intact. The intended config is an SSH deploy key, where this can't arise, but `memory_remote` is free text from config. Scrubbing `://…@` (or logging only the exit code) matches what the comment already claims.

### 7. The UNKNOWN discipline isn't applied in two checks

The module opens by saying "I could not look" is not "it is absent", and conflating them turns a broken collector into a clean bill of health. Then:

- `_no_virtual_key` — `if _fact(f, "auth_mode") != "subscription": return True`. An unread `agent.yml` makes `auth_mode` UNKNOWN, which is `!= "subscription"`, so the check reports **ok** with "not a subscription agent — a key is correct here".
- `_credential_reaches_the_cli` — same shape: UNKNOWN mode falls into the gateway branch and is evaluated as if it were one.

Both should return `None` when the mode is UNKNOWN. As written, the failure that mattered most during the build-out (`auth_mode` stuck on gateway because nothing wrote the field) is also the one that makes these two checks stop being able to see it.

### 8. Two copies of the stub threshold

`_STUB_PROFILE_BYTES = 512` in `memory_seed.py`, and a bare `if n <= 512` in `cloud_verify._profile_is_not_a_stub`. Same constant, same PR, two places.

Related, and a real if small data-loss path: a user who types a *short* soul at create gets a profile under 512 bytes, which the seeder classifies as a stub and overwrites with the remote's. Comparing on "has a non-empty Soul section" — which `soul_body()` in this same PR already knows how to compute — would distinguish "the Hub's empty stub" from "the operator wrote two sentences". At minimum, log when a non-empty profile is replaced.

### 9. It seeds, but nothing syncs — so durability is still open

Scoping this to seed-only is defensible and the PR is honest about it. The thing worth saying out loud: the plan doc marks criterion 3, "memory survives sandbox death via git remote", **done** — proven by a spike that cloned *and pushed*. What ships here is the clone half. Notes an agent writes after seeding still die with the sandbox, and the copy is `shutil.copytree` out of a temp clone, so the agent's memory tree has no remote to push back to.

Not a request to widen this PR. It is a request not to let the doc's green checkmark be read as covered by it.

### 10. `memory_remote` needs the same two writers `auth_mode` needed

The PR says the provisioner must set it, which is right. Naming where, because this exact pair was missed for `auth_mode` and made that feature inert end to end: `agent_store._agent_yaml` (the sole writer of the sandbox's `agent.yml`) **and** `_YAML_TO_BUNDLE` (so it survives an edit round-trip). Adding it to `_build_install_bundle` alone is the trap that already caught us once.

### 11. `BAD_FRAME` is being encoded as expected behaviour with no ticket behind it

`test_known_bad_frame_noise_does_not_fail_the_bridge_check` and the `"known list_invites noise"` string in `_bridge_connected` are, as far as I can tell, the only durable record that puffo-server rejects those frames. Tolerating it in the check is right — it demonstrably doesn't stop an agent working. Encoding it as permanently-expected without a ticket reference means the next person to see 31 `BAD_FRAME`s finds a test comment telling them it's fine, and no way to find out whether it still is. Worth filing and referencing the id here.

### 12. `verify()` touches every running sandbox

It lists running sandboxes, then `Sandbox.connect` + `commands.run` on each until the slug matches. On staging that's a handful of other people's agents getting a shell command run in them to answer a question about a different agent. AIM already holds the `agent_id → sandbox_id` binding; a `--sandbox-id` flag, or a lookup, avoids the scan entirely.

Two smaller things in the same file: `facts["template_id"] = getattr(sandbox, ...) or UNSET` in `collect()` is immediately overwritten by `facts["template_id"] = info.template_id` in `verify()`, so the first is dead; and `UNSET` is defined below its use (fine at runtime, surprising to read).

### 13. `tools/seed_cloud_agent.py` has no tests

`test_memory_seed.py` and `test_cloud_verify.py` cover the runtime halves well. The CLI — `soul_body`, `discover`, `resolve_one`, `stage`, `agent_name` — has none, and the evidence for it is a manual `--all --dry-run`.

cloud-infra#174 ships that exact set of functions with 22 tests (`tools/tests/test_import_local_agent.py`). If #174 is closed as superseded, port that module across first: it's the same logic, and it would move from a tree cloud-infra's CI doesn't run into one that does. Detail in the review on #174.

### Worth keeping as-is

`memory_is_empty` excluding `.git`, with the reasoning recorded — counting it would make every agent look seeded and silently skip forever, and that's a bug that would have been very hard to see. `memory_seed_name` handling both real slug shapes, derived from a survey rather than assumed. `discover` requiring both `profile.md` and `agent.yml` because half the fleet is scratch directories. Refusing `--token`, and saying why. And `evaluate` running every check rather than short-circuiting, because the pattern of failures is the diagnosis — that's the right shape for something an operator reads at 1am.

PR #364, opened 2026-09-13
fix: keep Windows background daemon console hidden

Glimmer0x:
On Windows, `start --background` and `start --detach` invoke the active Python executable, which can be a uv/venv redirector. With `DETACHED_PROCESS`, its second launch creates a visible console for the real daemon. Closing that window ends the service and disconnects the machine.

Use `CREATE_NO_WINDOW` instead of `DETACHED_PROCESS` so the redirector and real interpreter use a windowless console. Preserve process grouping, job breakaway, log redirection, and POSIX behavior.

Validation:
- Native Windows 3.11/3.12 regression creates a real venv and checks the real interpreter has no visible console while retaining its environment and log output. Both versions failed on the old implementation at the visible-console assertion (run 34751788054).
- Fixed native Windows 3.11/3.12 and all full remote CI passed on 54123910754c8cd700ee55297adb3b3e833944fe (run 34751889247).
- Local full suite: 4443 passed, 3 conditional/platform skips. All pre-commit checks passed; independent source review found no blockers.

The Windows hosted runner rejects job breakaway, so the native test disables that flag only in the test to isolate console inheritance. Production keeps it, and the flag contract remains covered separately. This does not claim an interactive terminal-close or restricted-job lifetime acceptance. The initial runner access-denied failure is not counted as reproduction evidence.


Glimmer0x:
Superseded by #365, merged as cab6f62. Its pythonw-based launch preserves the venv and includes the broader shutdown/archive fixes plus Windows CI and native uv/venv validation. Keeping one launch implementation avoids overlapping changes.

PR #374, opened 2026-09-14
feat(codex): subscription auth_mode on the codex spec path (PUF-396 §4c)

shanjing:
## What
Staging e2e 2026-09-14: the Hub stored a Codex plan credential, AIM's vault delivered it as `CODEX_SUBSCRIPTION_AUTH_JSON`, and the worker died at startup:

```
SubscriptionUnsupported: runtime.auth_mode='subscription' is not supported by the 'codex' runtime path yet; only claude-code implements it.
```

`_prepare_codex_spec` refused the mode before touching the credential, although `resolve_subscription_credentials("codex", …)` already knew how to place the document. This PR gives the codex spec the same branch the claude-code spec has:

- resolve the `auth.json` document from the daemon's environment, write it at 0600 under the agent's `CODEX_HOME` (config.toml already pins `cli_auth_credentials_store = "file"`)
- hand the child that `CODEX_HOME` and no `OPENAI_API_KEY`
- a configured gateway (`llm_base_url` + `api_key`) is **dropped, not merged** — codex prefers a provider block over its own OAuth, so keeping it would bill the metered account in silence (same rule the claude spec applies to `ANTHROPIC_BASE_URL`)
- the host auth view is never consulted for a plan agent (a sandbox has no operator login); the refresher already skips subscription agents

## Passing evidence
```
pytest tests/test_codex_subscription_spec.py tests/test_subscription_credentials.py tests/test_task_timeout_knob.py   → 26 passed
pytest -k "codex or local_runtime or subscription or child_env" (minus two files that fail collection on main in this venv) → green
ruff check → All checks passed
```

Follow-ups (cloud-infra): build `puffo-codex-v2-2` / `puffo-claude-v2-2` from this ref and drop staging's `AIM_TEMPLATE_OVERRIDE`, which currently pins every create to a claude-only image.


shanjing:
Rebased onto `main` (`6cf58199`, 52 commits; only CHANGELOG.md conflicted — both entries kept). Tests on the rebased branch: `test_codex_subscription_spec` + `test_subscription_credentials` + `test_task_timeout_knob` → 26 passed, ruff clean.

Why it matters now: today's `puffo-codex-v2-7` was cut from `main`, which still raises `SubscriptionUnsupported` for a Codex plan agent — so `codex:stable` on staging was rolled back to `puffo-codex-v2-2` (has this branch, lacks #402) until this merges and a new codex template is cut.

shanjing:
> **Soft review — advisory, not a gate.** Findings, evidence and reasoning; not a
> change request. Yours to weigh and decide.

## Review

This closes the gap I raised on #348, where `_codex()` was unreachable by construction: `_prepare_codex_spec` refused the mode before anything could call the resolver, so the codex branch and its tests described an intended contract rather than live behaviour. Now the branch runs, and the staging failure in the body is what that dead code looked like from the outside.

The shape is right, and three things I went to check turned out already correct:

- **`sync_host_codex_auth_view` is in the `else` branch**, so it never runs for a plan agent. That was the clobber I expected to find — a host view refresh after the write would have overwritten `auth.json` with the (absent) operator login, and the symptom would have been "codex ignores the credential" rather than an error.
- **`cli_auth_credentials_store = "file"` really is unconditional** — `mcp/config.py:198` says *"Always emits"* and `:212` writes it. The comment's claim that "config.toml pins it, so that is where codex looks" holds for every branch, not just this one.
- **Dropping the gateway is done in the right place.** `gateway = None if subscription` flows into `config_kwargs["provider"]`, so the provider block is absent from `config.toml` itself rather than merely unset in the environment. Given codex prefers a configured provider over its own OAuth, removing it from the file is the half that actually decides this.

Also worth noting: `_codex()` no longer returns `extra_allowed=("CODEX_HOME",)` while putting it in `env` — the contradiction from #348 is gone, and the new docstring explains why it travels in `controlled`. Good cleanup.

Two things.

### 1. `CODEX_HOME` is derived twice, and the second one silently wins

```python
codex_home = agent_codex_user_dir(self.agent_id)          # spec: agent_home_dir(id)/".codex"
...
controlled: dict[str, str] = {"CODEX_HOME": str(codex_home)}
creds = resolve_subscription_credentials("codex", agent_home=self.agent_home, …)
controlled.update(creds.env)                               # resolver: agent_home/".codex"
```

`_codex()` computes `agent_home / ".codex"`; the spec computes `agent_codex_user_dir(agent_id)`, which is `agent_home_dir(agent_id) / ".codex"`. They are the same path today, so the `update` is a no-op overwrite and everything works.

The failure mode if they ever diverge is quiet and ugly, which is why I would not leave it implicit:

- `config.toml` and `AGENTS.md` are written to the **spec's** `codex_home`;
- `auth.json` is written to the **resolver's**;
- `CODEX_HOME` ends up pointing at the **resolver's**, because `update` runs last.

So codex would find its credential and not its config — no provider block, no MCP servers, no `cli_auth_credentials_store` pin. An agent that authenticates and then cannot post is the PUF-377 shape, and nothing here would say why.

Cheapest fix: have `_codex()` take the already-computed `codex_home` instead of re-deriving it from `agent_home`, so there is one source. Failing that, `assert creds.env["CODEX_HOME"] == str(codex_home)` costs a line and turns a silent split into a loud one.

### 2. The macOS refusal is harness-blind, and its reason cannot apply to codex

`resolve_subscription_credentials` raises before dispatching on harness:

```python
if is_macos():
    raise SubscriptionUnsupported(
        "…: setting CLAUDE_CODE_OAUTH_TOKEN makes the CLI delete the operator's …"
    )
```

That guard exists for a real and specific bug — anthropics/claude-code#37512, where setting `CLAUDE_CODE_OAUTH_TOKEN` makes the *Claude* CLI delete the user's Keychain credential on exit. None of it is reachable through the codex path: the credential is a file under `CODEX_HOME`, that env var is never set, and no Keychain entry is touched.

Two consequences. A codex plan agent on macOS is refused with an error about a Claude-only failure, which will read as a bug in the message rather than a policy. And more practically, **the path this PR adds cannot be exercised outside a sandbox** — a developer on a Mac cannot run it locally at all, so the only way to test it is the staging e2e that prompted the PR.

Either scope the guard to `harness == "claude-code"`, or keep it and say in the message that codex inherits the refusal for a different reason (if the reason is "we have not verified codex plan auth on macOS", that is a fine reason — it just is not the one the text gives).

### Smaller

- The body's *"the refresher already skips subscription agents"* checks out: `daemon.py::_is_subscription` keys off `runtime.auth_mode` before the `llm_base_url` heuristic, which is exactly why it holds for codex too.
- Fail-closed works by inheritance — a blank `CODEX_SUBSCRIPTION_ENV` raises `SubscriptionCredentialMissing` from `_codex()` before the child launches. Worth one test asserting that from the *spec* path rather than only from the resolver, since that is the boundary the staging failure crossed.

PR #379, opened 2026-09-17
fix(agent): correct the model instead of binning its answer (PUF-400)

shanjing:
Fixes **PUF-400**. Found while verifying the PUF-393 channel-wake fix on staging: the wake works, the agents think, and two of three answers were thrown away by their own runtime.

## The issue
A channel message reaches an agent as a **global-inbox notice**, and `handle_global_inbox_turn` passes `allow_plain_fallback=False`. That flag is correct — a notice can name several targets, so there is no single destination for loose prose. What was wrong is what happened next: when the model answered in prose instead of calling `send_message`, `_route_turn_result` logged `[no-send]` at **INFO** and returned `None`. The answer vanished, with no signal to the human, the operator or the sender.

From the channel that is indistinguishable from an agent ignoring the room, and indistinguishable from PUF-393 before it was fixed — which is exactly how it stayed hidden.

## Evidence (staging, 2026-09-17, a three-agent channel, one question addressed to "all")
| Time (UTC) | Event |
| -- | -- |
| 05:19:23 | `wake PutEvents delivered woken=3 failed=0` |
| 05:19:25 | all three write `PUT /v2/messages/cursors/channel/…` — they read it |
| 05:19:29 | all three subscribe and fetch the channel roster |
| 05:19:40 | `desk-iv` logs `[no-send]` **and still answers** |
| 05:19:47 | `sentiment--2369` → `[no-send]`, nothing posted |
| 05:19:52 | `dex-cloud-6888` → `[no-send]`, nothing posted |

```
[INFO] puffo_agent.agent.core: agent <slug>: [no-send] [global inbox] @:
       ignoring plain assistant output; outbound chat requires send_message
```

Runtimes 2.0.4 and 2.0.5a4 — the newest of the three failed and the one that succeeded logged the same line on a different turn, so this is per-turn, not per-version. Deleting and recreating the agents would not have helped.

## The fix
One corrective ask through the existing `run_retry_turn` seam (the same machinery the rate-limit retry uses), naming the tool and the eligible targets:

> Your last reply was plain text, so it was NOT delivered — nobody saw it. … Call `mcp__puffo__send_message` now for: `sp_axe/ch_general`. Send the answer you already wrote. If you meant to stay quiet, reply with `[SILENT]` instead.

A model that already has the answer almost always emits the call on the second ask, and the `send_message`-only invariant — which carries the target, the thread and the human-visibility flag — is **preserved rather than bypassed**.

Deliberately **not** done: flipping `allow_plain_fallback` to true. With several targets there is no destination, and posting raw assistant text is what leaked literal `send_message(...)` syntax into chat on 2026-07-09.

If the correction also fails, the give-up is loud: `WARNING` with `error_category=plain_output_undelivered` and a reporter event carrying a preview. An answer that could not be routed never disappears without a trace again.

## Tests
- corrected then delivered — exactly one ask, kick names both the tool and the target;
- corrected and still plain — warned, categorized, reporter event, nothing posted;
- `[SILENT]` on the second ask — a decision, not a failure, no warning;
- a turn that called `send_message` — never pays for a retry.

The pre-existing "plain output is internal, not an implicit message" test keeps its assertions unchanged; its stub gains the `run_retry_turn` seam it predates.

## Harness
`ruff check` and `ruff format` clean. **4678 passed, 15 skipped, 0 failed.** `tests/test_docker_codex_wiring.py` and `tests/test_reanswer_boundary.py` fail to collect on `main` as well (missing `docker_cli` and `_NEW_MESSAGE_BOUNDARY` imports) and are excluded; they are unrelated and pre-existing.

## Risk
The only new cost is one extra turn on a path that previously produced nothing, and it fires only when an answer would otherwise have been discarded. The happy path is untouched and asserted to be.

## Seen alongside, unticketed
All three agents log `bridge error category=BAD_FRAME` once every 60 seconds, continuously. Unrelated to this change; nobody is watching it.

https://claude.ai/code/session_01FB5KJuGrzwZ3BhpXFXPutm


shanjing:
[agent:claude] Self-review pass — no external review arrived (the Codex bot reported it had hit its usage limits), so I reviewed my own diff against the gateway checklist and found two things worth fixing, both mine.

**1. I reformatted a file I did not change.** Running `ruff format` over `tests/test_global_inbox_runtime.py` rewrote 426 lines of unrelated code, which is exactly the "behaviour, refactor and formatting mixed into one unreviewable diff" the code-delivery gateway calls a blocking failure. Reverted to `main`'s content; that file now carries **10 added lines** — the retry seam its pre-PUF-400 stub needs — and nothing else.

**2. My tests pushed that file past the repo's 2000-line structural limit.** They now live in `tests/test_plain_output_correction.py`, named for the behaviour rather than buried. Better placement regardless of the limit.

Fixed in `da7a83a9`. Structural violations attributable to this PR went 2 → 0.

**On the remaining red check, and why I am merging anyway.** `pre-commit` still fails with exactly one violation: `src/puffo_agent/agent/puffo_core_client.py:1: file is 2003 lines (limit 2000)`. That file is **byte-identical to `main`** here — it crossed the limit in #378 yesterday, and `main`'s own CI run (`6564b201`) is red for it. It is not caused by, related to, or fixable within this change, and merging does not make `main` worse than it already is. Waived with that reason, per the release gateway's "green or waived with a reason".

Everything else is green: pytest on 3.11 and 3.12, Windows lifecycle on both, CodeQL, all three Analyze jobs. Locally: **4678 passed, 0 failed**, ruff check clean.

**Separately**: someone should decide whether to trim `puffo_core_client.py` under the limit or raise/split it. It is three lines over and it blocks the gate for every PR in this repo, not just this one. I did not do it here because unrelated edits to a core client file are precisely what this PR should not contain. Happy to send it as a one-line PR of its own if you want it.

PR #384, opened 2026-09-17
fix(agent): introduce yourself when added to a space (PUF-402)

shanjing:
Fixes **PUF-402**. A local agent introduces itself when added to a space; a cloud agent arrives silently. It is in the roster and answers when spoken to, but the room never learns who joined.

## The issue
**The introduction is a passenger on the client-side accept.** A local agent accepts its own invite, and that routine carries the nudge:
```
puffo_core_client.py:1024   accept_invite(..., enqueue_intro=self._enqueue_channel_intro_nudge, ...)
```
A cloud agent never runs it. Because `agent_status.machine_id = "puffo-cloud"`, puffo-server auto-accepts inside the same transaction as the invite (`spaces.rs:1218-1229`) — deliberately, since a paused agent would otherwise leave every invitation hanging — so the agent is a member before its sandbox is consulted.

The one fallback route is `_handle_channel_accept_event` (`puffo_core_client.py:442`), which needs an `ACCEPT_CHANNEL_INVITE` carrying `original_invite`. The server's two auto-accept builders are asymmetric:

| builder | payload |
|---|---|
| `build_auto_accept_synthetic_event` (**channel**) | `space_id`, `channel_id`, invite id, nonce, **`original_invite`** |
| `build_auto_accept_synthetic_space_event` (**space**) | `space_id`, invite id, nonce — **no `original_invite`, no `channel_id`** |

So inviting a cloud agent to a *channel* directly would introduce it, while adding it to a *space* — the normal path, the one the auto-accept exists for — never does. The channel cascade does not help either: the space-accept arm inserts membership rows, and rows are not events.

## The fix
`added_to_space` is the one join signal a cloud agent actually receives, and its handler already refreshes the spaces cache. The introduction now happens there, in a new `introduce_on_space_join`.

It targets the space's **first channel**: the bridge's `list_spaces` carries no `is_public`, but the server orders each space's channels `is_public DESC, lower(name)` (`cloud_agent::bridge::handle_list_spaces`), so the first entry is public whenever the space has one — the same "first public channel" rule the native path's `find_public_general_channel` applies, without a second HTTP round trip or its retry ladder.

Three properties worth calling out:
- **Gated on `trigger_space_id`.** The same refresh runs on every boot and reconnect; introducing into every known space each time an agent restarts is the obvious way to get this wrong.
- **Idempotent independently.** `has_channel_intro_been_prompted` already guards the nudge, so a duplicate push (reconnect, retried delivery) cannot introduce twice.
- **Fail-soft.** The cache seed this follows is what lets the agent post at all; an introduction is a nicety and must not take the seeding down with it.

**One repo, no server change.** The symmetrical alternative — make the space synthetic carry `original_invite` and teach the client to resolve the channel — is the same outcome shipped across two repos.

## Tests (`tests/test_space_join_introduction.py`)
- a join introduces **exactly once**, into the first public channel;
- a **startup** refresh introduces nowhere;
- a space with no visible channel logs and still seeds;
- a **failing** nudge leaves the channel seeding intact;
- two pushes for the same space prompt once — this one drives the **real** nudge rather than a recorder, so the once-only marker is genuinely exercised.

## Harness
`ruff check` + `ruff format` clean. `python structural limits` passes (the extraction keeps `refresh_bridge_spaces` under the 100-line function limit). Suite: **4694 passed, 15 skipped, 0 failed**, excluding `test_docker_codex_wiring.py` and `test_reanswer_boundary.py`, which fail to collect on `main` too.

One flake seen and dismissed: `test_opencode_driver.py::test_close_kills_then_cancels_a_child_with_inherited_stdout` failed once under full-suite load and passes three times in isolation on this branch, as well as on `main`. Process-tree shutdown timing, unrelated.

## Acceptance
Add a cloud agent to a space from the Hub: it posts a 2–3 sentence self-introduction in the space's general channel, exactly once, including when it is auto-paused at the moment of the join (it is already woken for `space_join`), and not again on the next reconnect.

https://claude.ai/code/session_01FB5KJuGrzwZ3BhpXFXPutm


shanjing:
`Windows lifecycle (py3.11)` failed once on `test_control_client.py::test_failed_sender_releases_paused_transport[False]` — a blocked CLOSE-frame write timing out on Windows — and passed on re-run with nothing changed. Unrelated to this diff, which touches `bridge_transport.py` and one new test file; `test_control_client.py` references neither. All nine checks green.

PR #385, opened 2026-09-17
fix(codex): surface CLI upgrade-required failures

SoujiOkita98:
## Problem

When Codex rejects a model because the installed CLI is too old, Puffo reduces the actionable provider response to `provider_error` / “The provider could not complete the turn.” An operator cannot tell that upgrading the CLI is the required fix.

Reproduced with Codex CLI **0.142.2**: the provider returned HTTP 400 with “The 'gpt-5.6-sol' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again.” The same credentials and model succeeded on **0.154.0**. A control request using GPT-5.5 also succeeded on 0.142.2.

## Change

Recognize the explicit `requires a newer version of Codex` diagnostic as `codex_upgrade_required` in the existing provider-failure registry. Both JSON-RPC request errors and failed `turn/completed` notifications use this normalization. Operators receive fixed, safe guidance to upgrade the Codex CLI used by Puffo and restart the agent to retry pending messages; raw provider payloads are not added to public errors.

This is deliberately limited to diagnosis: it does not auto-upgrade executables, change model selections, reset sessions, or alter Inbox retry/presence policy.

## Validation

- Both new regression cases fail on the original implementation (`provider_error`) and pass with this change.
- Focused Codex/runtime/Inbox tests: **95 passed**.
- Repository pre-commit checks and Python structural checks passed.
- Full suite on macOS / Python 3.13: **4,699 passed, 12 skipped** (platform-specific and opt-in live tests), 24 warnings.


Glimmer0x (approved):
Reviewed head 9ecbc694643850c94c356ab2112eee95b810e182 and its clean integration with main b7e421b8.

No blocking findings. The explicit CLI-version diagnostic is mapped to fixed operator-safe guidance through both JSON-RPC errors and failed turn/completed notifications. The registry preserves non-retryable provider failure handling and does not expand the public runtime-event vocabulary.

Validation on macOS / Python 3.13.15:
- Both new regression cases fail on current main with provider_error, then pass with this PR; all 30 Codex JSON-RPC tests pass.
- Combined Codex, authentication classification and drained-state tests: 190 passed.
- Full integration suite: 4,738 collected, 4,714 passed, 12 failed, 12 skipped. Eight failures were caused by inherited CODEX_HOME; rerunning both affected modules without that variable passes all 81 tests. The remaining four LingTai inherited-child-pipe cancellation cases also fail on unmodified main under the same interpreter. They are unrelated to this PR, so this is not a claim of a fully green local suite.
- All five PR CI checks passed (Python 3.11/3.12, Windows lifecycle 3.11/3.12, pre-commit).
- git diff --check passed.

No live outdated-Codex/provider request was repeated. No daemon or installed CLI was upgraded during review.

PR #392, opened 2026-09-18
fix(portal): self-heal rpc-service listener after Windows proactor accept failure (WinError 64)

3XIA00:
## Problem

On Windows, a worker-recycle / credential-reload storm mass-kills MCP subprocesses (`taskkill /T`) while some of them have connects to rpc-service (127.0.0.1:63385) in flight. The killed peer's RST completes the pending `AcceptEx` with **WinError 64 (ERROR_NETNAME_DELETED)**, and CPython's `ProactorEventLoop` reacts to *any* `OSError` on the accept future by logging `Accept failed on a socket` and **closing the listening socket permanently** (POSIX's selector loop retries instead). aiohttp's `AppRunner` never notices, so the daemon keeps running with a dead control plane: every agent flips to `mcp_unreachable`, and the worker's `no hello` recycle then kills more MCP children mid-connect — the very storm that killed the listener — until a manual daemon restart.

Root-cause report: `puffo-agent-2.0.6-proactor-accept-winerror64.md`.

## Environment

- puffo-agent 2.0.6, installed as a `uv tool` (site-packages under `%APPDATA%\uv\tools\puffo-agent\`)
- uv-managed CPython 3.14 on Windows 11 → asyncio default `ProactorEventLoop`
- Windows-only: POSIX selector event loop retries `ECONNABORTED`-class accept failures and is unaffected
- Fleet of 15 agents; each MCP subprocess hellos rpc-service roughly every 6s

## Reproduction

Observed twice in production on one host on 2026-09-18 (03:14 and 09:38 Beijing time; the second occurrence left a full asyncio traceback after the listener had run stably for 41 min):

1. Run the daemon on Windows with ≥15 agents on persistent-child harnesses (high-frequency hello traffic to 127.0.0.1:63385).
2. Trigger a batch kill of the claude/MCP process trees — a credential replacement (adapter reload) or a batch transport-probe recycle both do this via `taskkill /T`; a human rewriting the shared `credentials.json` from desktop Claude Code fires the same path.
3. With probability proportional to in-flight connects, a killed client's connection sits in the accept queue when the RST lands → pending `AcceptEx` completes with WinError 64 → asyncio logs `Accept failed on a socket` and closes the listener. Port 63385 disappears from `Get-NetTCPConnection -State Listen`; access log flatlines; all agents flip `mcp_unreachable` and recycle-loop.

Deterministic minimal reproduction (used by the regression tests): close the listening asyncio `Server` beneath a running aiohttp `AppRunner` — the exact post-state ProactorEventLoop leaves behind — then observe that nothing rebinds and the worker probe recycles.

## Changes

1. **Windows listener watchdog** (`portal/rpc_service.py`): probes `bind_host:port` with a real TCP connect every 15s; on failure, rebuilds the `AppRunner` and strictly rebinds the *same* port (mcp-configs on disk pin the port, so a fallback port would listen for nobody). `stop_rpc_service` cancels the watchdog and cleans up any rebound runner. Spawned through the supervised `tasks.spawn` helper so a dying watchdog is itself reported. Worst case after a hit drops from "fleet-wide permanent outage + manual restart" to "~15s of lost hellos, self-healed".

2. **Fault-domain split** (`portal/worker.py`, report §4.2): before a `no hello` strike, the probe asks rpc_service whether its own listener still accepts (`listener_reachable()`). Not accepting → log the daemon-side fault and skip recycle (recycling would re-feed the storm and misattribute a daemon single-point failure to each agent's transport); accepting → the agent's own transport stays the suspect and recycle proceeds unchanged.

3. **`_read_loop` race guard** (`agent/harness/drivers/claude_code.py`): cleanup nulls `self._proc` while the read task is still scheduled, which crashed the task with `AttributeError: 'NoneType' object has no attribute 'stdout'` (seen repeatedly in both storms). Hold a local reference and exit the loop instead.

## Verification (Windows)

- The watchdog and read-loop changes are line-equivalent to a site-packages hotfix that was applied and verified live on the affected Windows host: after patch + daemon restart, all 15 agents returned to `runtime=running` within minutes, port 63385 stayed listening, hellos stable at HTTP 200, and a forced listener-close simulation self-healed within one probe cycle.
- All four regression tests below were run on Windows 11 / uv-managed CPython 3.14 and each was verified to **fail on the pre-fix code** and pass with the fix.
- Full pytest suite run on the same Windows environment: the only branch-caused failure found during development (a bare `create_task` tripping `test_no_bare_task_spawn_remains_in_tree`) is fixed in this branch; the remaining local failures (`test_slim_packaging_headless`, `test_workspace_layout` ×2, `test_ws_local_client` path-swap) reproduce identically on a clean checkout of upstream `main` on this host (symlink-privilege / packaging-matrix environment issues, unrelated to this change).
- `tools/check_python_structure.py` passes.

## Tests (each verified to fail on the pre-fix code)

- `test_listener_watchdog_rebinds_dead_listener` — dead listener is rebound and serves again within the probe cadence.
- `test_probe_skips_recycle_while_rpc_listener_down` — rpc listener unreachable ⇒ no strike, no recycle; listener back ⇒ recycle resumes.
- `test_listener_death_pauses_recycle_until_watchdog_rebinds` — combined incident shape (report §4.5): kill the live asyncio Server beneath the aiohttp runner, assert no recycle while down, watchdog rebinds same port, then normal fault attribution resumes.
- `test_read_loop_exits_cleanly_when_proc_cleared` — the cleanup race ends the loop instead of crashing the task.

## Follow-ups (out of scope, from report §4)

- Surface watchdog rebind events into health/telemetry (currently log-only).
- Stagger batch credential-reload/recycle and prefer a graceful MCP-child shutdown before tree-kill to shrink the trigger surface.
- Report upstream to CPython: proactor accept should retry on transient `OSError`s (at least ERROR_NETNAME_DELETED) instead of closing the listener.


3XIA00:
**QA gate — PASS (independent verification).** Reviewed at head `78722fb` by Puffo QA (@qa-70d8bfa2).

- **All four fix criteria verified:** strict same-port listener rebind (no fallback), `_read_loop` null-guard, fault-domain split (`listener_reachable() is not False` → skip strike/recycle only when the listener is provably down), and the combined §4.5 regression test.
- **Regression tests are non-vacuous** — two-sided (listener down ⇒ no recycle; up ⇒ recycle *does* fire), and each fails on the pre-fix code.
- **Recovery logic verified byte-equivalent** to the site-packages hotfix that recovered 15 agents live on the affected Windows host; the new `tasks.spawn` wrapper is `create_task` + supervised death-reporting (verified), so the watchdog behaviour is unchanged and strictly better observed.
- **CI: 5/5 green on `78722fb`** (pre-commit, pytest 3.11/3.12, Windows-lifecycle 3.11/3.12).
- **No secrets** in any of the changed files.

Evidence-tier note (not a blocker): the fault-domain split is new code (not part of the live-verified hotfix), proven by the simulation tests rather than production. The `daemon-side fault — skipping recycle` log line is the marker to watch on the first real storm post-merge.

Clear to merge from the puffo-ai side. No changes requested.

3XIA00:
**QA update — RETRACTING the PASS above; recommend HOLD.** Two defects surfaced in follow-up review (credit: Jeff Dean's Windows-compat review), independently confirmed against the diff at head `78722fb`:

**HIGH — recovery re-triggers the root-cause storm.** `worker.py::_rpc_listener_down` gates recycle on listener reachability only, with no post-rebind grace. While the listener is down no agent's hello lands, so the instant the watchdog rebinds, *every* agent simultaneously has a stale/missing hello, all pass the now-"up" gate, and all recycle at once → batch `taskkill /T` of MCP children with connects in flight against the freshly-rebound listener → the exact WinError-64 condition this PR targets, potentially oscillating every watchdog cycle. The combined test `test_listener_death_pauses_recycle_until_watchdog_rebinds` *codifies* this: it asserts `reload_calls == [False]` on the first probe after rebind. The test passing means the code does what the test says — not that immediate post-recovery recycle is correct. My earlier PASS verified that test's two-sidedness but did not question whether the "up ⇒ recycle now" arm was itself the right behavior. That was the gap.

**MEDIUM — AppRunner leak on the rebind-failure path + unawaited watchdog cancel.** In `_listener_watchdog`, if `runner.setup()` succeeds but `site.start()` raises, the `except: … continue` branch does not `cleanup()` the just-created runner → one leaked runner per failed rebind. Separately, `stop_rpc_service` calls `_watchdog_task.cancel()` without awaiting it, so a rebind in flight at shutdown can leak a runner created but not yet stored in `_watchdog_runner`.

Neither is present in the live-proven hotfix's happy path — which is why the production evidence (15 agents recovered) didn't surface them; they live on the failure/recovery edges the live run never exercised.

Since open PR #386 reportedly addresses both with a supervise-existing-runner + recovery-watermark + probe-grace design, the fix-approach/convergence call belongs to the authors, not QA. If #386 becomes the target I'll re-gate against it with the same criteria. Holding sign-off on #392 until these two are resolved.

